PR #46, opened 2022-07-13
docs: wrong link and add linux binary usage 

zdtsw:


lee0c:
It appears that the usage is similar or identical between Windows and Linux other than slight differences in how the executable/binary is called - does the full help text need to be reiterated?

It does look like the Windows documentation currently lacks any information about the `validate` command and its options, though.

zdtsw:
> It appears that the usage is similar or identical between Windows and Linux other than slight differences in how the executable/binary is called - does the full help text need to be reiterated?
> 
> It does look like the Windows documentation currently lacks any information about the `validate` command and its options, though.

Yes, I exactly the usage after downloaded linux binary.
I cannot confirm if windows binary has "validate" option or not since I do not have windows to verify it.
But from GH action release, seems all 3 are named after sbom-tool-<os>-x64[.exe]
so I think `dotnet run -p src/Microsoft.Sbom.Tool generate` is just the way for local dev env. to use.

lee0c:
@aasim since it is auto-generated - is the lack of `validate` help text because that command does not exist for the Windows tool, or is it a bug?

zdtsw:
> @zdtsw That file is auto generated when any parameter is changed on the IConfiguration class. If you want to make the documentation for the parameters more user friendly, just edit the classes in the Config/Args folder to update the help strings.

Thanks @aasim ! I will have this PR closed

aasim:
@lee0c the validate command is currently not supported and is included only for future use

PR #69, opened 2022-07-20
fix: Do not delete manifest folder if env variable is not set.

### LINKED ISSUE #66: Deletes the _manifest folder anyway
When I run the tool a second time, I get this error message:

````
$ ./sbom-tool-linux-x64 generate [...]
##[error]Encountered an error while generating the manifest.
##[error]Error details: Unable to create manifest directory at path ./_manifest. Error: The BuildDropRoot folder already contains a _manifest folder. Please delete this folder before running the generation or set the DeleteManifestDirIfPresent environment variable to 'true' to overwrite this folder.
````

The `DeleteManifestDirIfPresent` environment variable is not set, but it deletes the folder anyway. 

I've tested it after setting `DeleteManifestDirIfPresent` as described and it works as expected.

### PR DESCRIPTION
**Pull request by ByAgenT**
Fix an issue when tool was deleting _manifest folder even when environment variable `DeleteManifestDirIfPresent` wasn't set.

Fixes #66

### CODE REVIEW

File: src/Microsoft.Sbom.Api/Workflows/SBOMGenerationWorkflow.cs [hunk 1]
```diff
@@ -109,6 +112,14 @@ await using (SBOMConfigs.StartJsonSerializationAsync())
 
                     return !validErrors.Any();
                 }
+                catch (ManifestFolderExistsException e)
+                {
+                    Recorder.RecordException(e);
```
jiaw37, line 117: instead of having duplicate code from below, maybe in the generic exception, check if exception is type `ManifestFolderExistsException` and then update `deleteSBOMDir`

pseudo code
```
if (e is not ManifestFolderExistsException)
{
   deleteSBOMDir = true
}
```

src/Microsoft.Sbom.Api/Exceptions/ManifestFolderExistsException.cs
```diff
@@ -0,0 +1,27 @@
...
+            : base(message)
+        {
+        }
+
+        public ManifestFolderExistsException(string message, Exception innerException)
+            : base(message, innerException)
+        {
+        }
+    }
+}
```
jiaw37, line 27: nit: add a whitespace

ksigmund (approved):

aasim (approved):

PR #80, opened 2022-07-25
Check for custom attributes across all assemblies in the AppDomain

ksigmund:


aasim (approved):

src/Microsoft.Sbom.Api/Utils/AssemblyConfig.cs
```diff
@@ -24,25 +25,29 @@ public class AssemblyConfig : IAssemblyConfig
         /// <inheritdoc/>
         public string AssemblyDirectory => AssemblyDirectoryValue.Value;
 
-        private static readonly Lazy<string> DefaultSBOMBaseNamespaceUri = new Lazy<string>(() =>
-        {
-            return Assembly.GetExecutingAssembly().GetCustomAttribute<DefaultNamespaceBaseUriAttribute>()?.DefaultBaseNamespaceUri;
-        });
+        private static readonly Lazy<string> DefaultSBOMBaseNamespaceUri
+            = GetCustomAttributeValue<DefaultNamespaceBaseUriAttribute, string>(a => a?.DefaultBaseNamespaceUri);
```
edgarrs, line 29: Is it possible add UTs for this?

ByAgenT (approved):

src/Microsoft.Sbom.Api/Utils/AssemblyConfig.cs
ksigmund: I looked into trying to do this but for a number of reasons it's rather difficult.  One is that `Assembly` is a sealed class which makes it difficult to mock.  I'll continue to try for a bit but there is _some_ coverage around this feature already which is passing.

PR #91, opened 2022-08-02
Add DeleteManifestDirIfPresent argument to API and cli args

aasim:


analyzers.ruleset
```diff
@@ -228,7 +228,7 @@
     <Rule Id="SA1620" Action="None" />
     <Rule Id="SA1621" Action="None" />
     <Rule Id="SA1622" Action="None" />
-    <Rule Id="SA1623" Action="Warning" />
+    <Rule Id="SA1623" Action="None" />
```
aasim, line 231: This is annoying, keeps asking you to start your comments with gets or sets even if the property is boolean

ByAgenT (approved):

PR #92, opened 2022-08-02
Add ability to supply custom package supplier value

### LINKED ISSUE #84: Wrong supplier in RootPackage of SBOM
Hey there, 

I am working on generating an SBOM for a GitHub Action that sets up this SBOM-Tool in a convenient way (see: https://github.com/philips-software/sbom-tool-installer-action). After generating the SBOM, I noticed that the [supplier](https://spdx.github.io/spdx-spec/package-information/#75-package-supplier-field) field of the ``RootPackage`` is set to ``Organization: Microsoft``, which seems incorrect. 
There should be a parameter to control this value. 
  
I uploaded the SBOM as a Gist. Please have a look here:
https://gist.github.com/Brend-Smits/90b62120de7abc989c2768c92a2a49c8#file-sbom-tool-installer-action-sbom-L10757-L10772

ByAgenT:
This PR adds new argument to provide supplier (organization) for generated SBOM.

Remaining to-do:
- [x] Add assembly attribute for supply parameter.
- [x] Add associated tests.
- [x] Update documentation to CLI include -ps parameter usage.
- [x] Update documentation to the API to include PackageSupplier in SBOMMetadata.

Resolves #84


test/Microsoft.Sbom.Api.Tests/Config/ConfigSanitizerTests.cs
```diff
@@ -243,7 +242,40 @@ public void UserProviderNamespaceUriBaseShouldReturnDefaultValue_Succeeds()
...
+            config.PackageSupplier = new ConfigurationSetting<string>
+            {
+                Source = SettingSource.CommandLine,
+                Value = actualOrg
+            };
+
+            config.ManifestToolAction = ManifestToolActions.Validate;
+            configSanitizer.SanitizeConfig(config);
+
+            Assert.AreEqual(actualOrg, config.PackageSupplier.Value);
```
ksigmund, line 278: AreEqual's arguments are expected followed by actual.

src/Microsoft.Sbom.Api/Config/Args/GenerationArgs.cs
```diff
@@ -59,6 +59,10 @@ public class GenerationArgs : CommonArgs
             "SBOM generation fails.")]
         public string PackageVersion { get; set; }
 
+        [ArgShortcut("ps")]
+        [ArgDescription("Supplier of the package this SBOM represents.")]
```
ksigmund, line 63: nit: might be clearer to say "Supplier of the package _that_ this SBOM represents."

docs/setting-up-github-actions.md
```diff
@@ -66,7 +66,7 @@ jobs:
       run: |
         curl -Lo $RUNNER_TEMP/sbom-tool https://github.com/microsoft/sbom-tool/releases/latest/download/sbom-tool-linux-x64
         chmod +x $RUNNER_TEMP/sbom-tool
-        $RUNNER_TEMP/sbom-tool generate -b ./buildOutput -bc . -pn Test -pv 1.0.0 -nsb https://sbom.mycompany.com -V Verbose
+        $RUNNER_TEMP/sbom-tool generate -b ./buildOutput -bc . -pn Test -pv 1.0.0 -ps MyCompany -nsb https://sbom.mycompany.com -V Verbose
         ls ./buildOutput/_manifest/spdx_2.2/
```
aasim, line 70: Not related, but I had added these two lines for testing, can you please remove them while you're in this part of the code?

line 70 and 71

src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Generator.cs
```diff
@@ -129,7 +129,7 @@ private void EnsureRequiredHashesPresent(Sbom.Contracts.Checksum[] fileHashes)
                 Created = identityUtils.GetGenerationTimestamp(internalMetadataProvider),
                 Creators = new List<string>
                 {
-                    $"Organization: {Constants.OrganizationValue}",
+                    $"Organization: {identityUtils.GetPackageSupplier(internalMetadataProvider)}",
```
ksigmund, line 132: I know this isn't totally in the scope of this particular PR, but looking at `IdentityUtils` I'm not seeing the point of it not being a static class.  And if I'm correct, then all of it's methods can probably become extension methods.  That would result in you being able to have cleaner call here to `internalMetadataProvider.GetPackageSupplier()` 

If it's not worth doing in this PR, maybe refactoring separately would make sense.

src/Microsoft.Sbom.Api/Config/ConfigSanitizer.cs
```diff
@@ -115,6 +118,26 @@ private ConfigurationSetting<string> GetNamespaceBaseUriFromAssembly(Configurati
...
 
+        private ConfigurationSetting<string> GetPackageSupplierFromAssembly(Configuration configuration)
+        {
+            if (string.IsNullOrWhiteSpace(assemblyConfig.DefaultPackageSupplier))
+            {
+                return configuration.PackageSupplier;
+            }
+
+            // Give priority to package supplier provided as an argument.
+            if (!string.IsNullOrWhiteSpace(configuration.PackageSupplier?.Value))
```
aasim, line 129: The default value should have higher priority, can you also print a warning as we do in the getnamespaceuri method? We want to let folks know that if a default value is present, providing this argument value will be useless and we will use the default value anyway. That way they won't get confused when this param doesnt work as expected.

test/Microsoft.Sbom.Api.Tests/Config/ConfigSanitizerTests.cs
ByAgenT: Oh this actually maybe just a misleading name for a variable, `actualOrg` is an expected value for test, `config.PackageSupplier.Value` is an actual :)

src/Microsoft.Sbom.Api/Config/ConfigSanitizer.cs
ByAgenT: Actually your suggestion appears counterintuitive to me :) Usually it is expected to supply default value if nothing is provided and use provided value if exists. The behavior where default value has higher priority actually bring confusion (and require to provide a warning so user understand what's going on) and also reduces flexibility, e.g. what if you need to override default value via parameter (example case would be all orgs in MSFT use MSFT, but LinkedIn would like to use LinkedIn as an org value, but still use MSFT-built version).

ByAgenT: @edgarrs @daneshbadlani thoughts? Should we prioritize default value over supplier or should we provide ability to customize supplier value even when default incorporated into the tool?

aasim: The 'default' value is a misnomer here. What we want is a mandated value, as in the company you are generating the SBOM for can create a mandated value or aka compliant value that *must* be used to generate the SBOM. 

So, in the normal case, the user *will* have to provide a value here since there is no mandated/default value.

However, a company might want to enforce the name or namespace value such that its developers won't be able to override this value. In such a case, they will add this assembly attribute that will enforce that all SBOMs will have the same company name.

edgarrs: I think by providing a parameter value in the CLI or API call, the user expects this value to override any other default\previously set value. So I agree with Anton that the parameter value should have a higher priority than the assembly attribute

aasim: If that is the case, we don’t even need the assembly attribute. We can just set the default value on the configuration class. The assembly attribute is only there for companies to enforce a particular value.

aasim: But that means that anyone can override the default name and we will have to catch it in validation and fail so as to enforce they don’t use a non conforming value

src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Generator.cs
```diff
@@ -21,11 +21,8 @@ namespace Microsoft.Sbom.Parsers.Spdx22SbomParser
     /// </summary>
     public class Generator : IManifestGenerator
     {
-        private readonly IdentityUtils identityUtils;
-
         public Generator()
```
edgarrs, line 24: We can remove the default constructor if we don't have more constructors

Microsoft.Sbom.sln
```diff
@@ -30,6 +30,16 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Sbom.Api.Tests",
 EndProject
```
edgarrs, line 30: Need to add the new parameter here as well:
https://github.com/microsoft/sbom-tool/blob/e1b2e30d5cde6b7d37e9c287f1528719ca577f1c/README.md?plain=1#L57

src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Utils/InternalMetadataProviderIdentityExtensions.cs
```diff
@@ -16,7 +16,7 @@ namespace Microsoft.Sbom.Parsers.Spdx22SbomParser.Utils
     /// <summary>
     /// Provides helper functions to generate identity strings for SPDX.
     /// </summary>
-    public class IdentityUtils
+    public static class InternalMetadataProviderIdentityExtensions
```
aasim, line 19: wont this make it hard to mock this class?
whats your reasoning behind making this an extension class?

aasim (approved):

src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Utils/InternalMetadataProviderIdentityExtensions.cs
ByAgenT: This suggestion - https://github.com/microsoft/sbom-tool/pull/92#discussion_r939105511 .  I don't think it will make testing harder, but @ksigmund you have any thoughts on that?

edgarrs (approved):

PR #98, opened 2022-08-08
Added docker reference type component. Added check in the converter to filter out null values.

aasim:


src/Microsoft.Sbom.Api/Executors/ComponentToPackageInfoConverter.cs
```diff
@@ -52,7 +52,16 @@ async Task ConvertComponentToPackage(ScannedComponent scannedComponent, Channel<
                     try
                     {
                         var sbom = scannedComponent.ToSbomPackage(report);
-                        await output.Writer.WriteAsync(sbom);
+
+                        if (sbom == null)
```
aasim, line 56: Ideally, the adapter should throw and we can act on the failure accordingly. This needs to be fixed in the adapter code, it currently just fails silently.

ksigmund:
Can you add a test here?  https://github.com/microsoft/sbom-tool/blob/d920e52e6cf3a9e2ae9f166f31415b2a643e1636/test/Microsoft.Sbom.Adapters.Tests/ComponentDetectionToSBOMPackageAdapterTests.cs

ByAgenT (approved):

PR #104, opened 2022-08-11
Revert to v1.1.12 version of CD

aasim:


aasim:
The older version doesn't support Docker reference types, so removed that detector.

ksigmund (approved):

PR #110, opened 2022-08-18
Add dockerfile

sbs2001:


aasim:
Instead of building the app from source, can you please just use the release artifacts from [here](https://github.com/microsoft/sbom-tool/releases/latest) and run it. If you use the url `https://github.com/microsoft/sbom-tool/releases/latest` to download the latest artifact.

sbs2001:
@aasim 

>Instead of building the app from source, can you please just use the release artifacts from [here](https://github.com/microsoft/sbom-tool/releases/latest) and run it. If you use the url https://github.com/microsoft/sbom-tool/releases/latest to download the latest artifact.

 that way the image built by the dockerfile won't be tracking the actual code. It can't be tested or be used for development. 

Also it seems very fragile to have the URL hardcoded and doing bash magic to obtain the download URL . It would get only worse since multiple platforms would need to be targeted.

FYI this was made referring to https://docs.microsoft.com/en-us/dotnet/core/docker/build-container?tabs=linux#create-the-dockerfile

sbs2001:
@aasim  ping ^

sbs2001:
@aasim  ping again

aasim:
@sbs2001 the docker image generation can be incorporated in our pipelines, thereby making sure we deploy a new image to a container registry every time a new release goes out. Since we will use a linux based container, we won't need to worry about downloading the right binaries or even using a URL, since we will have all the *signed* artifacts in the build folder. Unfortunately, our build pipeline is internal since we sign all our artifacts, and I will manually update that.

The change you made can be actually used for development purposes only and we should document it as such. The docker file looks good, can you add a section in the readme file titled, 'building the sbom-tool as a docker image' and also in the https://github.com/microsoft/sbom-tool/blob/main/docs/building-from-source.md file. Once you add the docs, I think we are good to go for your PR.

And sorry about the delay, I got caught up in other work items and missed this.

sbs2001:
@aasim sorry for the delay. I've implemented your suggestions.

ByAgenT (approved):
Looks good to me. It would be awesome to add some examples how to mount drop and source code and run the tool inside container in `Run the tool` section of Readme, but it should not block this PR anymore. Thanks for the contribution!

Issue #95, opened 2022-08-03
Remove and rename references to DropValidator


edgarrs:
https://github.com/search?q=repo%3Amicrosoft%2Fsbom-tool%20DropValidator&type=code

PR #111, opened 2022-08-22
Remove mentions of DropValidator

edgarrs:
Built and ran UTs after renames

### DIFF
--- a/src/Microsoft.Sbom.Api/Workflows/SbomToolWorkflow.cs
+++ b/src/Microsoft.Sbom.Api/Workflows/SbomToolWorkflow.cs
@@ -25,7 +25,7 @@ namespace Microsoft.Sbom.Api.Workflows
     /// <summary>
     /// Defines the workflow steps for the drop validation action.
     /// </summary>
-    public class DropValidatorWorkflow : IWorkflow
+    public class SbomToolWorkflow : IWorkflow
     {
         private readonly IConfiguration configuration;
         private readonly DirectoryWalker directoryWalker;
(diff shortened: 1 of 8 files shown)


edgarrs:
Fixes #95

src/Microsoft.Sbom.Api/Workflows/SbomToolWorkflow.cs
```diff
@@ -25,7 +25,7 @@ namespace Microsoft.Sbom.Api.Workflows
     /// <summary>
     /// Defines the workflow steps for the drop validation action.
     /// </summary>
-    public class DropValidatorWorkflow : IWorkflow
+    public class SbomToolWorkflow : IWorkflow
```
ByAgenT, line 28: I think this workflow should better be called `SbomValidationWorkflow`. It is not the root workflow as it may appear, it is a workflow for validation process along with `SBOMGenerationWorkflow` that lives in the same folder as describes generation process.

ksigmund: Also if we can prefer "SbomXxx" to "SBOMXxx" that would be more consistent with coding standards.

edgarrs: @ksigmund I'd prefer SbomXxx but to be consistent with the actual code I'll use SBOMXxx. We can later do a renaming pass to convert SBOM to Sbom in the code

ByAgenT: Yea, I would also prefer to rename to `SbomXxx` to standardize the code

ByAgenT (approved):

PR #113, opened 2022-08-23
Pacakge external references values should have dash character instead…

### LINKED ISSUE #108: The referencecategory should be PACKAGE-MANAGER
Using the [spdx online tool](https://tools.spdx.org/app/validate/) to verify the generated spdx file has a warning
b7a34e07f1820c100c1222ef10517b6
The referencecategory should be PACKAGE-MANAGER instead of PACKAGE_MANAGER.

edgarrs:
… of underscore to meet SPDX 2.2 specs
Fix #108

### DIFF
--- a/src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Entities/ExternalReference.cs
+++ b/src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Entities/ExternalReference.cs
@@ -17,7 +17,7 @@ namespace Microsoft.Sbom.Parsers.Spdx22SbomParser.Entities
         /// Gets or sets the category for the external reference.
         /// </summary>
         [JsonPropertyName("referenceCategory")]
-        public ReferenceCategory ReferenceCategory { get; set; }
+        public string ReferenceCategory { get; set; }
 
         /// <summary>
         /// Gets or sets type of the external reference. These are definined in an appendix in the SPDX specification.
--- a/test/Microsoft.Sbom.Parsers.Spdx22SbomParser.Tests/Utils/SPDXExtensionsTest.cs
+++ b/test/Microsoft.Sbom.Parsers.Spdx22SbomParser.Tests/Utils/SPDXExtensionsTest.cs
@@ -43,7 +43,7 @@ namespace SPDX22SBOMParserTest
         {
             spdxPackage.AddPackageUrls(packageInfo);
             var externalRef = spdxPackage.ExternalReferences.First();
-            Assert.AreEqual(ReferenceCategory.PACKAGE_MANAGER, externalRef.ReferenceCategory);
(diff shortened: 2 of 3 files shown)


aasim:
Why not add a     [EnumMember(Value = "string value")] to the enum itself, and use a string enum converter to serialize the data

ByAgenT (approved):

aasim (changes requested):
I think adding just the enum converter would greatly reduce the changes in this PR

test/Microsoft.Sbom.Parsers.Spdx22SbomParser.Tests/Utils/SPDXExtensionsTest.cs
```diff
@@ -133,5 +133,17 @@ public void AddSpdxIdTest_SpdxFile_Success()
...
+        public void ReferenceCategoryToNormalizedString_DoesNotContainUnderscore()
+        {
+            foreach (ReferenceCategory referenceCategory in Enum.GetValues(typeof(ReferenceCategory)))
+            {
+                var value = referenceCategory.ToNormalizedString();
+                Assert.IsFalse(
+                    value.Contains('_'), 
+                    $"The value {value} of the {nameof(ReferenceCategory)} enum contains an underscore character.");
+            }
+        }
```
ksigmund, line 147: 👏

ksigmund (approved):

src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Entities/ExternalReference.cs
```diff
@@ -17,7 +17,7 @@ public class ExternalReference
         /// Gets or sets the category for the external reference.
         /// </summary>
         [JsonPropertyName("referenceCategory")]
-        public ReferenceCategory ReferenceCategory { get; set; }
+        public string ReferenceCategory { get; set; }
```
ksigmund, line 20: I think this PR looks good, but I did want to throw out some food for thought.  This approach relies on the consumer to remember to call the new extension method.  Since I think in the code it's just one place, I'm sure this is fine.  For similar situations where the call is required in a lot of places, I'd suggest a class -- e.g. `ReferenceCategoryName` -- that encapsulates Reference category and whose `ToString()` method is overridden.  That way, instead of getting a string and needing to remember to call a method, the type system will "force" you to use the new class (and the string substitution would then occur implicitly).

edgarrs: I thought about that but I was puzzled by the idea of having logic inside the enum.

aasim (approved):
https://github.com/dotnet/runtime/issues/29975

This feature is not yet available on System.Text.Json.

PR #121, opened 2022-08-30
Fix exception when re-adding arguments to the CD scan

ByAgenT:


src/Microsoft.Sbom.Api/Utils/ComponentDetectionCliArgumentBuilder.cs
```diff
@@ -85,7 +85,7 @@ public ComponentDetectionCliArgumentBuilder Scan()
 
         public ComponentDetectionCliArgumentBuilder AddDetectorArg(string name, string value)
         {
-            detectorArgs.Add(name, value);
+            detectorArgs.TryAdd(name, value);
```
aasim, line 88: Can we log some error if the return value is false?

aasim (approved):

edgarrs (approved):

src/Microsoft.Sbom.Api/Utils/ComponentDetectionCliArgumentBuilder.cs
ByAgenT: We don't really need to log this error, otherwise it will show up on every execution when cd arguments are provided, because we call for CD multiple times, but just for the subsequent calls with the same args we cache the result.

PR #133, opened 2022-09-09
Added parser for SPDX files 

aasim:


test/Microsoft.Sbom.Parsers.Spdx22SbomParser.Tests/Parser/TestParser.cs
```diff
@@ -0,0 +1,58 @@
...
+using Microsoft.Sbom.Parsers.Spdx22SbomParser;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+using System.Text.Json;
+
+namespace Microsoft.Sbom.Parser
+{
+    internal class TestParser
```
aasim, line 11: This is a temporary testing class, once the final parser is written, I will update this to use that class instead.

test/Microsoft.Sbom.Parsers.Spdx22SbomParser.Tests/Parser/TestParser.cs
```diff
@@ -0,0 +1,58 @@
...
+        public IEnumerable<SBOMFile> GetFiles(Stream stream)
+        {        
+            stream.Read(buffer);
+
+            while (GetFiles(stream, out SBOMFile sbomFile) != 0)
+            {
+                yield return sbomFile;
+            }
+
+            long GetFiles(Stream stream, out SBOMFile sbomFile)
```
aasim, line 26: A ref struct can't be used inside an iterator method, so used an inline method that has its own stack.

src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Constants.cs
```diff
@@ -61,5 +61,28 @@ internal class Constants
                                                             Name = Constants.SPDXName,
                                                             Version = Constants.SPDXVersion
                                                         };
+
+        internal const int ReadBufferSize = 30;
+        internal const int MaxReadBufferSize = 32768;
```
edgarrs, line 66: Why these values?

src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Parser/ParserUtils.cs
```diff
@@ -0,0 +1,108 @@
...
+    /// <summary>
+    /// Read the next JSON token in the reader from the input buffer.
+    /// If the buffer is small and doesn't contain all the text for the next token, 
+    /// a call to GetMoreBytesFromStream is made to read more data into the buffer.
+    /// </summary>
+    /// <exception cref="EndOfStreamException"></exception>
+    public static void Read(Stream stream, ref byte[] buffer, ref Utf8JsonReader reader)
+    {
+        // If the buffer is empty, refill the buffer.
+        if (buffer[0] == 0)
```
edgarrs, line 27: >buffer[0] [](http://example.com/codeflow?start=12&length=9)

Could this throw if buffer is empty?

src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Parser/ParserUtils.cs
```diff
@@ -0,0 +1,108 @@
...
+/// </summary>
+internal class ParserUtils
+{
+    /// <summary>
+    /// Read the next JSON token in the reader from the input buffer.
+    /// If the buffer is small and doesn't contain all the text for the next token, 
+    /// a call to GetMoreBytesFromStream is made to read more data into the buffer.
+    /// </summary>
+    /// <exception cref="EndOfStreamException"></exception>
+    public static void Read(Stream stream, ref byte[] buffer, ref Utf8JsonReader reader)
```
edgarrs, line 24: >stream [](http://example.com/codeflow?start=35&length=6)

check for null

edgarrs, line 24: >Utf8JsonReader [](http://example.com/codeflow?start=66&length=14)

Is SPDX UTF8?

src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Parser/ParserUtils.cs
```diff
@@ -0,0 +1,108 @@
...
+
+    /// <summary>
+    /// Asserts if the reader is at the current expected token.
+    /// </summary>
+    /// <exception cref="ParserError"></exception>
+    internal static void AssertTokenType(Stream stream, ref Utf8JsonReader reader, JsonTokenType expectedTokenType)
+    {
+        if (reader.TokenType != expectedTokenType)
+        {
+            throw new ParserError($"Expected a '{Constants.JsonTokenStrings[(byte)expectedTokenType]}' at position {stream.Position}");
```
edgarrs, line 50: >(byte)expectedTokenType [](http://example.com/codeflow?start=76&length=23)

How can you guarantee that Constants.JsonTokenStrings contains the element you are looking for? What is the relationship between JsonTokenType and Constants.JsonTokenStrings?

src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Parser/SbomFileParser.cs
```diff
@@ -0,0 +1,294 @@
...
+        }
+        catch (EndOfStreamException)
+        {
+            sbomFile = null;
+            return 0;
+        }
+        catch (JsonException e)
+        {
+            sbomFile = null;
+            throw new ParserError($"Error while parsing JSON, addtional details: ${e.Message}");
```
edgarrs, line 79: >ParserError [](http://example.com/codeflow?start=22&length=11)

Maybe add the JsonException and InnerException?

src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Constants.cs
aasim: Oh good catch, these were for testing, I changed it back to 4096.

This size should be greater than the largest token you can have in your JSON file. 4096 is a good buffer size in general and should accommodate all the parameters that we have

src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Parser/SbomFileParser.cs
```diff
@@ -0,0 +1,294 @@
...
+        {
+            throw new ParserError($"Missing required value(s) for file object at position {stream.Position}: {string.Join(",", missingProps)}");
+        }
+    }
+
+    private void ParseProperty(ref Utf8JsonReader reader, ref byte[] buffer)
+    {
+        switch (reader.GetString())
+        {
+            case "fileName":
```
edgarrs, line 129: >"fileName" [](http://example.com/codeflow?start=17&length=10)

Can you use consts?

src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Parser/ParserUtils.cs
aasim: Yes, fixed by adding null check

src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Parser/ParserUtils.cs
aasim: Its an enum, and I have enumerated every element in the enum, so it should always have the required element

src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Parser/SbomFileParser.cs
```diff
@@ -0,0 +1,294 @@
...
+        }
+    }
+
+    private void ValidateSbomFile(SBOMFile sbomFile)
+    {
+        // I want to use the DataAnnotations Validator here, but will check with CB first
+        // before adding a new dependency.
+        var missingProps = new List<string>();
+
+        if (sbomFile.Checksum == null || sbomFile.Checksum.Where(c => c.Algorithm.Name == "SHA256").Count() == 0)
```
edgarrs, line 94: >"SHA256" [](http://example.com/codeflow?start=90&length=8)

Don't we have an enum for this?

src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Parser/SbomFileParser.cs
```diff
@@ -0,0 +1,294 @@
...
+
+        while (reader.TokenType != JsonTokenType.EndArray)
+        {
+            ParserUtils.Read(stream, ref buffer, ref reader);
+            if (reader.TokenType == JsonTokenType.EndArray)
+            {
+                break;
+            }
+
+            licenses.Add(reader.GetString());
```
edgarrs, line 185: You could also have the following code to avoid the break:
if (reader.TokenType == JsonTokenType.EndArray)
{
       licenses.Add(reader.GetString());
}

src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Parser/SbomFileParser.cs
```diff
@@ -0,0 +1,294 @@
...
+    }
+
+    private List<string> ParseLicenseInfoInFilesArray(ref Utf8JsonReader reader, ref byte[] buffer)
+    {
+        var licenses = new List<string>();
+
+        // Read the opening [ of the array
+        ParserUtils.AssertTokenType(stream, ref reader, JsonTokenType.StartArray);
+
+        while (reader.TokenType != JsonTokenType.EndArray)
```
edgarrs, line 177: Would this loop indefinetely if there is no EndArray token?

src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Parser/SbomFileParser.cs
```diff
@@ -0,0 +1,294 @@
...
+
+        // Move to the first property token
+        ParserUtils.Read(stream, ref buffer, ref reader);
+        ParserUtils.AssertTokenType(stream, ref reader, JsonTokenType.PropertyName);
+
+        while (reader.TokenType != JsonTokenType.EndObject)
+        {
+            switch (reader.GetString())
+            {
+                case "algorithm":
```
edgarrs, line 230: >"algorithm" [](http://example.com/codeflow?start=21&length=11)

Enums or consts?

test/Microsoft.Sbom.Parsers.Spdx22SbomParser.Tests/Parser/Strings/JsonStrings.cs
```diff
@@ -0,0 +1,285 @@
...
+                ],
+                ""licenseConcluded"": ""NOASSERTION"",
+                ""licenseInfoInFiles"": [
+                    ""NOASSERTION"", ""GNU""
+                ],
+                ""copyrightText"": ""NOASSERTION""
+            }
+        ]";
+
+    public const string MalformedJson = @"
```
ByAgenT, line 50: Maybe also add more negative tests? Empty JSON (`{}`,`[]`), empty content these two comes to my mind

src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Parser/ParserUtils.cs
aasim: Fixed

aasim: Yes, its required to be UTF8

src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Parser/SbomFileParser.cs
aasim: Fixed

src/Microsoft.Sbom.Contracts/Contracts/SBOMFile.cs
```diff
@@ -34,5 +34,10 @@ public class SBOMFile
         /// Gets or sets the relative path to the BuildDropPath of the file in the SBOM.
         /// </summary>
         public string Path { get; set; }
+
+        /// <summary>
+        /// Gets or sets unique Identifier for the file.
+        /// </summary>
+        public string SPDXId { get; set; }
```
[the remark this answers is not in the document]
aasim: Right, I had remembered to change it before sending out the PR... yes this should be just Id here.

src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Parser/SbomFileParser.cs
aasim: No, the reader call will fail once we reach end of stream, I will add a test

src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Parser/SbomFileParser.cs
```diff
@@ -0,0 +1,294 @@
...
+
+    public SbomFileParser(Stream stream)
+    {
+        this.stream = stream ?? throw new ArgumentNullException(nameof(stream));
+
+        sbomFile = new ();
+    }
+
+    /// <summary>
+    /// Returns an iterator of <see cref="SBOMFile"/> objects by parsing the SPDX JSON files section.
```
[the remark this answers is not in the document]
aasim: Fixed

test/Microsoft.Sbom.Parsers.Spdx22SbomParser.Tests/Parser/Strings/JsonStrings.cs
aasim: Fixed

src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Parser/SbomFileParser.cs
aasim: The licenses.Add method and the line before this == both read to the buffer which is a mutable call and I added the break just to avoid this.

src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Parser/SbomFileParser.cs
aasim: Fixed

src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Parser/SbomFileParser.cs
aasim: Fixed

src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Parser/SbomFileParser.cs
```diff
@@ -0,0 +1,314 @@
...
+        }
+
+        if (sbomFile.LicenseInfoInFiles == null || sbomFile.LicenseInfoInFiles.Count == 0)
+        {
+            missingProps.Add(nameof(sbomFile.LicenseInfoInFiles));
+        }
+
+        if (missingProps.Count() > 0)
+        {
+            throw new ParserError($"Missing required value(s) for file object at position {stream.Position}: {string.Join(",", missingProps)}");
```
ksigmund, line 141: I personally tend to prefer returning something like `ParserResult` and avoid throwing until the last possible moment.  It makes testing somewhat easier and given that all of these scenarios are expected, it's hard to classify them as truly "exceptional."  Just a thought.

src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Exceptions/ParserError.cs
```diff
@@ -0,0 +1,33 @@
...
+using System;
+using System.Runtime.Serialization;
+
+namespace Microsoft.Sbom.Exceptions;
+
+/// <summary>
+/// Thrown when the SPDX parser detects an error in the JSON file.
+/// </summary>
+[Serializable]
+public class ParserError : Exception
```
ksigmund, line 13: I think exceptions generally are named something like i.e. ParserException

src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Parser/SbomFileParser.cs
aasim: hmm.. I like that as well, might be faster as well, since throwing and catching is expensive. Did you think of something like a Monad IResult type?

aasim: lets merge this PR as is, and I will send out a new PR with the results type in it. That will simplify some other stuff as well.

src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Exceptions/ParserError.cs
aasim: Fixed

ByAgenT (approved):

PR #136, opened 2022-09-13
Added SBOM packages parser

aasim:
* Added the packages parser, which is exactly the same as the files parser with different properties.
* Moved some code into the ParserUtils struct for reusability. I will continue to move more stuff there to reuse in future PRs.
* Changed return type to SPDXFile and SPDXPackage, which makes sense since this is a SPDX parser. The conversion to SbomFile should be handled separately (most likely using an implicit constructor).


src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Parser/ParserUtils.cs
```diff
@@ -7,6 +7,10 @@
 using System;
 using Microsoft.Sbom.Parsers.Spdx22SbomParser;
 using Microsoft.Sbom.Exceptions;
+using System.Runtime.InteropServices.ComTypes;
```
ByAgenT, line 10: nit: sort usings

ByAgenT (approved):

PR #141, opened 2022-09-21
Package Microsoft.Sbom.Tool as .NET tool

edgarrs:
Update Microsoft.Sbom.Tool package to be a .NET tool instead of a single dll


edgarrs:
We opted to create a new package to avoid a breaking change of the existing package

PR #151, opened 2022-09-27
Fix read/write directory permission checks

ksigmund:


aasim:
Talked offline, make FileSystemUtils abstract and create Windows and Unix filesystemutils to inherit from this

aasim (approved):

PR #166, opened 2022-11-07
Users/ksigmund/remove ninject extensions conventions

ksigmund:


src/Microsoft.Sbom.Api/Bindings.cs
```diff
@@ -23,12 +23,17 @@
 using Microsoft.Sbom.Common;
```
edgarrs, line 23: Sort and remove unused

src/Microsoft.Sbom.Api/Bindings.cs
```diff
@@ -23,12 +23,17 @@
...
 using Microsoft.Sbom.Contracts.Interfaces;
 using Ninject;
-using Ninject.Extensions.Conventions;
```
edgarrs, line 26: Remove the package from the packages file

Directory.Packages.props
```diff
@@ -47,5 +47,6 @@
         <PackageVersion Include="NuGet.Frameworks" Version="5.11.3" />
```
edgarrs, line 47: Remove Ninject.Extensions.Conventions from this file

edgarrs (approved):

Issue #105, opened 2022-08-12
Add a default 6-minute timeout on CD


aasim:
Add a timeout while calling the CD code so that we don't spend more than 6 minutes running their code and to prevent any hangs.

JamieMagee:
We discussed this on our side, and we should be using the standard `CancellationToken` pattern[^1] if possible. From your side you will create a `CancellationTokenSource`, add a timeout, and provide us with the `CancellationToken`

```C#
var cts = new CancellationTokenSource();
cts.CancelAfter(TimeSpan.FromMinutes(6));
componentDetectionOrchestrator(args, cts.Token);
```

[^1]: https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/concepts/async/cancel-async-tasks-after-a-period-of-time

ksigmund:
We ended up using a default of 15 minutes, which can be overridden by adding the `Timeout` argument, alone with the number of seconds, to the list of `AdditionalComponentDetectorArgs`.

PR #168, opened 2022-11-09
Add a default 6-minute timeout on CD

ksigmund:


ksigmund:
Addresses #105

ByAgenT:
How did we decide on 6 minutes? Are we sure we don't have customers who succeeding scanning but for whom 6 minutes is not enough? Maybe we could make it configurable, for example with environment variable?

src/Microsoft.Sbom.Api/Utils/ComponentDetectorCachedExecutor.cs
```diff
@@ -18,6 +19,7 @@ public class ComponentDetectorCachedExecutor
         private readonly ILogger log;
         private readonly ComponentDetector detector;
         private ConcurrentDictionary<int, ScanResult> results;
+        private readonly TimeSpan componentDetectionTimeout = TimeSpan.FromMinutes(6);
```
edgarrs, line 22: Should we consider this to be overridable using a parameter or env var?

edgarrs: If not, let's bump the timeout to make sure we allocate enough time for large codebases

ksigmund: I would prefer making this configurable as well.  Next time let's make that an explicit requirement in the task ;-)

ksigmund:
> How did we decide on 6 minutes? Are we sure we don't have customers who succeeding scanning but for whom 6 minutes is not enough? Maybe we could make it configurable, for example with environment variable?

Per offline discussion, this will be overridable via the `AdditionalComponentDetectorArgs` parameter and will default to 15 minutes.

src/Microsoft.Sbom.Api/Utils/ComponentDetectionCliArgumentBuilder.cs
```diff
@@ -17,13 +17,18 @@ public class ComponentDetectionCliArgumentBuilder
         private string action;
         private VerbosityMode verbosity = VerbosityMode.Quiet;
         private string sourceDirectory;
-        private Dictionary<string, string> detectorArgs = new Dictionary<string, string>();
+        private Dictionary<string, string> detectorArgs = new Dictionary<string, string>() { 
+            { TimeoutArgsParamName, TimeoutDefaultSeconds.ToString() }
```
edgarrs, line 21: How does a user override this timeout?

ksigmund: Users can override this by adding "Timeout={timeout in seconds}" to the list of `AdditionalComponentDetectorArgs`.

ksigmund:
> How did we decide on 6 minutes? Are we sure we don't have customers who succeeding scanning but for whom 6 minutes is not enough? Maybe we could make it configurable, for example with environment variable?

Apparently this was a WAG.  For now we'll default to 15 minutes and allow the user to override it.

aasim (approved):

Issue #177, opened 2022-11-16
Return validation results for SBOM validator API


aasim:
Currently the validator returns a Boolean value and the validation result is written to disk. Change this signature and make it return the actual validation result.

https://github.com/microsoft/sbom-tool/blob/2290a485fabf0b7a166fe994cfb0ff0fb1e49059/src/Microsoft.Sbom.Contracts/ISBOMValidator.cs#LL32C11-L32C11

https://github.com/microsoft/sbom-tool/blob/2290a485fabf0b7a166fe994cfb0ff0fb1e49059/src/Microsoft.Sbom.Api/Workflows/SBOMParserBasedValidationWorkflow.cs#L141

PR #185, opened 2022-12-07
Return result object in validation API

aasim:


src/Microsoft.Sbom.Contracts/Contracts/SBOMValidationResult.cs
```diff
@@ -0,0 +1,29 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
```
aasim, line 1: I created a new file instead of reusing the SBOMGenerationResult is because:

I would break existing customer of the generator API by doing that.
In the future validation and generation APIs could return different objects, and I feel like it's logical to separate out the two objects.

aasim:
Fixes #177

src/Microsoft.Sbom.Contracts/Contracts/SBOMValidationResult.cs
```diff
@@ -0,0 +1,29 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections.Generic;
+
+namespace Microsoft.Sbom.Contracts
+{
+    /// <summary>
+    /// Represents the result of a SBOM validation action.
+    /// </summary>
+    public class SBOMValidationResult
```
ksigmund, line 11: What if this was an abstract class and you created two classes that extend it like `SbomValidationSuccess` and `SbomValidationFailure`?  The first could effectively be empty as its existence implies success and the second could have just the errors.  This way you'd never risk erroneously have an instance that is both nominally successful as well as contains a non-empty list of errors.

aasim: Fixed

ksigmund (approved):

PR #186, opened 2022-12-14
Added parameter to Display tool version.

### LINKED ISSUE #146: Add command line parameter to display the version of the tool
e.G. sbom-tool -v

sebasgomez238:
Closes #146 

Possible approach which opens the possibility of displaying additional relevant information such as: version of Component Detection being used. Feedback would be appreciated as I wasn't sure if a workflow should be added to accomplish this task.

**Usage:**
`sbom-tool info`

**Output:**
Screen Shot 2022-12-14 at 2 26 26 PM


sebasgomez238:
> 

ByAgenT:
I would prefer to have `--version` parameter instead of `info`. It is more commonly used across cli tools
image

src/Microsoft.Sbom.Api/Config/SbomToolCmdRunner.cs
```diff
@@ -131,5 +139,22 @@ public async Task Generate(GenerationArgs generationArgs)
...
+        /// <summary>
+        /// Prints the version of the tool.
+        /// </summary>
+        [ArgActionMethod]
+        [ArgDescription("Displays information about the version of the tool being used.")]
+        public void Info()
+        {
+            if (!string.IsNullOrEmpty(Version))
+            {
+                Console.WriteLine($"SBOM Tool Version: {Version}");
```
ByAgenT, line 152: nit: I would say just `Console.WriteLine(Version);` should be enough. It would make it easier for automatic tools to parse version

ByAgenT (changes requested):
It would be great to change `info` to `--version` and should be good to check in after

sebasgomez238:
Changed naming and added a shortcut so that the action can be invoked using either:
`sbom-tool version` or `sbom-tool --version`

src/Microsoft.Sbom.Api/Config/SbomToolCmdRunner.cs
```diff
@@ -131,5 +139,23 @@ public async Task Generate(GenerationArgs generationArgs)
                 IsFailed = true;
             }
         }
+
+        /// <summary>
+        /// Prints the version of the tool.
+        /// </summary>
+        [ArgActionMethod]
+        [ArgShortcut("--version")]
```
jiaw37, line 147: if we are following the --version convention, we can add another shortcut, `-v`

sebasgomez238: I considered adding it but saw that `-v` is used as a verbosity parameter inside the generate and validate actions. Adding it shouldn't cause any issues but thought it might be confusing for users.

ByAgenT (approved):

PR #192, opened 2022-12-22
Check if the package version and name are null or whitespace for SPDX

jiaw37:
Check that the packageVersion and packageName are not null or whitespace on the SPDX SBOM
Added test and manually tested with parameters `-pv " " -pn " "` and the output in the SBOM is `"name": "   ",`


src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Utils/InternalMetadataProviderIdentityExtensions.cs
```diff
@@ -34,13 +34,13 @@ public static string GetPackageName(this IInternalMetadataProvider internalMetad
             }
 
             // First check if the user provided a package name.
-            if (internalMetadataProvider.TryGetMetadata(MetadataKey.PackageName, out object packageName))
+            if (internalMetadataProvider.TryGetMetadata(MetadataKey.PackageName, out object packageName) && !string.IsNullOrWhiteSpace(packageName as string))
```
ksigmund, line 37: I know you didn't necessarily write the original code here and are just amending it, but looking at it, I think there's a chance for some cleanup.

Since so much of this is copy/paste, maybe you can create a method that encapsulates the Try and whitespace check.

This is a total guess but maybe an extension method on the internalmetadataprovider that's something that has a signature like:

`public bool TryGetMetadata<T>(this InternalMetadataProvider provider, string keyName, out T returnVal)`

jiaw37: @ksigmund the change has been made, if it looks good, please resolve the comment

src/Microsoft.Sbom.Api/Manifest/Configuration/SbomConfigProvider.cs
```diff
@@ -162,6 +162,19 @@ public bool TryGetMetadata(MetadataKey key, out object value)
         return false;
     }
 
+    public bool TryGetMetadata(MetadataKey key, out string value)
+        {
```
aasim, line 166: Nit: fix the braces

aasim (approved):

edgarrs (approved):

PR #219, opened 2023-03-06
Fix console error when generating sbom.

sebasgomez238:
Stops "Object reference not set to an instance of an object." from being displayed in the console at the end of the Generation task.


sebasgomez238:
Closing. This commit is included in #220.

PR #221, opened 2023-03-09
Update Package Parser to fit SPDX2.2 specs.

sebasgomez238:
Package parser needed to be updated to fit SPDX2.2 specs. when filesAnalyzed is false then licenseInfoFromFiles can be omitted and should not throw an error.

### DIFF
--- a/src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Parser/SbomPackageParser.cs
+++ b/src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Parser/SbomPackageParser.cs
@@ -109,7 +109,7 @@ internal ref struct SbomPackageParser
             missingProps.Add(nameof(sbomPackage.LicenseConcluded));
         }
 
-        if (sbomPackage.LicenseInfoFromFiles == null || sbomPackage.LicenseInfoFromFiles.Count == 0)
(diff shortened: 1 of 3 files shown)


src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Parser/SbomPackageParser.cs
```diff
@@ -109,7 +109,7 @@ private void ValidateSbomPackage(SPDXPackage sbomPackage)
             missingProps.Add(nameof(sbomPackage.LicenseConcluded));
         }
 
-        if (sbomPackage.LicenseInfoFromFiles == null || sbomPackage.LicenseInfoFromFiles.Count == 0)
+        if ((sbomPackage.LicenseInfoFromFiles == null && sbomPackage.FilesAnalyzed == true) || (sbomPackage.LicenseInfoFromFiles?.Count == 0 && sbomPackage.FilesAnalyzed == true))
```
ksigmund, line 112: This condition can probably be simplified.  Maybe something like:

`if (sbomPackage.FilesAnalyzed && (sbomPackage.LicenseInfoFromFiles == null || bomPackage.LicenseInfoFromFiles?.Count == 0))`

ksigmund (approved):

PR #228, opened 2023-03-20
Throw appropriate error message when a manifest file is not present during validation.

ksigmund:
I think where this fix is made is awkward, but given how the objects are initialized, this seems to be the most appropriate place for it.  Initially, I wanted to add this code within the validation workflow, but at that point it's "too late" and the condition handled in this PR has already occurred.

### DIFF
--- a/src/Microsoft.Sbom.Api/Manifest/Configuration/SbomConfigProvider.cs
+++ b/src/Microsoft.Sbom.Api/Manifest/Configuration/SbomConfigProvider.cs
@@ -12,6 +12,7 @@ using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.Sbom.Common.Extensions;
 using System.IO;
+using Microsoft.Sbom.Common.Config;
 
 namespace Microsoft.Sbom.Api.Manifest.Configuration
 {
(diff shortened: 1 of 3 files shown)


src/Microsoft.Sbom.Api/Manifest/Configuration/SbomConfigProvider.cs
```diff
@@ -87,12 +88,16 @@ public class SbomConfigProvider : ISbomConfigProvider
...
+            ArgumentNullException.ThrowIfNull(nameof(manifestInfo));
+
+            if (ConfigsDictionary.TryGetValue(manifestInfo, out var value))
             {
-                throw new ArgumentNullException(nameof(manifestInfo));
+                return value;
+            }
+            else
+            {
+                throw new FileNotFoundException($"Unable to handle a manifest of type {manifestInfo} or no manifest file was found.");
```
edgarrs, line 99: Is it possible to check for the existence of the file at this point to distinguish the two different causes of error?

ksigmund: I don't believe so.  At this point in the code, we are dealing with instances of `IManifestConfigHandler`, which only exposes a single method, `TryGetManifestConfig`.  

We could have the various implementations of `IManifestConfigHandler` throw a `FileNotFoundException` though.  I'm OK doing this instread, although we just need to ensure that each handler handles this condition.

At this point in the code, all we know is that there's no configuration available for the manifestInfo in question.  We can only infer that 1) no handler is available or 2) that no manifest is present.

edgarrs (approved):

Issue #230, opened 2023-03-24
Fix experience to create an SBOM from an existing container


edgarrs:
Right now, in order to produce an SBOM of an existing container, the tool need to be provided fake parameters that should not be needed:
Microsoft.Sbom.Tool.exe generate **-b d:\test -bc d:\test** -pn MyPackage -pv 1.0 -ps MyCompany -nsb http://foo -di algo_my_image

The correct experience should be something like:
Microsoft.Sbom.Tool.exe generate -pn MyPackage -pv 1.0 -ps MyCompany -nsb http://foo -di algo_my_image

edgarrs:
Fix is checked-in, just waiting for the new release

PR #232, opened 2023-03-31
Let docker component detection run without -bc

sebasgomez238:
Initial attempt at resolving #230. For context running the tool to produce an SBOM of a container requires:

Microsoft.Sbom.Tool.exe generate **-b d:\test -bc d:\test** -pn MyPackage -pv 1.0 -ps MyCompany -nsb http://foo/ -di algo_my_image

Users should be able to generate an SBOM of just the contents of the docker Image. To do this the new behavior is:

Microsoft.Sbom.Tool.exe generate **-m d:\test** -pn MyPackage -pv 1.0 -ps MyCompany -nsb http://foo/ -di algo_my_image

This change required adjusting our config validation to allow -b to not be required in certain scenarios. In this approach we create and delete a temp directory to fill in the gaps where a path must be provided.

The previous behavior is still supported for cases where the user wants to include more than just the contents of the image.


src/Microsoft.Sbom.Api/Providers/ExternalDocumentReferenceProviders/CGExternalDocumentReferenceProvider.cs
```diff
@@ -74,7 +74,7 @@ protected override (ChannelReader<JsonDocWithSerializer> results, ChannelReader<
 
         protected override (ChannelReader<ScannedComponent> entities, ChannelReader<FileValidationResult> errors) GetSourceChannel()
         {
-            var (output, cdErrors) = sbomComponentsWalker.GetComponents(Configuration.BuildComponentPath?.Value);
+            var (output, cdErrors) = sbomComponentsWalker.GetComponents(Configuration.BuildComponentPath?.Value ?? Configuration.BuildDropPath?.Value);
```
ksigmund, line 77: Considering this logic exists in a couple places, consider creating a method that just takes the `Configuration` and encapsulates this logic (i.e. the null-coalescing)

edgarrs, line 77: This will use the build drop path for the component scanning, we don't want that. We want to only scan containers but not other components

sebasgomez238: So we would need a separate parameter to use as the desired output path? Maybe add a '-o' parameter similar to how we do it in the Validation args?

src/Microsoft.Sbom.Api/Config/ConfigSanitizer.cs
```diff
@@ -45,6 +45,12 @@ public IConfiguration SanitizeConfig(IConfiguration configuration)
                 .WriteTo.Console(outputTemplate: Constants.LoggerTemplate)
                 .CreateLogger();
 
+            // If BuildDropPath is null then run the logic to check whether it is required or not based on the current configuration.
+            if (configuration.BuildDropPath?.Value == null || (configuration.DockerImagesToScan?.Value != null && configuration.BuildComponentPath?.Value == null))
```
edgarrs, line 49: I don't think configuration.BuildComponentPath?.Value == null check is needed since it was true, the second check will never be reached

edgarrs: If configuration.BuildDropPath?.Value == null and configuration.DockerImagesToScan?.Value == null, this should throw, right?

src/Microsoft.Sbom.Api/Config/ConfigSanitizer.cs
```diff
@@ -88,6 +94,30 @@ private ConfigurationSetting<IList<ManifestInfo>> GetDefaultManifestInfoForValid
...
+            {
+                throw new ValidationArgException($"Please provide a (-m) if you intend to create an SBOM with only the contents of the Docker image or a (-bc) if you intend to include other components in your SBOM.");
+            }
+            else if (configuration.ManifestToolAction != ManifestToolActions.Validate && configuration.ManifestDirPath?.Value == null && configuration.DockerImagesToScan?.Value != null)
+            {
+                throw new ValidationArgException($"Please provide a value for the ManifestDirPath (-m) parameter to generate the SBOM for the specified Docker image.");
+            }
+            else
+            {
+                throw new ValidationArgException($"Please provide a value for the BuildDropPath (-b) parameter to generate the SBOM.");
```
edgarrs, line 117: I would create a separate method to validate the configuration and call it before calling GetTempBuildDropPath

src/Microsoft.Sbom.Api/Executors/ComponentDetectionBaseWalker.cs
```diff
@@ -47,6 +47,14 @@ public abstract class ComponentDetectionBaseWalker
         {
             log.Debug($"Scanning for packages under the root path {buildComponentDirPath}.");
 
+            // If the buildComponentDirPath is null or empty, make sure we have a ManifestDirPath create a new temp directory with a random name.
+            if (!Directory.Exists(configuration.BuildComponentPath?.Value) && Directory.Exists(configuration.ManifestDirPath?.Value))
+            {
+                log.Debug($"The build component directory path {buildComponentDirPath} does not exist. Creating a new temp directory.");
```
edgarrs, line 53: Since this is an implementation detail, I wouldn't log it

src/Microsoft.Sbom.Api/Executors/ComponentDetectionBaseWalker.cs
```diff
@@ -47,6 +47,14 @@ public abstract class ComponentDetectionBaseWalker
         {
             log.Debug($"Scanning for packages under the root path {buildComponentDirPath}.");
 
+            // If the buildComponentDirPath is null or empty, make sure we have a ManifestDirPath create a new temp directory with a random name.
+            if (!Directory.Exists(configuration.BuildComponentPath?.Value) && Directory.Exists(configuration.ManifestDirPath?.Value))
```
edgarrs, line 51: I think you should use FileSystemUtils

src/Microsoft.Sbom.Api/Workflows/SBOMGenerationWorkflow.cs
```diff
@@ -147,6 +147,20 @@ await using (sbomConfigs.StartJsonSerializationAsync())
                     {
                         DeleteManifestFolder(sbomDir);
                     }
+
+                    try
+                    {
+                        // Delete the generated temp folder if necessary 
+                        if (fileSystemUtils.DirectoryExists(IConfiguration.RandomTempPath))
+                        {
+                            log.Debug($"Deleting the temp directory {IConfiguration.RandomTempPath}.");
```
edgarrs, line 156: I would not log this

src/Microsoft.Sbom.Common/Config/IConfiguration.cs
```diff
@@ -19,6 +20,11 @@ namespace Microsoft.Sbom.Common.Config
     /// </summary>
     public interface IConfiguration
     {
+        /// <summary>
+        /// RandomTempPath to be used in scenarios where a (-m) and (-di) are provided and the user does not provide a (-b) path.
+        /// </summary>
+        public static string RandomTempPath = Path.Combine(Path.GetTempPath(), Path.GetRandomFileName());
```
edgarrs, line 26: This looks like a method that should be in FileSystemUtils

edgarrs (dismissed):

src/Microsoft.Sbom.Api/Executors/ComponentDetectionBaseWalker.cs
sebasgomez238: I considered changing the message to something about scanning a docker image but this would cause it to logged along with other workflows that occur when we scan the empty temp path we create. So in the end I opted for just removing it. Let me know if you have any other suggestions.

src/Microsoft.Sbom.Api/Config/ConfigSanitizer.cs
sebasgomez238: This check is there to cover the scenario where the user provides -b and -di but no -bc and no -m.

bAndDi

sebasgomez238: Correct, If configuration.BuildDropPath?.Value == null and configuration.DockerImagesToScan?.Value == null we will throw an error with a message asking for a BuildDropPath.

src/Microsoft.Sbom.Api/Config/ConfigSanitizer.cs
sebasgomez238: Fixed.

src/Microsoft.Sbom.Common/IFileSystemUtils.cs
```diff
@@ -65,5 +65,10 @@ public interface IFileSystemUtils
         /// Returns whether a directory is empty.
         /// </summary>
         bool IsDirectoryEmpty(string directoryPath);
+
+        /// <summary>
+        /// RandomTempPath to be used in scenarios where a (-m) and (-di) are provided and the user does not provide a (-b) path.
+        /// </summary>
+        public static string RandomTempPath = Path.Combine(Path.GetTempPath(), Path.GetRandomFileName());
```
ByAgenT, line 72: I would prefer not to store static value in the interface. Can you change this interface to declare method instead, implement this method on fileSystemUtils classes that we have, call it early when initializing config, and then pass it around in config object rather than referencing this value from interface in many places?

ByAgenT: Interfaces should only declare behaviors and should very rarely contain any implementations (ideally never).

src/Microsoft.Sbom.Common/Config/IConfiguration.cs
```diff
@@ -9,6 +9,7 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
+using System.IO;
```
ByAgenT, line 12: nit: unnecessary using

src/Microsoft.Sbom.Api/Config/ConfigSanitizer.cs
```diff
@@ -45,6 +45,13 @@ public IConfiguration SanitizeConfig(IConfiguration configuration)
                 .WriteTo.Console(outputTemplate: Constants.LoggerTemplate)
                 .CreateLogger();
 
+            // If BuildDropPath is null then run the logic to check whether it is required or not based on the current configuration.
+            if (configuration.BuildDropPath?.Value == null || (configuration.DockerImagesToScan?.Value != null && configuration.BuildComponentPath?.Value == null))
+            {
+                ValidateBuildDropPathConfiguration(configuration);
+                configuration.BuildDropPath = GetTempBuildDropPath(configuration);
```
ByAgenT, line 52: Why is this temporary path needed? Can we just ignore providers that use `BuildDropPath` if  `BuildDropPath` is not provided? We should still enforce valid arguments from the client by arg checking, but on workflow/provider level, we can potentially just ignore and continue to the next provider if `BuildDropPath` is empty. Though we can add logging that says something like _"No `BuildDropPath` provided, skipping"_ in these providers.

src/Microsoft.Sbom.Api/Config/Validators/ConfigValidator.cs
```diff
@@ -2,8 +2,10 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using Microsoft.Sbom.Api.Utils;
+using Microsoft.VisualBasic;
```
ByAgenT, line 5: visual basic 😱😱😱

nit: unused imports here :)

ByAgenT (changes requested):

src/Microsoft.Sbom.Common/IFileSystemUtils.cs
sebasgomez238: Fixed in new commit.

src/Microsoft.Sbom.Api/Config/ConfigSanitizer.cs
```diff
@@ -84,14 +91,43 @@ private ConfigurationSetting<IList<ManifestInfo>> GetDefaultManifestInfoForValid
...
+                    defaultManifestInfo
+                }
+            };
+        }
+
+    private void ValidateBuildDropPathConfiguration(IConfiguration configuration)
         {
-            Source = SettingSource.Default,
-            Value = new List<ManifestInfo>()
+            if (configuration.ManifestToolAction != ManifestToolActions.Validate && configuration.ManifestDirPath?.Value != null && configuration.DockerImagesToScan?.Value != null)
```
ByAgenT, line 105: These conditionals are hard to read. I see they all have common condition `configuration.ManifestToolAction != ManifestToolActions.Validate`, maybe you can move it out to outer if statement, e.g.
```csharp
if (configuration.ManifestToolAction != ManifestToolActions.Validate) 
{
    if ()
    else if ()
    ...
}
```

ByAgenT: And it probably would be more logical to invert this top condition:
`if (configuration.ManifestToolAction == ManifestToolActions.Generate)`

src/Microsoft.Sbom.Api/Config/ConfigSanitizer.cs
```diff
@@ -84,14 +91,43 @@ private ConfigurationSetting<IList<ManifestInfo>> GetDefaultManifestInfoForValid
...
+            {
+                throw new ValidationArgException($"Please provide a (-m) if you intend to create an SBOM with only the contents of the Docker image or a (-bc) if you intend to include other components in your SBOM.");
+            }
+            else if (configuration.ManifestToolAction != ManifestToolActions.Validate && configuration.ManifestDirPath?.Value == null && configuration.DockerImagesToScan?.Value != null)
+            {
+                throw new ValidationArgException($"Please provide a value for the ManifestDirPath (-m) parameter to generate the SBOM for the specified Docker image.");
+            }
+            else
+            {
+                throw new ValidationArgException($"Please provide a value for the BuildDropPath (-b) parameter to generate the SBOM.");
```
ByAgenT, line 119: This exception message might appear during validation with these parameters. These parameters are not particularly valid, but exception message for empty -b param should not confuse customer about validation vs generation.
```
>> Microsoft.Sbom.Tool.exe Validate -b "" -o X:\test\output.json -mi SPDX:2.2 -v Verbose
Please provide a value for the BuildDropPath (-b) parameter to generate the SBOM.
```

src/Microsoft.Sbom.Api/Workflows/SBOMGenerationWorkflow.cs
```diff
@@ -142,11 +142,24 @@ await using (sbomConfigs.StartJsonSerializationAsync())
                     recorder.RecordTotalErrors(validErrors);
                 }
 
-                // Delete the generated _manifest folder if generation failed.
+               // Delete the generated _manifest folder if generation failed.
                 if (deleteSBOMDir || validErrors.Any())
                 {
-                    DeleteManifestFolder(sbomDir);
+                        DeleteManifestFolder(sbomDir);
```
ByAgenT, line 148: nit: too many tabs

src/Microsoft.Sbom.Api/Workflows/SBOMGenerationWorkflow.cs
```diff
@@ -142,11 +142,24 @@ await using (sbomConfigs.StartJsonSerializationAsync())
...
                 }
+
+                try
+                {
+                    // Delete the generated temp folder if necessary 
+                    if (fileSystemUtils.DirectoryExists(fileSystemUtils.GetSbomToolTempPath()))
+                    {
+                        fileSystemUtils.DeleteDir(fileSystemUtils.GetSbomToolTempPath(), true);
+                    }
+                    }
```
ByAgenT, line 158: nit: not aligned closing bracket

test/Microsoft.Sbom.Api.Tests/Executors/SBOMComponentsWalkerTests.cs
```diff
@@ -108,9 +108,9 @@ public async Task GetComponentsWithFiltering()
             ComponentsFound = scannedComponents
         };
 
-        mockDetector.Setup(o => o.ScanAsync(It.IsAny<string[]>())).Returns(Task.FromResult(scanResult));
-        var walker = new SBOMComponentsWalker(mockLogger.Object, mockDetector.Object, mockConfiguration.Object, mockSbomConfigs.Object);
-        var packagesChannelReader = walker.GetComponents("root");
+            mockDetector.Setup(o => o.ScanAsync(It.IsAny<string[]>())).Returns(Task.FromResult(scanResult));
```
ByAgenT, line 111: nit: looks like different indentation here as well

test/Microsoft.Sbom.Api.Tests/Workflows/ManifestGenerationWorkflowTests.cs
```diff
@@ -272,15 +272,15 @@ public async Task ManifestGenerationWorkflowTests_Succeeds(bool deleteExistingMa
...
-            configurationMock.Object,
-            new ChannelUtils(),
-            mockLogger.Object,
-            sbomConfigs,
-            new PackageInfoJsonWriter(
-                manifestGeneratorProvider,
-                mockLogger.Object),
-            packageInfoConverterMock.Object,
-            new PackagesWalker(mockLogger.Object, mockDetector.Object, configurationMock.Object, sbomConfigs));
+                configurationMock.Object,
```
ByAgenT, line 275: I see misaligned indentation in multiple places. Was this changed automatically or you followed any IDE suggestion? Makes me wonder maybe I setup wrong tab length in .editorconfig

src/Microsoft.Sbom.Api/Config/ConfigSanitizer.cs
sebasgomez238: Fixed.

ByAgenT (approved):

Issue #234, opened 2023-04-07
"+" signs not being serialized properly in manifest.


sebasgomez238:
When generating a manifest that contains a "+" sign it will be serialized as "\u002B".

Current Behavior:
"versionInfo": "2.1.27\u002Bdfsg-2.1" 

Expected:
"versionInfo": "2.1.27+dfsg-2.1"

PR #236, opened 2023-04-10
fix serialization of '+' signs

sebasgomez238:
Addresses #234.

This seems to be a common issue. This was the only way I could get the sign to be displayed as a '+' instead of '\u002B'. I have linked some threads below to consider before we make a decision on making this change.

I attempted other methods before going with this approach such as creating custom encoder settings that allowed the specific character in question to be allowed or allowing all unicode chars to be allowed. None of these worked.

1. https://github.com/dotnet/runtime/issues/35281
2. https://github.com/dotnet/runtime/issues/50998

caution

### DIFF
--- a/src/Microsoft.Sbom.Api/Output/ManifestToolJsonSerializer.cs
+++ b/src/Microsoft.Sbom.Api/Output/ManifestToolJsonSerializer.cs
@@ -4,7 +4,9 @@
 using Microsoft.Sbom.Extensions;
 using System;
 using System.IO;
+using System.Text.Encodings.Web;
 using System.Text.Json;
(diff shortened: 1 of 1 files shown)


aasim:
Fixes #234

sebasgomez238:
> Great research and write up. I think we need to be more specific in solving this instead of relaxing all escaping. I wonder if just using ASCII chars is okay and if there is a way to do that, just thinking out loud here..

There is a JavaScriptEncoder.Create(UnicodeRanges.All) option but this didn't fix the problem.

edgarrs (approved):

PR #237, opened 2023-04-12
Make -nsb parameter optional.

### LINKED ISSUE #206: Consider making the namespace parameter optional
We might be able to follow kubernetes tooling to to use the spdx.org public URL:
https://github.com/kubernetes/release/pull/2319

sebasgomez238:
Resolves #206

Updated -nsb to provide a default value when it is null. This was done following the spdx specs which allow the use of a default value. (Specified in section 6.5.2).

https://spdx.github.io/spdx-spec/v2.2.2/document-creation-information/#653-examples

### DIFF
--- a/src/Microsoft.Sbom.Api/Config/ConfigSanitizer.cs
+++ b/src/Microsoft.Sbom.Api/Config/ConfigSanitizer.cs
@@ -12,6 +12,7 @@ using Serilog;
 using Serilog.Core;
 using System;
 using System.Collections.Generic;
+using System.Reflection;
 using Constants = Microsoft.Sbom.Api.Utils.Constants;
 
 namespace Microsoft.Sbom.Api.Config
@@ -24,6 +25,13 @@ namespace Microsoft.Sbom.Api.Config
         private readonly IHashAlgorithmProvider hashAlgorithmProvider;
         private readonly IFileSystemUtils fileSystemUtils;
         private readonly IAssemblyConfig assemblyConfig;
+
+        internal static string SBOMToolVersion => VersionValue.Value;
+
+        private static readonly Lazy<string> VersionValue = new Lazy<string>(() =>
+        {
+            return typeof(SbomToolCmdRunner).GetTypeInfo().Assembly.GetCustomAttribute<AssemblyInformationalVersionAttribute>()?.InformationalVersion ?? string.Empty;
+        });
 
         public ConfigSanitizer(IHashAlgorithmProvider hashAlgorithmProvider, IFileSystemUtils fileSystemUtils, IAssemblyConfig assemblyConfig)
         {
@@ -50,8 +58,8 @@ namespace Microsoft.Sbom.Api.Config
             // set ManifestDirPath after validation of DirectoryExist and DirectoryPathIsWritable, this wouldn't exist because it needs to be created by the tool.
             configuration.ManifestDirPath = GetManifestDirPath(configuration.ManifestDirPath, configuration.BuildDropPath.Value, configuration.ManifestToolAction);
 
-            // Set namespace value if provided in the assembly
-            configuration.NamespaceUriBase = GetNamespaceBaseUriFromAssembly(configuration, logger);
+            // Set namespace value, this handles default values and user provided values.
+            configuration.NamespaceUriBase = GetNamespaceBaseUri(configuration, logger);
 
             // Set default ManifestInfo for validation in case user doesn't provide a value.
             configuration.ManifestInfo = GetDefaultManifestInfoForValidationAction(configuration);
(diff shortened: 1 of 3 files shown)


src/Microsoft.Sbom.Api/Config/ConfigSanitizer.cs
```diff
@@ -106,14 +114,28 @@ private ConfigurationSetting<AlgorithmName> GetHashAlgorithmName(IConfiguration
...
+            // If assembly name is not defined but a namespace was provided, then return the current value.
+            if (string.IsNullOrWhiteSpace(assemblyConfig.DefaultSBOMNamespaceBaseUri) && !string.IsNullOrEmpty(configuration.NamespaceUriBase?.Value))
             {
                 return configuration.NamespaceUriBase;
             }
 
+            // If assembly name is not defined and namespace was not provided then return the default namespace as per spdx spec https://spdx.github.io/spdx-spec/v2.2.2/document-creation-information/#653-examples.
+            if (string.IsNullOrWhiteSpace(assemblyConfig.DefaultSBOMNamespaceBaseUri) && string.IsNullOrEmpty(configuration.NamespaceUriBase?.Value))
+            {
+                string defaultNamespaceUriBase = "https://spdx.org/spdxdocs/sbom-tool-" + SBOMToolVersion + "-" + Guid.NewGuid();
```
ksigmund, line 128: Use string interpolation here.

src/Microsoft.Sbom.Api/Config/Validators/ConfigValidator.cs
```diff
@@ -50,12 +50,12 @@ public void Validate(string propertyName, object propertyValue, AttributeCollect
                 return;
             }
 
-            // If default value for namespace base uri is provided in the assembly info, skip value check requirements.
-            if (propertyName == nameof(IConfiguration.NamespaceUriBase) && !string.IsNullOrEmpty(assemblyConfig.DefaultSBOMNamespaceBaseUri))
+            // Skip validation of the NamespaceUriBase if it is empty. This is handled in the ConfigSanitizer
+            if (propertyName == nameof(IConfiguration.NamespaceUriBase) && (propertyValue == null || (!string.IsNullOrEmpty(assemblyConfig.DefaultSBOMNamespaceBaseUri))))
```
ksigmund, line 54: If there's a way to break this out into a method (even a local method) that might help readability.  As is, this seems to have evolved to something that's difficult to reason about just by looking at it.

Also, if possible, we probably want a test around this.

src/Microsoft.Sbom.Api/Config/ConfigSanitizer.cs
```diff
@@ -106,14 +114,28 @@ private ConfigurationSetting<AlgorithmName> GetHashAlgorithmName(IConfiguration
...
+
+                logger.Information($"No namespace URI base provided, using default value {defaultNamespaceUriBase}");
+
+                return new ConfigurationSetting<string>
```
ksigmund, line 132: This may be out of scope for this particular PR, but it seems like a constructor that accepts a Value and a Source (and defaults the value of the Source to `SettingSource.Default` might be a good idea.  As it is, need to "remember" to set the source type seems like a bug waiting to happen.

src/Microsoft.Sbom.Api/Config/ConfigSanitizer.cs
```diff
@@ -25,6 +26,13 @@ public class ConfigSanitizer
         private readonly IFileSystemUtils fileSystemUtils;
         private readonly IAssemblyConfig assemblyConfig;
 
+        internal static string SBOMToolVersion => VersionValue.Value;
+
+        private static readonly Lazy<string> VersionValue = new Lazy<string>(() =>
+        {
```
ksigmund, line 32: I think you can omit the `return` and braces here and just let this be a [expression lambda](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/lambda-expressions#expression-lambdas).

src/Microsoft.Sbom.Api/Config/ConfigSanitizer.cs
sebasgomez238: Confirmed that if no Source is provided then telemetry picks it up as "Source=Default".

test/Microsoft.Sbom.Api.Tests/Config/ConfigurationBuilderTestsForGeneration.cs
```diff
@@ -202,6 +202,35 @@ public async Task ConfigurationBuilderTest_Generation_NSBaseUri_Validated()
...
+                NamespaceUriBase = null,
+                PackageSupplier = "Contoso"
+            };
+
+            var config = await cb.GetConfiguration(args);
+
+            Assert.IsNotNull(config); 
+            Assert.IsNotNull(config.ManifestDirPath);
+            Assert.IsNotNull(config.NamespaceUriBase);
+            Assert.AreEqual(Path.Join("ManifestDirPath", Constants.ManifestFolder), config.ManifestDirPath.Value);
```
ksigmund, line 229: nit:  I'd change "ManifestDirPath" to `args.ManifestDirPath` since that's what you're really testing.

ksigmund (approved):

src/Microsoft.Sbom.Api/Config/ConfigSanitizer.cs
```diff
@@ -25,6 +26,10 @@ public class ConfigSanitizer
         private readonly IFileSystemUtils fileSystemUtils;
         private readonly IAssemblyConfig assemblyConfig;
 
+        internal static string SBOMToolVersion => VersionValue.Value;
+
+        private static readonly Lazy<string> VersionValue = new Lazy<string>(() => typeof(SbomToolCmdRunner).GetTypeInfo().Assembly.GetCustomAttribute<AssemblyInformationalVersionAttribute>()?.InformationalVersion ?? string.Empty);
```
edgarrs, line 31: Is this how we are retrieving the version in other places?

sebasgomez238: Yes this the same as in the --version parameter. Would it be better to store it in the Constants?

PR #238, opened 2023-04-13
Check manifest exists on path before validating.

sebasgomez238:
Different approach to #228 

We have had issues when attempting to run validation on a buildDropPath that doesn't contain a manifest throws a misleading error about a missing key. Checking if the manifest is present on the path before running the validation workflow will allow us to catch this.

Question: The new logic checks for manifest.json or manifest.spdx.json. Do we know if this could cause any issues? If so, would just looking for a JSON be good enough?


test/Microsoft.Sbom.Api.Tests/Config/ConfigSanitizerTests.cs
sebasgomez238, line 1: A lot of tests were failing because the ConfigSanitizer would always throw an error about the manifest file not existing so I changed the mockFileSystemUtils for these tests to return true when checking if a file exists.

PR #243, opened 2023-04-19
Add Support for PATCH_FOR relationships

sebasgomez238:
Tested. Constants will be available to users of the API but the logic for when this relationship type should be applied will need to be provided by users of the API.


src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Generator.cs
```diff
@@ -258,9 +258,10 @@ private SPDXRelationshipType GetSPDXRelationshipType(RelationshipType relationsh
             case RelationshipType.DESCRIBES: return SPDXRelationshipType.DESCRIBES;
             case RelationshipType.PREREQUISITE_FOR: return SPDXRelationshipType.PREREQUISITE_FOR;
             case RelationshipType.DESCRIBED_BY: return SPDXRelationshipType.DESCRIBED_BY;
+            case RelationshipType.PATCH_FOR: return SPDXRelationshipType.PATCH_FOR;
             default:
-                throw new NotImplementedException($"The relationship {relationshipType} is currently not " +
-                                                  $"mapped to any SPDX 2.2 relationship type.");
+            throw new NotImplementedException($"The relationship {relationshipType} is currently not " +
```
ByAgenT, line 263: nit: is this correct indentation?

ByAgenT (approved):

src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Generator.cs
sebasgomez238: Fixed.

PR #244, opened 2023-04-20
Optimize calls to `TryGetManifestConfig` in `SbomConfigProvider`

ByAgenT:
I changed getter for `ConfigsDictionary` to exit fast if config dictionary already exists. This getter is a hot path during file hash generation [here](https://github.com/microsoft/sbom-tool/blob/b47f3e1d43c2974a67db1a8ab591f7e29adcaf55/src/Microsoft.Sbom.Api/Executors/FileHasher.cs#L129). Because of that we were calling `TryGetManifestConfig` and `RecordSBOMFormat` methods on every file enumerated, doing unnecessary allocations.

### DIFF
--- a/src/Microsoft.Sbom.Api/Manifest/Configuration/SbomConfigProvider.cs
+++ b/src/Microsoft.Sbom.Api/Manifest/Configuration/SbomConfigProvider.cs
@@ -26,8 +26,14 @@ namespace Microsoft.Sbom.Api.Manifest.Configuration
         {
             get
             {
-                configsDictionary ??= new Dictionary<ManifestInfo, ISbomConfig>();
-
+                if (configsDictionary is not null)
+                {
+                    // Exit fast if config map is already initialized.
+                    return configsDictionary;
+                }
+               
+                // Initialize new config map.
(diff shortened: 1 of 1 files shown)


edgarrs (approved):

src/Microsoft.Sbom.Api/Manifest/Configuration/SbomConfigProvider.cs
```diff
@@ -73,11 +79,6 @@ public class SbomConfigProvider : ISbomConfigProvider
             ILogger logger,
             IRecorder recorder)
         {
-            if (manifestConfigHandlers is null)
```
edgarrs, line 76: Why is this removed?

ByAgenT: It is a duplicate null check, there's null check for this right after this statement
```csharp
this.manifestConfigHandlers = manifestConfigHandlers ?? throw new ArgumentNullException(nameof(manifestConfigHandlers));
```

PR #252, opened 2023-05-01
Users/segomez/path sparator fix new

sebasgomez238:
This PR brings back Kevin's changes related to using the correct path separators. It also adds logic related to which tests we run depending on the operating system that the test is running on.


test/Microsoft.Sbom.Api.Tests/Filters/ManifestFolderFilterTests.cs
sebasgomez238, line 1: Might revisit this and add a test for linux paths if I have time this week.

src/Microsoft.Sbom.Common/Attributes/DirectoryAttribute.cs
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+
+namespace Microsoft.Sbom.Common.Config.Attributes;
+
+/// <summary>
+/// Validate if the property value is a valid URI.
```
aasim, line 9: Fix this comment

sebasgomez238: Fixed

edgarrs (approved):

test/Microsoft.Sbom.Api.Tests/Filters/ManifestFolderFilterTests.cs
```diff
@@ -37,4 +52,35 @@ public void ManifestFolderFilterTest_CheckAllManifestFolder_Succeeds()
         Assert.IsFalse(filter.IsValid("c:/test/_manifest\\manifest.json"));
         configMock.VerifyAll();
     }
+
+    [TestMethod]
+    public void ManifestFolderFilterTest_CheckAllManifestFolder_Succeeds_LinuxBased()
+    {
+        // if OS is windows in the name then don't run the linux test
+        if (isWindows)
```
aasim, line 60: This pattern keeps repeating. We should have a attribute for operating systems for test, so that we can  just say which tests are for Windows and Linux.

src/Microsoft.Sbom.Common/Attributes/DirectoryAttribute.cs
```diff
@@ -0,0 +1,14 @@
...
+
+using System;
+
+namespace Microsoft.Sbom.Common.Config.Attributes;
+
+/// <summary>
+/// This attribute is used to mark a property as a directory.
+/// </summary>
+[AttributeUsage(AttributeTargets.Property, Inherited = false, AllowMultiple = false)]
+public sealed class DirectoryAttribute : Attribute
```
aasim, line 12: Technically this is any path. I would rename this to FilePath or just Path attribute (you probably will have to check with name collision)

PR #253, opened 2023-05-02
Improved telemetry logging.

sebasgomez238:


src/Microsoft.Sbom.Api/Output/Telemetry/TelemetryRecorder.cs
```diff
@@ -90,6 +151,30 @@ public void RecordSBOMFormat(ManifestInfo manifestInfo, string sbomFilePath)
         this.sbomFormats[manifestInfo] = sbomFilePath;
     }
 
+    /// <summary>
+    /// Write telemetry object to the telemetryFilePath.
+    /// </summary>
+    /// <param name="telemetry">The telemetry object to be written to the file.</param>
+    /// /// <param name="telemetryFilePath">The file path we want to write the telemetry to.</param>
+    public async Task RecordToFile(SBOMTelemetry telemetry, string telemetryFilePath)
```
edgarrs, line 159: does this method need to be public? Or can it be private?

src/Microsoft.Sbom.Api/Output/Telemetry/TelemetryRecorder.cs
```diff
@@ -45,6 +48,64 @@ public TelemetryRecorder(IFileSystemUtils fileSystemUtils, IConfiguration config
...
+        return new TelemetryRecorder(configuration, fileSystemUtils);
+    }
+
+    /// <summary>
+    /// Method to log telemetry in conditions when the tool is not able to start execution of workflow.
+    /// </summary>
+    /// <param name="exception">Exception that we want to log.</param>
+    public async Task LogToConsole(Exception exception)
+    {
+        var logger = new LoggerConfiguration()
```
edgarrs, line 73: Since you don't exactly know when someone can call this method. I'll suggest you to do either of two things?

1) Check if the Log method is null and then just follow your existing logic. If Log is not null then use it instead of creating a new Logger.
2) If you don't want this method to be called hen Log is not null, then throw an exception if Log is not null. 

I suggest the first option but it is your call.

src/Microsoft.Sbom.Api/Output/Telemetry/TelemetryRecorder.cs
```diff
@@ -45,6 +48,64 @@ public TelemetryRecorder(IFileSystemUtils fileSystemUtils, IConfiguration config
...
+    public static TelemetryRecorder Create(IConfiguration configuration, IFileSystemUtils fileSystemUtils)
+    {
+        return new TelemetryRecorder(configuration, fileSystemUtils);
+    }
+
+    /// <summary>
+    /// Method to log telemetry in conditions when the tool is not able to start execution of workflow.
+    /// </summary>
+    /// <param name="exception">Exception that we want to log.</param>
+    public async Task LogToConsole(Exception exception)
```
edgarrs, line 71: Looks like you are not only logging to Console, should this method be named something like LogException, instead?

edgarrs (commented):

src/Microsoft.Sbom.Api/Output/Telemetry/TelemetryRecorder.cs
sebasgomez238: Only used inside this class. Changed to private

src/Microsoft.Sbom.Api/Output/Telemetry/TelemetryRecorder.cs
sebasgomez238: Decided 1st option makes the most sense for what we are trying to do here.

src/Microsoft.Sbom.Api/Config/ConfigPostProcessor.cs
```diff
@@ -37,18 +41,27 @@ public void Process(IConfiguration source, IConfiguration destination, Resolutio
...
+            try
+            {
+                // Run validators on all properties.
+                configValidators.ForEach(v => v.Validate(property.DisplayName, property.GetValue(destination), property.Attributes));
+            }
+            catch (Exception ex)
+            {
+                var recorder = TelemetryRecorder.Create(destination, fileSystemUtils);
+                _ = recorder.LogException(ex);
+                throw;
```
jiaw37, line 56:
```suggestion
                throw ex;
```
Do we want to throw the exception ?

src/Microsoft.Sbom.Api/Output/Telemetry/TelemetryRecorder.cs
```diff
@@ -45,6 +48,68 @@ public TelemetryRecorder(IFileSystemUtils fileSystemUtils, IConfiguration config
...
+    /// Method to log telemetry in conditions when the tool is not able to start execution of workflow.
+    /// </summary>
+    /// <param name="exception">Exception that we want to log.</param>
+    public async Task LogException(Exception exception)
+    {
+        var logger = Log;
+        if (Log is null)
+        {
+            logger = new LoggerConfiguration()
+            .MinimumLevel.ControlledBy(new LoggingLevelSwitch { MinimumLevel = Configuration.Verbosity.Value })
```
jiaw37, line 77: indentation

sebasgomez238: Fixed.

edgarrs (approved):

Issue #255, opened 2023-05-05
Add a warning to indicate no dependency packages were found when generating an SBOM if the -bc parameter was specified


edgarrs:
It is very uncommon that there are not package dependencies in an SBOM so we should indicate so when we detect it.

PR #263, opened 2023-05-17
Users/segomez/add package count to telemetry

sebasgomez238:
Adds package count to generation and validation workflows.
- Added small change to address #255 with the same PR.

### DIFF
--- a/src/Microsoft.Sbom.Api/Executors/PackagesWalker.cs
+++ b/src/Microsoft.Sbom.Api/Executors/PackagesWalker.cs
@@ -5,7 +5,6 @@ using System.Collections.Generic;
 using System.Linq;
 using Microsoft.ComponentDetection.Contracts.BcdeModels;
 using Microsoft.ComponentDetection.Contracts.TypedComponent;
-using Microsoft.Sbom.Api.Output.Telemetry;
 using Microsoft.Sbom.Api.Utils;
 using Microsoft.Sbom.Common;
 using Microsoft.Sbom.Common.Config;
@@ -19,8 +18,8 @@ namespace Microsoft.Sbom.Api.Executors;
 /// </summary>
 public class PackagesWalker : ComponentDetectionBaseWalker
 {
-    public PackagesWalker(ILogger log, IRecorder recorder, ComponentDetectorCachedExecutor componentDetector, IConfiguration configuration, ISbomConfigProvider sbomConfigs, IFileSystemUtils fileSystemUtils)
-        : base(log, recorder, componentDetector, configuration, sbomConfigs, fileSystemUtils)
+    public PackagesWalker(ILogger log, ComponentDetectorCachedExecutor componentDetector, IConfiguration configuration, ISbomConfigProvider sbomConfigs, IFileSystemUtils fileSystemUtils)
+        : base(log, componentDetector, configuration, sbomConfigs, fileSystemUtils)
     {
     }
 
--- a/src/Microsoft.Sbom.Api/Executors/PackagesWalker.cs
+++ b/src/Microsoft.Sbom.Api/Executors/PackagesWalker.cs
@@ -5,6 +5,7 @@ using System.Collections.Generic;
 using System.Linq;
 using Microsoft.ComponentDetection.Contracts.BcdeModels;
 using Microsoft.ComponentDetection.Contracts.TypedComponent;
+using Microsoft.Sbom.Api.Output.Telemetry;
 using Microsoft.Sbom.Api.Utils;
 using Microsoft.Sbom.Common;
 using Microsoft.Sbom.Common.Config;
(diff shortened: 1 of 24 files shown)


src/Microsoft.Sbom.Api/Executors/PackagesWalker.cs
```diff
@@ -18,8 +19,8 @@ namespace Microsoft.Sbom.Api.Executors;
 /// </summary>
 public class PackagesWalker : ComponentDetectionBaseWalker
 {
-    public PackagesWalker(ILogger log, ComponentDetectorCachedExecutor componentDetector, IConfiguration configuration, ISbomConfigProvider sbomConfigs, IFileSystemUtils fileSystemUtils)
-        : base(log, componentDetector, configuration, sbomConfigs, fileSystemUtils)
+    public PackagesWalker(ILogger log, IRecorder recorder , ComponentDetectorCachedExecutor componentDetector, IConfiguration configuration, ISbomConfigProvider sbomConfigs, IFileSystemUtils fileSystemUtils)
```
ryanbrandenburg, line 22: GitHub doesn't like the formatting here.

src/Microsoft.Sbom.Api/Output/Telemetry/TelemetryRecorder.cs
```diff
@@ -194,6 +197,22 @@ public void RecordException(Exception exception)
         this.exceptions.Add(exception);
     }
 
+    public void RecordTotalNumberOfPackages(int packageCount)
+    {
+        this.totalNumberOfPackages += packageCount;
```
ryanbrandenburg, line 202: It seems like this `IRecorder` gets registered as a singleton which might be accessed by multiple threads, but this operation is not thread-safe. Our counter may therefore be incorrect in certain race conditions.

ryanbrandenburg (commented):
Spotted one potential issue but I'm not sure I have the context to give an Approval.

src/Microsoft.Sbom.Api/Output/Telemetry/TelemetryRecorder.cs
ryanbrandenburg: Ahh, you're only doing a console App for now (got confused by the name `Microsoft.Sbom.Api`. This would be safe for now then, though we might put a warning on the class/method that it's not thread-safe if it ever might be a real API (I see a bunch of Dictionaries in here which would also be issues).

ByAgenT: This library also getting shipped as nuget package and sbom generation/validation utilizes multithreading. I believe this might be a valid concern that worth investigating

sebasgomez238: Made a change that I believe would help with the addition part. Are the dictionaries you are referring to part of my changes or were they there already?

ryanbrandenburg: @sebasgomez238 and I talked offline, about the += stuff, but WRT the `IDictionary`, `IList` and `HashSet` stuff that's pre-existing (other than the HashSet that was added).

src/Microsoft.Sbom.Api/Output/Telemetry/TelemetryRecorder.cs
```diff
@@ -194,6 +198,25 @@ public void RecordException(Exception exception)
         this.exceptions.Add(exception);
     }
 
+    public void RecordTotalNumberOfPackages(int packageCount)
+    {
+        Interlocked.Add(ref this.totalNumberOfPackages, packageCount);
+    }
+
+    public void RecordUniquePackages(IEnumerable<ScannedComponent> uniqueComponents)
```
edgarrs, line 206: Where is this method being called?

sebasgomez238: In the ComponentDetectionBaseWalker.cs

src/Microsoft.Sbom.Api/Output/Telemetry/TelemetryRecorder.cs
```diff
@@ -194,6 +195,15 @@ public void RecordException(Exception exception)
         this.exceptions.Add(exception);
     }
 
+    /// <summary>
+    /// Record the total number of packages that were processed during the execution of the SBOM tool.
+    /// </summary>
+    /// <param name="packageCount">The total package count after execution.</param>
+    public void RecordTotalNumberOfPackages(int packageCount)
+    {
+        Interlocked.Exchange(ref this.totalNumberOfPackages, packageCount);
```
aasim, line 204: Why do you need this? aren't integer operations atomic by default?

sebasgomez238: I'll change this to a normal operation. This was left from my previous implementation where I was counting packages in the ComponentBaseWalker which could be used by multiple threads.

ryanbrandenburg (approved):

edgarrs (approved):

PR #264, opened 2023-05-18
Make FileHashesDictionary case aware based on OS

aasim:


edgarrs (approved):

src/Microsoft.Sbom.Extensions.DependencyInjection/ServiceCollectionExtensions.cs
```diff
@@ -111,7 +111,10 @@ public static IServiceCollection AddSbomTool(this IServiceCollection services, L
             .AddSingleton<IFileTypeUtils, FileTypeUtils>()
             .AddSingleton<ISignValidationProvider, SignValidationProvider>()
             .AddSingleton<IManifestParserProvider, ManifestParserProvider>()
-            .AddSingleton(_ => new FileHashesDictionary(new ConcurrentDictionary<string, FileHashes>()))
+            .AddSingleton(x => {
+                var comparer = x.GetService<IOSUtils>()?.GetFileSystemStringComparer();
```
ksigmund, line 115: Should this be `GetRequiredService`?  Unless you're OK with the comparer being passed as null...

PR #267, opened 2023-05-26
Users/segomez/path separator fix

sebasgomez238:
Fixed issues that forced us to revert the previous changes. Tested on both Linux and Windows.

### DIFF
--- a/src/Microsoft.Sbom.Common/Utils/PathUtils.cs
+++ b/src/Microsoft.Sbom.Common/Utils/PathUtils.cs
@@ -1,0 +1,30 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.IO;
+using System.Linq;
+using Microsoft.Sbom.Common.Config;
+using Microsoft.Sbom.Common.Config.Attributes;
+
+namespace Microsoft.Sbom.Common.Utils;
+
+/// <summary>
+/// Provides utility function to convert an IConfiguration object to use OS-specific path separators.
+/// </summary>
+public static class PathUtils
+{
+    public static void ConvertToOSSpecificPathSeparators(IConfiguration configuration)
+    {
+        var pathProps = configuration.GetType().GetProperties().Where(p => p.GetCustomAttributes(typeof(PathAttribute), true).Any());
+        foreach (var pathProp in pathProps)
+        {
+            var path = pathProp.GetValue(configuration) as ConfigurationSetting<string>;
+            if (path != null)
+            {
+                path.Value = path.Value.Replace('\\', Path.DirectorySeparatorChar);
+                pathProp.SetValue(configuration, path);
+            }
+        }
+
+    }
+}
--- a/test/Microsoft.Sbom.Api.Tests/Config/ConfigSanitizerTests.cs
+++ b/test/Microsoft.Sbom.Api.Tests/Config/ConfigSanitizerTests.cs
@@ -4,6 +4,7 @@
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Runtime.InteropServices;
 using System.Security.Cryptography;
 using Microsoft.Sbom.Api.Config;
 using Microsoft.Sbom.Api.Exceptions;
@@ -27,6 +28,7 @@ public class ConfigSanitizerTests
     private Mock<IHashAlgorithmProvider> mockHashAlgorithmProvider;
     private Mock<IAssemblyConfig> mockAssemblyConfig;
     private ConfigSanitizer configSanitizer;
+    private bool isWindows;
 
     [TestInitialize]
     public void Initialize()
@@ -50,6 +52,8 @@ public class ConfigSanitizerTests
             });
 
         mockAssemblyConfig = new Mock<IAssemblyConfig>();
+
+        isWindows = RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
 
         configSanitizer = new ConfigSanitizer(mockHashAlgorithmProvider.Object, mockFileSystemUtils.Object, mockAssemblyConfig.Object);
     }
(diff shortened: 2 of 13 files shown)


src/Microsoft.Sbom.Common/Utils/PathUtils.cs
```diff
@@ -0,0 +1,30 @@
...
+        foreach (var pathProp in pathProps)
+        {
+            var path = pathProp.GetValue(configuration) as ConfigurationSetting<string>;
+            if (path != null)
+            {
+                path.Value = path.Value.Replace('\\', Path.DirectorySeparatorChar);
+                pathProp.SetValue(configuration, path);
+            }
+        }
+
```
ryanbrandenburg: Nitpick: Extra newline.

src/Microsoft.Sbom.Common/Utils/PathUtils.cs
```diff
@@ -0,0 +1,30 @@
...
+namespace Microsoft.Sbom.Common.Utils;
+
+/// <summary>
+/// Provides utility function to convert an IConfiguration object to use OS-specific path separators.
+/// </summary>
+public static class PathUtils
+{
+    public static void ConvertToOSSpecificPathSeparators(IConfiguration configuration)
+    {
+        var pathProps = configuration.GetType().GetProperties().Where(p => p.GetCustomAttributes(typeof(PathAttribute), true).Any());
```
ryanbrandenburg, line 18: Do we need to walk up the inheritance tree? We can save ourselves some extra reflection if we don't (also chill if you have reason to believe we will need to in the future and we just want to avoid a future bug).

Nitpick: I find it helpful to explicitly provide the parameter name for bare values like bool and int where their purpose isn't immediately obvious. So this would be `p.GetCustomAttributes(typeof(PathAttribute), inherit: true)`.

test/Microsoft.Sbom.Api.Tests/Config/ConfigSanitizerTests.cs
```diff
@@ -51,6 +53,8 @@ public void Initialize()
 
         mockAssemblyConfig = new Mock<IAssemblyConfig>();
 
+        isWindows = RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
```
ryanbrandenburg, line 56: Nitpick: Personally I would move this up to line 31 since it's simple and doesn't interop with any other Initialization.

test/Microsoft.Sbom.Api.Tests/Config/ConfigSanitizerTests.cs
```diff
@@ -281,4 +286,35 @@ public void ShouldNotOverridePackageSupplierIfProvided_Succeeds()
 
         Assert.AreEqual(actualOrg, config.PackageSupplier.Value);
     }
+
+    [TestMethod]
+    [DataRow(ManifestToolActions.Validate)]
+    [DataRow(ManifestToolActions.Generate)]
+    public void ConfigSantizer_Validate_ReplacesBackslashes_Linux(ManifestToolActions action)
+    {
+        if (!isWindows)
```
ryanbrandenburg, line 295: My original suggestion was going to be that we should use something like `[IgnoreIf(!isWindows)]` but as far as I can tell that doesn't exist in MSTest.

Instead I'll suggest that we test replacing forward slashes when testing on windows (I know the tests only run on linux ATM, but it would run locally and be good future-proofing).

test/Microsoft.Sbom.Api.Tests/Config/ConfigurationBuilderTestsForGeneration.cs
```diff
@@ -197,17 +203,21 @@ public async Task ConfigurationBuilderTest_Generation_NSBaseUri_Validated()
...
+
+        var expectedPath = Path.Join("ManifestDirPath", Constants.ManifestFolder);
+        Assert.AreEqual(Path.GetFullPath(expectedPath), Path.GetFullPath(config.ManifestDirPath.Value));
 
         fileSystemUtilsMock.VerifyAll();
     }
 
     [TestMethod]
-    public async Task ConfigurationBuilderTest_Generation_NullNSBaseUriChangesToDefault()
+    public async Task ConfigurationBuilderTest_Generation_BadNSBaseUriWithDefaultValue_Succeds()
```
ryanbrandenburg, line 214:
```suggestion
    public async Task ConfigurationBuilderTest_Generation_BadNSBaseUriWithDefaultValue_Succeeds()
```

test/Microsoft.Sbom.Api.Tests/Converters/SbomToolManifestPathConverterTests.cs
```diff
@@ -37,85 +40,167 @@ public void Setup()
...
+        osUtils.Setup(o => o.GetCurrentOSPlatform()).Returns(os);
+
+        var (path, isOutsideDropPath) = converter.Convert(rootPath + @"\hello\.\World");
+        Assert.AreEqual("/hello/World", path);
+    }
+
+    [TestMethod]
+    [DataRow(nameof(OSPlatform.Linux))]
+    [DataRow(nameof(OSPlatform.OSX))]
+#if !NETFRAMEWORK
```
ryanbrandenburg, line 147: Curious why we need the condition here.

test/Microsoft.Sbom.Api.Tests/Converters/SbomToolManifestPathConverterTests.cs
```diff
@@ -37,85 +40,167 @@ public void Setup()
     }
 
     [TestMethod]
-    public void SbomToolManifestPathConverterTests_ValidPath_Succeeds()
+    public void SbomToolManifestPathConverterTests_ValidPath_Succeeds_Windows()
     {
+        var os = OSPlatform.Windows;
+        if (!isWindows)
```
ryanbrandenburg, line 46: I think I'm nervous about the number of tests this change has caused to become conditional on OS-type. Maybe I'm thinking about the issue wrong, but if we've made things resilient shouldn't we only need to make test changes in the configSanitizerTests, one test for windows receiving linux and one for linux receiving windows?

test/Microsoft.Sbom.Api.Tests/Config/ConfigSanitizerTests.cs
sebasgomez238: So at the moment there is no logic in the PathUtils.cs that would replace forward slashes to backslashes on windows. I left this out because both are supported on windows so it wouldn't cause an issue if either are provided.

test/Microsoft.Sbom.Api.Tests/Converters/SbomToolManifestPathConverterTests.cs
sebasgomez238: Paths on windows starting with "C:\" and "/home/" on linux was causing a lot of tests to fail exclusively on linux which was why I ended up splitting the tests.

test/Microsoft.Sbom.Api.Tests/Converters/SbomToolManifestPathConverterTests.cs
```diff
@@ -37,85 +40,167 @@ public void Setup()
...
+        {
+            // End the test for windows if the os is not windows
+            return;
```
ryanbrandenburg, line 49: Actually, instead of `return;` on these can we do `Assert.Inconclusive()`? I get nervous when we should green checkmarks for tests that didn't actually run.

test/Microsoft.Sbom.Api.Tests/Converters/SbomToolManifestPathConverterTests.cs
ryanbrandenburg: Hmm, maybe instead we could condition the `rootPath` value on the OS, that way we get to use the same test across OSs instead of maintaining X versions of it.

test/Microsoft.Sbom.Api.Tests/Config/ConfigSanitizerTests.cs
sebasgomez238: It would be visible in telemetry which the user could see if they output telemetry to a file path or use -V verbose. It is a simple change. We would just use path.Value = path.Value.Replace('\\', Path.AltDirectorySeparatorChar). The AltDirectorySeparatorChar is a forward slash on both linux and windows. The only concern would be that any tests that rely on Path.Join() when asserting values would fail since that would return values on backslashes. I ran a quick test and doing this would only require changes in 2 tests.

aasim: My 2 cents: Since windows already supports / we should use that and make our lives simpler.

We should also use path.join whenever we join paths since that will be more correct than just string contactenation. If this causes problems in Windows, we should have our own PathJoin method in IFileSystemUtils, which does path.join internally and normalizes the path to / before returning.

edgarrs (approved):

PR #269, opened 2023-06-06
Add optional validation for sboms with no packages

sebasgomez238:
The parameter shortcut is -n for "noPackages" let me know if there are any suggestions for a better flag.


src/Microsoft.Sbom.Api/Config/ConfigSanitizer.cs
sebasgomez238, line 1: Noticed that Nsb was being generated during validation workflow. This fixes that issue.

src/Microsoft.Sbom.Api/Workflows/SBOMParserBasedValidationWorkflow.cs
sebasgomez238, line 1: I went with this approach where a new list is created because adding the error to the already existing list would cause the failure to show up as another File validation failure which isn't accurate.

test/Microsoft.Sbom.Api.Tests/Config/ConfigSanitizerTests.cs
sebasgomez238, line 1: Action was changed to Generate since the fix was causing this test to fail. Let me know if there are any reasons why we shouldn't do this. My understanding is that we don't need the Namespace uri at all during validation.

src/Microsoft.Sbom.Api/Entities/output/ValidationResultGenerator.cs
```diff
@@ -63,16 +65,34 @@ public ValidationResultGenerator WithValidationResults(IList<FileValidationResul
...
     {
         List<FileValidationResult> validationErrors;
+        List<FileValidationResult> optionalValidationErrors;
         List<FileValidationResult> skippedErrors;
 
         validationErrors = NodeValidationResults.Where(r => !Constants.SkipFailureReportingForErrors.Contains(r.ErrorType)).ToList();
+
+        // Put all the errors into a list for the output to handle, if there are none then create an empty list
+        if (OptionalNodeValidationResults != null)
+        {
```
edgarrs, line 88: Why we need to store the optional validation results in a different list?

sebasgomez238: At first I thought this was necessary but this question made me think it over and realized there was a simple change to get rid of the need for it. I updated the PR accordingly.

src/Microsoft.Sbom.Api/Workflows/SBOMParserBasedValidationWorkflow.cs
```diff
@@ -207,11 +215,22 @@ private void LogIndividualFileResults(IEnumerable<FileValidationResult> validFai
...
         log.Verbose("Unknown file failures:");
         log.Verbose(string.Empty);
         validFailures.Where(vf => vf.ErrorType == ErrorType.Other).ForEach(f => log.Verbose(f.Path));
         log.Verbose("------------------------------------------------------------");
+
+        if (validFailures.Where(vf => vf.ErrorType == ErrorType.NoPackagesFound).Count() == 1)
+        {
+            log.Verbose(string.Empty);
+            log.Verbose("------------------------------------------------------------");
+            log.Verbose("Optional validation results");
```
edgarrs, line 227: Should this be "Package validation results"?

src/Microsoft.Sbom.Api/Workflows/SBOMParserBasedValidationWorkflow.cs
```diff
@@ -238,6 +257,12 @@ private void LogResultsSummary(ValidationResult validationResultOutput, IEnumera
         log.Debug($"Additional files not in the manifest . . . . . . {validFailures.Count(v => v.ErrorType == ErrorType.AdditionalFile)}");
         log.Debug($"Files with invalid hashes . . . . . . . . . . . .{validFailures.Count(v => v.ErrorType == ErrorType.InvalidHash)}");
         log.Debug($"Files in the manifest missing from the disk . . .{validFailures.Count(v => v.ErrorType == ErrorType.MissingFile)}");
+
+        if (validFailures.Where(vf => vf.ErrorType == ErrorType.NoPackagesFound).Count() == 1)
+        {
+            log.Debug($"Optional validation failures . . . . . . . . . . {validFailures.Count(v => v.ErrorType == ErrorType.NoPackagesFound)}");
```
edgarrs, line 263: Same, here "Package validation failures"?

edgarrs (commented):
Please take a look at my feedback

sebasgomez238:
> Please take a look at my feedback

I made the suggested changes.

PR #285, opened 2023-06-29
Add Swid Purl to root package entry

sebasgomez238:


src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Utils/InternalMetadataProviderIdentityExtensions.cs
```diff
@@ -168,6 +168,25 @@ public static string GetDocumentNamespace(this IInternalMetadataProvider interna
...
+        string rootPackageVersion = internalMetadataProvider.GetPackageVersion();
+        string packageSupplierFromMetadata = internalMetadataProvider.GetPackageSupplier();
+        string rootPackageName = internalMetadataProvider.GetPackageName();
+
+        Uri namespaceUri = new Uri(internalMetadataProvider.GetSBOMNamespaceUri());
+
+        // Generate a guid for the new swid tag Id.
+        string tagId = Guid.NewGuid().ToString();
+
+        return $"pkg:swid/{packageSupplierFromMetadata}/{namespaceUri.Host}/{rootPackageName}@{rootPackageVersion}?tag_id={tagId}";
```
aasim, line 187: We should URL encode these strings.

edgarrs, line 187: Is this supposed to be a URI? If so, can we do a new URI().ToString() to validate it is really a URI?

sebasgomez238: Made the change to use a URI but left the method to still return a string as this is what the GetSBOMNamespaceUri() method does as well.

src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Utils/InternalMetadataProviderIdentityExtensions.cs
```diff
@@ -168,6 +168,27 @@ public static string GetDocumentNamespace(this IInternalMetadataProvider interna
...
+        string rootPackageVersion = internalMetadataProvider.GetPackageVersion();
+        string packageSupplierFromMetadata = internalMetadataProvider.GetPackageSupplier();
+        string rootPackageName = internalMetadataProvider.GetPackageName();
+
+        Uri namespaceUri = new Uri(internalMetadataProvider.GetSBOMNamespaceUri());
+
+        // Generate a guid for the new swid tag Id.
+        string tagId = Guid.NewGuid().ToString();
+
+        Uri swidPurl = new Uri($"pkg:swid/{packageSupplierFromMetadata}/{namespaceUri.Host}/{rootPackageName}@{rootPackageVersion}?tag_id={tagId}");
```
edgarrs, line 187: Would creating a Uri do the encoding that Aasim suggested in a previous iteration?

src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Utils/InternalMetadataProviderIdentityExtensions.cs
```diff
@@ -168,6 +168,27 @@ public static string GetDocumentNamespace(this IInternalMetadataProvider interna
         return internalMetadataProvider.GetSBOMNamespaceUri();
     }
 
+    public static string GetSwidTagId(this IInternalMetadataProvider internalMetadataProvider)
```
edgarrs, line 171: Please add UTs for these methods

edgarrs (commented):

src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Utils/InternalMetadataProviderIdentityExtensions.cs
sebasgomez238: Encoding would create a string like this "pkg%3Aswid%2FMicrosoft%2Fmicrosoft.com%2FMicrosoftSQLServer%403.2.12%3Ftag_id%3Dc5bffa01-30a9-45f5-b8a4-7dafd0f27441"

We don't do this for any of the purls in packages. Do we want to make this change?

edgarrs: I think the suggestion was only encoding some parts such as  packageSupplierFromMetadata, rootPackageName androotPackageName

edgarrs (approved):

PR #290, opened 2023-07-20
Handle incomplete reads

ryanbrandenburg:
`Stream.Read` does not always fill the entire buffer (if their own buffer was at its end for example), so in cases where that happened we would be reading "junk" data from a perevious `stream.Read`. We must therefore check afterward and fill up the rest of the buffer (TECHNICALLY this could be wrong for the same reason, but in real world cases [and at least the one that made us find this] it should suffice).

### DIFF
--- a/src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Parser/ParserUtils.cs
+++ b/src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Parser/ParserUtils.cs
@@ -228,6 +228,12 @@ internal class ParserUtils
 
             leftover.CopyTo(buffer);
             bytesRead = stream.Read(buffer.AsSpan(leftover.Length));
+
+            // stream.Read doesn't always return the whole buffer length, so we need to resize the buffer
+            if (bytesRead + leftover.Length != buffer.Length)
+            {
+                bytesRead = stream.Read(buffer.AsSpan(bytesRead + leftover.Length));
+            }
         }
         else
         {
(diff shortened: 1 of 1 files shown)


src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Parser/ParserUtils.cs
```diff
@@ -228,6 +228,12 @@ public static void GetMoreBytesFromStream(Stream stream, ref byte[] buffer, ref
 
             leftover.CopyTo(buffer);
             bytesRead = stream.Read(buffer.AsSpan(leftover.Length));
+
+            // stream.Read doesn't always return the whole buffer length, so we need to fill the rest
```
ByAgenT, line 232: If we expect to fill the entire buffer, would it be preferable to use [ReadExactly](https://learn.microsoft.com/en-us/dotnet/api/system.io.stream.readexactly?view=net-7.0#system-io-stream-readexactly(system-span((system-byte)))) instead, which according to doc advances the position until stream is filled?

ByAgenT (approved):

src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Parser/ParserUtils.cs
ryanbrandenburg: As discussed offline, I'd be super here for that except that `ReadExactly` was introduced in net7 and doesn't return a "bytesRead" count, so we'd have to upgrade this project to net6 and also restructure a bit.

PR #291, opened 2023-07-24
Update CD libraries to latest version

sebasgomez238:


src/Microsoft.Sbom.Extensions.DependencyInjection/ServiceCollectionExtensions.cs
```diff
@@ -111,6 +140,105 @@ public static IServiceCollection AddSbomTool(this IServiceCollection services, L
             .AddSingleton<IFileTypeUtils, FileTypeUtils>()
             .AddSingleton<ISignValidationProvider, SignValidationProvider>()
             .AddSingleton<IManifestParserProvider, ManifestParserProvider>()
+            .AddSingleton<Orchestrator>()
```
edgarrs, line 143: All this should be related to CD. CAn we somehow group them in a separate extension method?

sebasgomez238: I moved the majority of these CD dependencies into their own methods. Except for the Orchestrator and It's dependencies as they were there before, let me know if I should move these as well.

edgarrs: I'll move the Orchestrator and dependencies as well

edgarrs (approved):

PR #293, opened 2023-07-27
Added metadata parser

aasim:


src/Microsoft.Sbom.Extensions/ISbomParser.cs
```diff
@@ -58,7 +58,7 @@ public interface ISbomParser
     /// </summary>
     /// <param name="stream"></param>
     /// <returns></returns>
-    SBOMMetadata GetMetadata();
+    Spdx22Metadata GetMetadata();
```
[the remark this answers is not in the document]
ryanbrandenburg: Yeah, maybe Spdx22Metadata should extend from an `ISBOMMetadata` or something?

src/Microsoft.Sbom.Parsers.Spdx22SbomParser/SPDXParser.cs
```diff
@@ -159,8 +168,68 @@ ParserState MoveToNextState()
...
+                    metadata.SpdxVersion = nextTokenString;
+                    break;
+                case Constants.DataLicenseHeaderName:
+                    metadata.DataLicense = nextTokenString;
+                    break;
+                case Constants.DocumentNameHeaderName:
+                    metadata.Name = nextTokenString;
+                    break;
+                case Constants.DocumentNamespaceHeaderName:
+                    metadata.DocumentNamespace = new Uri(nextTokenString);
```
ryanbrandenburg, line 198: GitHub thinks this might NullRef.

src/Microsoft.Sbom.Parsers.Spdx22SbomParser/SPDXParser.cs
```diff
@@ -159,8 +168,68 @@ ParserState MoveToNextState()
...
+        return internalParserState;
+
+        ParserState ProcessPropertyInternal()
+        {
+            var reader = new Utf8JsonReader(buffer, isFinalBlock: isFinalBlock, readerState);
+            
+            switch (currentRootPropertyName)
+            {
+                case Constants.SPDXVersionHeaderName:
+                    metadata.SpdxVersion = nextTokenString;
```
ryanbrandenburg, line 189: Why do this instead of `ParserUtils.GetString()` or whatever? Whatever the reason is we should codify it in a comment for future generations.

src/Microsoft.Sbom.Parsers.Spdx22SbomParser/SPDXParser.cs
```diff
@@ -287,12 +361,12 @@ public IEnumerable<SBOMRelationship> GetRelationships()
             throw new ParserException($"The parser is not currently enumerating relationships. Current state: {CurrentState}");
         }
 
-        while (GetPackages(stream, out SPDXRelationship sbomRelationship) != 0)
+        while (GetRelationships(stream, out SPDXRelationship sbomRelationship) != 0)
```
ryanbrandenburg, line 364: lol thanks, I noticed this and got confused while reading the class.

src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Utils/ParserStateResult.cs
```diff
@@ -0,0 +1,7 @@
+﻿using Microsoft.Sbom.Contracts.Enums;
+
+namespace Microsoft.Sbom.Utils;
+
+#pragma warning disable SA1313 // Parameter names should begin with lower-case letter
```
ryanbrandenburg, line 5: I think I'd rather change the EditorConfig preferences, but we can do that in a separate PR, no need to block on this.

src/Microsoft.Sbom.Extensions/ISbomParser.cs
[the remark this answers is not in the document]
aasim: Yeah that was my bigges pet peeve about this. This should technically be agnostic to the format... however, the metadata included is very much tied to spdx22, so couldn't come up with a better name.

In the future, if spdx3 comes around, we should create a base class with the common properties, and spdx22metadata and spdx30metadata classes with additional properties. The API then should return the base class. 

Its not SBOMMetadata, since that has a different meaning in our world, its used for generation.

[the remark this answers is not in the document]
ryanbrandenburg: Hmm, maybe we could call this `GetSpdx22Metadata` then to future-proof ourselves?

src/Microsoft.Sbom.Parsers.Spdx22SbomParser/SPDXParser.cs
```diff
@@ -1,6 +1,8 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+// Copyright (c) Microsoft. All rights reserved.
```
ryanbrandenburg, line 4: The copyright got doubled somehow.

src/Microsoft.Sbom.Parsers.Spdx22SbomParser/SPDXParser.cs
```diff
@@ -159,8 +172,77 @@ ParserState MoveToNextState()
...
+            readerState = reader.CurrentState;
+            currentRootPropertyName = result.PropertyName;
+            nextTokenString = result.NextToken;
+            return result.State;
+        }
+    }
+
     private void ValidateParsingComplete()
     {
+        if (!metadataStateProcessed)
```
ryanbrandenburg, line 241: NitPick: I don't think this actually validates anything important to the parser. As far as I can tell it's just mandating that  the user calls `GetMetadata`. That makes sense for `GetReferences`, `GetPackages`, etc because they can control the state of the stream/parser, but the metadata is already parsed regardless, it's just a matter of whether the user views the result or not.

ryanbrandenburg (approved):

src/Microsoft.Sbom.Parsers.Spdx22SbomParser/SPDXParser.cs
aasim: The way the state machine is designed, the only way to get to the finished state without a failure event is to get the metadata. Since we want that to be called last. So this becomes necessary.

src/Microsoft.Sbom.Extensions/ISbomParser.cs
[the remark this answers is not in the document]
edgarrs: I would have made this method generic to return T. The entire interface would be ISbomParser<T> where T will be Spdx22Metadata  for this specific implementation. We will need to add ISbomMetadata interface or something to restrict T.

PR #320, opened 2023-08-22
Make calls to ClearlyDefinedApi

sebasgomez238:
Makes calls to the ClearlyDefinedAPI to populate license information about open-source packages.


src/Microsoft.Sbom.Api/Executors/ComponentDetectionBaseWalker.cs
```diff
@@ -32,6 +35,9 @@ public abstract class ComponentDetectionBaseWalker
     private readonly ISbomConfigProvider sbomConfigs;
     private readonly IFileSystemUtils fileSystemUtils;
 
+    public readonly Dictionary<string, string> LicenseDictionary = new Dictionary<string, string>();
```
ryanbrandenburg, line 38: Is this class created as a singleton or on an as-needed basis? I ask because if a Dictionary is accessed in a multi-threaded way it can cause all kinds of exceptions.

src/Microsoft.Sbom.Api/Executors/ComponentDetectionBaseWalker.cs
```diff
@@ -107,8 +113,147 @@ async Task Scan(string path)
 
             var uniqueComponents = FilterScannedComponents(scanResult);
 
+            var listOfLicenses = new List<string>();
+            var clearlyDefinedType = string.Empty;
+            var clearlyDefinedProvider = string.Empty;
+            var clearlyDefinedNamespace = string.Empty;
+            var clearlyDefinedName = string.Empty;
+            var clearlyDefinedRevision = string.Empty;
```
ryanbrandenburg, line 121: Looks like this variable isn't actually used right now?

src/Microsoft.Sbom.Api/Executors/ComponentDetectionBaseWalker.cs
```diff
@@ -107,8 +113,147 @@ async Task Scan(string path)
...
+                            {
+                                string responseContent = await response.Content.ReadAsStringAsync();
+
+                                log.Information($"Received response from {batch.Count} components in {elapsedTime}.");
+
+                                JObject responseObject = JObject.Parse(responseContent); // Parse the JSON string
+
+                                foreach (JToken packageInfoToken in responseObject.Values())
+                                {
+                                    JObject packageInfo = packageInfoToken.ToObject<JObject>();
```
ryanbrandenburg, line 223: It seems like it would be a bit cleaner to define a Model for this response.

src/Microsoft.Sbom.Api/Executors/ComponentDetectionBaseWalker.cs
```diff
@@ -107,8 +113,147 @@ async Task Scan(string path)
...
 
+            var listOfLicenses = new List<string>();
+            var clearlyDefinedType = string.Empty;
+            var clearlyDefinedProvider = string.Empty;
+            var clearlyDefinedNamespace = string.Empty;
+            var clearlyDefinedName = string.Empty;
+            var clearlyDefinedRevision = string.Empty;
+            // var clearlyDefinedExpand = "-files";
+
+            List<string> listOfComponentsForApi = new List<string>();
```
ryanbrandenburg, line 124: I think most of this new code makes sense in its own class (maybe `ILicenseFinder` with implementation `ClearlyDefinedLicenseFinder` or some such [I don't love "Finder" but am failing for other ideas]) which you'll create as a singleton, then you can give it the metadata here and have it do all the logic about weather it needs to hit the API or not internally. Creating it as a singleton also might save you from having a public static Dictionary.

src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Utils/GlobalLicenseDictionary.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections.Generic;
+
+// I know use of a global state is not ideal. I could use feedback on how to achieve this in a better way.
+public static class GlobalLicenseDictionary
+{
+    private static Dictionary<string, string> _licenseDictionary = new Dictionary<string, string>();
```
ryanbrandenburg, line 9: My suggestion above would have you move away from the static Dictionay to the SingletonClass, but I think you'll want to both move away from the Dictionary as the cache since this would grow unbounded if the parser were up for a long time. I also don't have a ton of domain context here, but I would assume that "PackageName" is not a complete key since the license might have changed from one version to the next?

sebasgomez238: I like this idea of using a singleton and you're right the key should be "packageName" + "packageVersion". I'll add this.

src/Microsoft.Sbom.Api/Executors/ComponentDetectionBaseWalker.cs
sebasgomez238: Thanks, this sounds a lot cleaner than the current approach.

src/Microsoft.Sbom.Adapters/Adapters/ComponentDetection/ExtendedScanResult.cs
```diff
@@ -0,0 +1,12 @@
...
+using Newtonsoft.Json;
+using Newtonsoft.Json.Serialization;
+
+namespace Microsoft.Sbom.Adapters.ComponentDetection;
+
+[JsonObject(MemberSerialization.OptOut, NamingStrategyType = typeof(CamelCaseNamingStrategy))]
+public class ExtendedScanResult : ScanResult
+{
+    new public IEnumerable<ExtendedScannedComponent> ComponentsFound { get; set; }
+}
```
JamieMagee, line 12: nit:
```suggestion
public sealed class ScanResultWithLicense : ScanResult
{
    public new IEnumerable<ScannedComponentWithLicense> ComponentsFound { get; init; }
}
```

I'd prefer `WithLicense` instead of `Extended` if possible. It's clearer what it's doing.

src/Microsoft.Sbom.Adapters/Adapters/ComponentDetection/NpmComponentExtensions.cs
```diff
@@ -15,20 +15,33 @@ internal static class NpmComponentExtensions
     /// <summary>
     /// Converts a <see cref="NpmComponent"/> to an <see cref="SbomPackage"/>.
     /// </summary>
```
JamieMagee, line 17: Add `<param>`?

src/Microsoft.Sbom.Adapters/Adapters/ComponentDetection/NpmComponentExtensions.cs
```diff
@@ -15,20 +15,33 @@ internal static class NpmComponentExtensions
...
+            PackageName = npmComponent.Name,
+            PackageVersion = npmComponent.Version,
+            Checksum = new[]
+            {
+                new Checksum { ChecksumValue = npmComponent.Hash },
+            },
+            Supplier = npmComponent.Author?.AsSupplier(),
+            FilesAnalyzed = false,
+            Type = "npm"
+        };
```
JamieMagee, line 33:
```suggestion
            LicenseInfo = license != null ? new LicenseInfo { Concluded = license } : null

        };
```

src/Microsoft.Sbom.Adapters/Adapters/ComponentDetection/ScannedComponentExtensions.cs
```diff
@@ -29,8 +29,8 @@ public static class ScannedComponentExtensions
             GoComponent goComponent => goComponent.ToSbomPackage(),
             LinuxComponent linuxComponent => linuxComponent.ToSbomPackage(),
             MavenComponent mavenComponent => mavenComponent.ToSbomPackage(),
-            NpmComponent npmComponent => npmComponent.ToSbomPackage(),
-            NuGetComponent nuGetComponent => nuGetComponent.ToSbomPackage(),
+            NpmComponent npmComponent => npmComponent.ToSbomPackage(component?.License),
+            NuGetComponent nuGetComponent => nuGetComponent.ToSbomPackage(component?.License), // Extending the TypedComponent class from CD was not possible so I end up with adding the license as a property of the ScannedComponent class through the ExtendedScannedComponent class, and passing that license to the ToSbomPackage method.
```
JamieMagee, line 33: Makes sense 👍

src/Microsoft.Sbom.Adapters/ComponentDetectionToSBOMPackageAdapter.cs
```diff
@@ -35,7 +35,7 @@ public class ComponentDetectionToSBOMPackageAdapter
 
         try
         {
-            ScanResult? componentDetectionScanResult = JsonConvert.DeserializeObject<ScanResult>(File.ReadAllText(bcdeOutputPath));
+            ExtendedScanResult? componentDetectionScanResult = (ExtendedScanResult)JsonConvert.DeserializeObject<ScanResult>(File.ReadAllText(bcdeOutputPath));
```
JamieMagee, line 38: Can you not deserialize this directly to the type you want?

src/Microsoft.Sbom.Api/Executors/ComponentDetectionBaseWalker.cs
```diff
@@ -39,13 +44,14 @@ public abstract class ComponentDetectionBaseWalker
         ComponentDetectorCachedExecutor componentDetector,
         IConfiguration configuration,
         ISbomConfigProvider sbomConfigs,
-        IFileSystemUtils fileSystemUtils)
+        IFileSystemUtils fileSystemUtils, ILicenseInformationFetcher licenseInformationFetcher)
```
JamieMagee, line 47: nit:
```suggestion
        IFileSystemUtils fileSystemUtils,
        ILicenseInformationFetcher licenseInformationFetcher)
```

src/Microsoft.Sbom.Api/Executors/ILicenseInformationFetcher.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Net.Http;
+using System.Threading.Tasks;
+using Microsoft.ComponentDetection.Contracts.BcdeModels;
+
+namespace Microsoft.Sbom.Api.Executors;
+public interface ILicenseInformationFetcher
```
JamieMagee, line 8: nit: XML documentation. Copilot is great at helping write this.

src/Microsoft.Sbom.Api/Executors/LicenseInformationFetcher.cs
```diff
@@ -0,0 +1,186 @@
...
+
+                // If the clearlyDefinedName contains a / then split it and use the first part as the clearlyDefinedNamespace and the second part as the clearlyDefinedName
+                if (componentName.Contains("/"))
+                {
+                    string[] clearlyDefinedNameParts = componentName.Split('/');
+                    clearlyDefinedNamespace = clearlyDefinedNameParts[0];
+                    componentName = clearlyDefinedNameParts[1];
+                }
+
+                // Each componentType has it's own edge cases. So more work is needed before I can add more types.
```
JamieMagee, line 48: Yeah, this needs unit test coverage.

src/Microsoft.Sbom.Api/Executors/LicenseInformationFetcher.cs
```diff
@@ -0,0 +1,186 @@
...
+        List<HttpResponseMessage> responses = new List<HttpResponseMessage>();
+
+        for (int i = 0; i < listOfComponentsForApi.Count; i += batchSize)
+        {
+            List<string> batch = listOfComponentsForApi.Skip(i).Take(batchSize).ToList();
+            string formattedData = "[" + string.Join(",", batch.Select(item => $"\"{item}\"")) + "]";
+
+            log.Information($"Retrieving license information for {batch.Count} components...");
+
+            using (HttpClient httpClient = new HttpClient())
```
JamieMagee, line 90: IMO this `HTTPClient` should be a separate class.

src/Microsoft.Sbom.Api/Executors/LicenseInformationService.cs
```diff
@@ -0,0 +1,76 @@
...
+
+            log.Information($"Retrieving license information for {batch.Count} components...");
+
+            using (HttpClient httpClient = new HttpClient())
+            {
+                Uri uri = new Uri("https://api.clearlydefined.io/definitions");
+                var content = new StringContent(formattedData, Encoding.UTF8, "application/json");
+
+                // Set the headers individually
+                content.Headers.ContentType = new System.Net.Http.Headers.MediaTypeHeaderValue("application/json");
```
ByAgenT, line 41: nit: why use fully qualified name instead of `using System.Net.Http.Headers`?

src/Microsoft.Sbom.Api/Executors/LicenseInformationService.cs
```diff
@@ -0,0 +1,76 @@
...
+
+        foreach (HttpResponseMessage response in responses)
+        {
+            if (response.IsSuccessStatusCode)
+            {
+                responseContent.Add(await response.Content.ReadAsStringAsync());
+            }
+            else
+            {
+                log.Error($"Error encountered while fetching license information from API.");
```
ByAgenT, line 77: Since total result could be partially complete, it might be useful to add in a message that license information might be incomplete in output SBOM.

ByAgenT (approved):

ByAgenT:
Do we need some kind of explicit acknowledgement to users that package name and versions are being sent to ClearlyDefined if option of populating license is on?

src/Microsoft.Sbom.Api/Executors/LicenseInformationService.cs
```diff
@@ -0,0 +1,85 @@
...
+        List<string> responseContent = new List<string>();
+
+        for (int i = 0; i < listOfComponentsForApi.Count; i += batchSize)
+        {
+            List<string> batch = listOfComponentsForApi.Skip(i).Take(batchSize).ToList();
+            string formattedData = "[" + string.Join(",", batch.Select(item => $"\"{item}\"")) + "]";
+
+            log.Debug($"Retrieving license information for {batch.Count} components...");
+
+            using (HttpClient httpClient = new HttpClient())
```
JamieMagee, line 37: Don't `new` this up. Accept it as a constructor parameter. See https://learn.microsoft.com/en-us/aspnet/core/fundamentals/http-requests?view=aspnetcore-7.0#typed-clients

src/Microsoft.Sbom.Api/Executors/LicenseInformationService.cs
```diff
@@ -0,0 +1,85 @@
...
+    private readonly ILogger log;
+
+    public LicenseInformationService(ILogger log)
+    {
+        this.log = log ?? throw new ArgumentNullException(nameof(log));
+    }
+
+    public async Task<List<string>> FetchLicenseInformationFromAPI(List<string> listOfComponentsForApi)
+    {
+        int batchSize = 400;
```
JamieMagee, line 26: I _think_ the API allows up to 500. Is this just to be conservative with batches?

JamieMagee (approved):
Approved pending comments

src/Microsoft.Sbom.Api/Executors/LicenseInformationService.cs
sebasgomez238: I wasn't sure of the limit. Moved it up to 500 without an issue.

src/Microsoft.Sbom.Api/Executors/LicenseInformationService.cs
```diff
@@ -0,0 +1,85 @@
...
+        List<HttpResponseMessage> responses = new List<HttpResponseMessage>();
+        List<string> responseContent = new List<string>();
+
+        for (int i = 0; i < listOfComponentsForApi.Count; i += batchSize)
+        {
+            List<string> batch = listOfComponentsForApi.Skip(i).Take(batchSize).ToList();
+            string formattedData = "[" + string.Join(",", batch.Select(item => $"\"{item}\"")) + "]";
```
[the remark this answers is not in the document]
sebasgomez238: Done. Thanks for the suggestion!

src/Microsoft.Sbom.Api/Executors/LicenseInformationService.cs
sebasgomez238: This one I wasn't too sure about. Would I have to move over some things to a TypedClientModel class as shown in the docs? And I couldn't figure out where the HttpClient would actually get instantiated I was able to test it out a bit and doing a .AddHTTPClient<LicenseInformationService>() in the program.cs works but I'm not sure how adding this would affect people using our API directly.

sebasgomez238:
Was moved over to separate PR

Issue #340, opened 2023-08-31
Fix analyzer warnings


JamieMagee:
This is an overarching issue to track analyzer warnings that are currently silenced. To fix a single warning:

1. Remove the warning silence from [`.editorconfig`](https://github.com/microsoft/sbom-tool/blob/main/.editorconfig)
1. Run a build locally
1. Fix any warnings that are emitted
1. Create a pull request
1. Link it to this issue
1. Once merged check the relevant box below

PR #341, opened 2023-09-01
chore: codestyle and roslyn analyzers best practices

### LINKED ISSUE #331: Migrate from `FxCopAnalyzers` to `NetAnalyzers`
When building the project, the following warning is output to the console?

>FxCopAnalyzers package has been deprecated in favor of 'Microsoft.CodeAnalysis.NetAnalyzers', that ships with the .NET SDK. Please refer to https://docs.microsoft.com/visualstudio/code-quality/migrate-from-fxcop-analyzers-to-net-analyzers to migrate to .NET analyzers.

JamieMagee:
The `Microsoft.CodeAnalysis.FxCopAnalyzers` package is no longer required as of .NET 5.0 as the analyzers are included in the .NET SDK itself[^1]. Closes #331.

This change adds the `.editorconfig` from Component Detection[^2] that comes from the .NET Boxed template[^3].

Other changes to MSBuild configuration include:

- [`EnforceCodeStyleInBuild`][2] set to `true`
- [`AnalysisLevel`][3] set to `latest`
- [`AnalysisMode`][4] set to `AllEnabledByDefault`
- [`TreatWarningsAsErrors`][5] set to `true`
  - Set the `severity` of every rule that currently fails to `suggestion` so as not to break the build
  - Errors will be addressed in followup PRs and #340 
- Moved `TargetFramework` from each individual `*.csproj` to `Directory.Build.props`

I plan to resolve some of the errors in followup PRs but wanted to minimize the size of this one to enable easier review.

[2]: https://docs.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#enforcecodestyleinbuild
[3]: https://docs.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#analysislevel
[4]: https://docs.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#analysismode
[5]: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-options/errors-warnings#treatwarningsaserrors

[^1]: https://docs.microsoft.com/en-us/visualstudio/code-quality/migrate-from-fxcop-analyzers-to-net-analyzers?view=vs-2022
[^2]: https://github.com/microsoft/component-detection/blob/main/.editorconfig
[^3]: https://github.com/Dotnet-Boxed/Templates/tree/main/Source/NuGetTemplate


global.json
```diff
@@ -1,6 +1,7 @@
 {
-	"sdk": {
-		"version": "8.0.100-preview.7.23376.3",
-		"rollForward": "latestMajor"
-	}
+  "sdk": {
+    "version": "6.0.400",
```
ryanbrandenburg, line 3: I understand not wanting to use the preview version, but why go back to 6.x.x? Maybe we could bump this to `7.0.400` unless there's a specific issue with that?

ryanbrandenburg (approved):
Looks good, one question about SDK version.

global.json
JamieMagee: I accidentally checked in the 8.0 preview yesterday. I agree that 7.0 makes sense, but I want to do a straight revert, and fast-follow with the upgrade to 7.0

PR #347, opened 2023-09-01
refactor: fix IDE0073 require file header

JamieMagee:
[IDE0073: Require file header](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0073) (The description in `.editorconfig` is incorrect)

Part of #340

This change adds the license header to files that are currently missing it, and also removes our custom `Verify-LicenseHeader.ps1` script. It's clear that it didn't cover everything, as we were missing some files.


ryanbrandenburg (approved):
This looks great to me, but I'm curious where the header to require is specified?

JamieMagee:
> This looks great to me, but I'm curious where the header to require is specified?

It's in `.editorconfig`

https://github.com/microsoft/sbom-tool/blob/9e9a9183a16e76b4917a30bff4d635e17f47e9d5/.editorconfig#L459-L460

PR #360, opened 2023-09-06
Add telemetry and error handling to license retrieval.

sebasgomez238:


src/Microsoft.Sbom.Api/Output/Telemetry/TelemetryRecorder.cs
```diff
@@ -263,7 +274,8 @@ public async Task FinalizeAndLogTelemetryAsync()
                 Parameters = Configuration,
                 SBOMFormatsUsed = sbomFormatsUsed,
                 Switches = this.switches,
-                Exceptions = this.exceptions.ToDictionary(k => k.GetType().ToString(), v => v.Message)
+                Exceptions = this.exceptions.GroupBy(e => e.GetType().ToString()).ToDictionary(group => group.Key, group => group.First().Message),
```
sebasgomez238, line 277: Changed this because adding 2 identical exceptions would throw an error.

src/Microsoft.Sbom.Api/Exceptions/ClearlyDefinedResponseNotSuccessfulException.cs
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+
+namespace Microsoft.Sbom.Api.Exceptions;
+
+/// <summary>
+/// Exception during argument validation used to indicate when we don't have access to a path passed as argument.
```
ryanbrandenburg, line 9: Seems like the descriptions of these exceptions are all the same. I'd rather have no doc-comments than ones that are misleading or vague.

test/Microsoft.Sbom.Api.Tests/Executors/LicenseInformationFetcherTests.cs
```diff
@@ -16,19 +17,20 @@ namespace Microsoft.Sbom.Api.Executors.Tests;
...
 {
     private readonly Mock<ILogger> mockLogger = new Mock<ILogger>();
+    private readonly Mock<IRecorder> mockRecorder = new Mock<IRecorder>();
     private readonly Mock<HttpClient> mockHttpClient = new Mock<HttpClient>();
     private Mock<LicenseInformationService> mockLicenseInformationService;
 
     [TestInitialize]
     public void Setup()
     {
-        mockLicenseInformationService = new Mock<LicenseInformationService>(mockLogger.Object, mockHttpClient.Object);
```
ryanbrandenburg, line 25: Rather than create mocks just to pass to a mock I tend to make an `ILicenseInformationService` which is what gets passed around. That also frees you to make `LicenseInformationService` into `ClearlyDefinedLicenseInformationService` and given the proper abstractions have other stores be configurable drop-ins. It can sometimes get a bit "Java-y" but I find that if you think about what methods go on the interface (and what parameters they have) it can help you realize what the right abstractions are.

test/Microsoft.Sbom.Api.Tests/Executors/LicenseInformationFetcherTests.cs
```diff
@@ -16,19 +17,20 @@ namespace Microsoft.Sbom.Api.Executors.Tests;
 public class LicenseInformationFetcherTests
 {
     private readonly Mock<ILogger> mockLogger = new Mock<ILogger>();
```
ryanbrandenburg, line 19: I generally advise that people only create Mocks with `MockBehavior.Strict`, because otherwise we might not be actually validating anything about what the mock does. That makes Mock's particularly annoying for things like ILogger, which is why we [created the NoopLogger](https://dev.azure.com/mseng/1ES/_git/SbomStorageService?path=/SbomStorageApi.Test/NoopLogger.cs).

ryanbrandenburg (approved):
A couple suggestions, but nothing to block on.

src/Microsoft.Sbom.Api/Executors/LicenseInformationFetcher.cs
```diff
@@ -101,9 +106,18 @@ public async Task<List<string>> FetchLicenseInformationAsync(List<string> listOf
                     extractedLicenses.TryAdd($"{packageName}@{packageVersion}", declaredLicense);
                 }
             }
+
+            // Filter out undefined licenses.
+            foreach (var kvp in extractedLicenses.Where(kvp => kvp.Value == "NOASSERTION" || kvp.Value == "Unlicense" || kvp.Value == "OTHER").ToList())
```
edgarrs, line 111: Do we need this check to be case-insensitive?

src/Microsoft.Sbom.Api/Exceptions/ClearlyDefinedResponseNotSuccessfulException.cs
sebasgomez238: Looks like I accidentally left the old description when I copied the class from one of our other exceptions. I'll change these to the appropriate descriptions.

src/Microsoft.Sbom.Api/Executors/LicenseInformationFetcher.cs
sebasgomez238: Not necessarily but I've added it to be safe.

Issue #32, opened 2022-06-27
Gendocs build hangs


aasim:
Make any chnages in the IConfiguration.cs file, and that will trigger the gendocs build, however the build currently hangs without showing any error.

Running locally on Ubuntu, the build fails with this error, the package is already present in the feed: 
```
/usr/share/dotnet/sdk/3.1.420/Sdks/Microsoft.NET.Sdk/targets/Microsoft.PackageDependencyResolution.targets(241,5): error NETSDK1064: Package Microsoft.CodeAnalysis.VersionCheckAnalyzer, version 3.3.0 was not found. It might have been deleted since NuGet restore. Otherwise, NuGet restore might have only partially completed, which might have been due to maximum path length restrictions. 
```

ByAgenT:
This hangs because of the same reason as #56. .NET 6 migration didn't fix the issue, so future path forward is to work with PowerArgs to push a new version w/ a fix or just remove PowerArgs from the project. Besides that, gendocs pipeline needs complete rehaul and it's probably better to create a script to update docs locally rather than using a pipeline for that, so not enough reasons for me to spend more time on it further.

PR #371, opened 2023-09-07
Update to Help Options

Charles1992Sail:
Updated to add:

1) "Validate -options" and all of the arguments.

2) Under "Generate -options" between "GenerationTimestamp (-gt)" and "Verbosity (-V)"        

    DeleteManifestDirIfPresent (-D)           If set to true, we will delete any previous manifest
                                              directories that are already present in the
                                              ManifestDirPath without asking the user for
                                              confirmation. The new manifest directory will then
                                              be created at this location and the generated SBOM
                                              will be stored there.


Charles1992Sail:

JamieMagee:
Thank you for the PR.

Interestingly, we do have [a workflow to update the docs automatically][1], but it [appears to have been manually disabled][2]

[1]: https://github.com/microsoft/sbom-tool/blob/main/.github/workflows/gen-docs.yml
[2]: https://github.com/microsoft/sbom-tool/actions/workflows/gen-docs.yml

JamieMagee:
@sebasgomez238 message me to let me know it's disabled because of #32. In the long-term, I would like to get the automation working, but in the short-term this is great!

JamieMagee (approved):

Charles1992Sail:
Thanks!  That fix was quick!  Bugs like https://github.com/microsoft/sbom-tool/issues/32 will do it for sure.

PR #387, opened 2023-09-19
Large Stream Json Library

ryanbrandenburg:
For my hack-a-thon project I split out the current logic we use to avoid putting the entire array of objects from parsing json in memory at once into its own library to make it easier to make changes to our existing implementation (SPDX2.2), future implementations (SPDX3.0) and anyone else who may be interested (I know Jamie was talking about another team doing similar stuff).


src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Utils/SPDXToSbomFormatConverterExtensions.cs
```diff
@@ -39,8 +47,20 @@ internal static SbomFile ToSbomFile(this SPDXFile spdxFile)
     /// </summary>
     /// <param name="spdxPackage"></param>
     /// <returns></returns>
-    internal static SbomPackage ToSbomPackage(this SPDXPackage spdxPackage)
+    public static SbomPackage ToSbomPackage(this SPDXPackage spdxPackage)
     {
+        if (spdxPackage.FilesAnalyzed &&
```
edgarrs, line 52: Is this a new validation? Or has it been there for a while?

src/JsonAsynchronousNodeKit/Constants.cs
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace JsonAsynchronousNodeKit;
```
edgarrs, line 4: Do you have UTs specific to this logic? I see you have lots of test coverage for the integration with SPDX parser already, but I don't see UTs for this project

edgarrs (commented):
I left you few questions

src/JsonAsynchronousNodeKit/Constants.cs
ryanbrandenburg: We have some UT's in this solution in the form of classes like [SbomPackageParserTests.cs](https://github.com/microsoft/sbom-tool/pull/387/files#diff-79aea1de84e6a25030c12ea5b285e31292963c3eff0cf01ef44b6904d953339d). This PR does not include direct tests of `LargeJsonParser` outside of `LargeJsonParser_RequiresFullEnumeration`.

src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Utils/SPDXToSbomFormatConverterExtensions.cs
```diff
@@ -14,18 +16,24 @@ namespace Microsoft.Sbom.Utils;
...
     /// <summary>
     /// Converts a <see cref="SPDXFile"/> object to a <see cref="SbomFile"/> object.
     /// </summary>
     /// <param name="spdxFile"></param>
     /// <returns></returns>
-    internal static SbomFile ToSbomFile(this SPDXFile spdxFile)
+    public static SbomFile ToSbomFile(this SPDXFile spdxFile)
     {
+        var checksums = spdxFile.FileChecksums?.Select(c => c.ToSbomChecksum());
+        if (checksums.All(c => c.Algorithm != AlgorithmName.SHA256))
```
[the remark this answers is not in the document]
ryanbrandenburg: Ah good catch.

src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Utils/SPDXToSbomFormatConverterExtensions.cs
ryanbrandenburg: This was previously accomplished via https://github.com/microsoft/sbom-tool/blob/ef03e2b22e84f6dc7e1afd5c25bd4b7b6e674987/src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Parser/SbomPackageParser.cs#L107

src/JsonAsynchronousNodeKit/LargeJsonParser.cs
```diff
@@ -0,0 +1,281 @@
...
+            default:
+                throw new InvalidOperationException($"Unknown {nameof(ParameterType)}: {handler.Type}");
+        }
+
+        return new ParserStateResult(propertyName, result, ExplicitField: true, YieldReturn: handler.Type == ParameterType.Array);
+    }
+
+    private ParserStateResult HandleExtraProperty(ref Utf8JsonReader reader, string propertyName)
+    {
+        // We skip objects and arrays to avoid having to parse them.
```
ByAgenT, line 183: I'm getting a bit lost here. It says that objects and arrays are skipped, but lines 190 and 191 calls `ParseArray` and `ParseObject`. Do you mean we skip parsing to specific object/enumerable, but we still have to parse to `JsonObject` and `JsonArray` in order to traverse them?

src/JsonAsynchronousNodeKit/LargeJsonParser.cs
```diff
@@ -0,0 +1,281 @@
...
+                var i = reader.GetInt32();
+                result = i;
+                break;
+            case ParameterType.Object:
+                var objType = handler.GetType().GetGenericArguments()[0];
+                result = this.GetObject(objType, ref reader);
+                break;
+            case ParameterType.Array:
+                var aryType = handler.GetType().GetGenericArguments()[0];
+                result = this.ParseArray(ref reader, aryType);
```
ByAgenT, line 172: nit
```suggestion
                result = this.ParseArray(ref reader, arrType);
```

src/JsonAsynchronousNodeKit/LargeJsonParser.cs
ByAgenT: If I understand it correctly, maybe it would be more clear?
```suggestion
        // We skip objects and arrays to avoid having to deserialize them.
```

src/JsonAsynchronousNodeKit/LargeJsonParser.cs
```diff
@@ -0,0 +1,281 @@
...
+        var jsonObject = ParserUtils.ParseObject(this.stream, ref this.buffer, ref reader);
+        object? result = jsonObject;
+        if (type != typeof(JsonNode))
+        {
+            result = jsonObject.Deserialize(type, this.jsonSerializerOptions);
+        }
+
+        if (result is null)
+        {
+            throw new NotImplementedException();
```
ByAgenT, line 276: Maybe throw InvalidOperationException here instead?

src/JsonAsynchronousNodeKit/PropertyHandler{T}.cs
```diff
@@ -0,0 +1,7 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace JsonAsynchronousNodeKit;
+
+public record PropertyHandler<T>(ParameterType ParameterType)
```
ByAgenT, line 6: What do you think about moving this into `PropertyHandler.cs`? I'm not a fan of having separate file for just one line :)

src/JsonAsynchronousNodeKit/LargeJsonParser.cs
```diff
@@ -0,0 +1,281 @@
...
+
+        try
+        {
+            var reader = new Utf8JsonReader(this.buffer, isFinalBlock: this.isFinalBlock, this.readerState);
+
+            if (!this.isParsingStarted)
+            {
+                ParserUtils.SkipNoneTokens(this.stream, ref this.buffer, ref reader);
+
+                ParserUtils.AssertTokenType(this.stream, ref reader, JsonTokenType.StartObject);
```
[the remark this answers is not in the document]
ryanbrandenburg: It's true that this doesn't handle a root-level-array but that's semi-intentional since for that scenario you're better off using `System.Text.Json.JsonSerializer.DeserializeAsyncEnumerable`. Since there's already a way to accomplish that and it makes the code simpler to have a grounded start to the parsing I decided not to support root-level arrays.

I'll add a comment a special exception to note this.

src/JsonAsynchronousNodeKit/LargeJsonParser.cs
```diff
@@ -0,0 +1,281 @@
...
+            }
+            else if (this.previousResultState is not null && this.previousResultState.YieldReturn && reader.TokenType == JsonTokenType.EndArray)
+            {
+                // yield returning json arrays means we can't pass it the same Utf8JsonReader ref, so we need to create a new one.
+                // BUT when we do that we end up consuming the next token, so we need to leave it in the array case to be eatten by the next caller.
+                ParserUtils.Read(this.stream, ref this.buffer, ref reader);
+            }
+
+            // If the end of the Json Object is reached, return null to indicate completion.
+            if (reader.TokenType == JsonTokenType.EndObject)
```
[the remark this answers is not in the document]
ryanbrandenburg: IIRC this is an artifact of a slightly different architecture for the `Next` method, though it shouldn't hurt anything. I can remove it for clarity.

src/JsonAsynchronousNodeKit/LargeJsonParser.cs
```diff
@@ -0,0 +1,281 @@
...
+            {
+                return null;
+            }
+
+            ParserUtils.AssertTokenType(this.stream, ref reader, JsonTokenType.PropertyName);
+            var propertyName = reader.GetString() ?? throw new NotImplementedException();
```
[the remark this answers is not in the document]
ryanbrandenburg: The null-coalescing is to coerse the type from string? to string (I don't approve of null-forgiveness if it can be avoided, though if that's the style in this repo I will abide by it), I'll change NotImplementedException to InvalidOperationException and give it a sensible title (that should never be seen.

src/JsonAsynchronousNodeKit/PropertyHandler{T}.cs
ryanbrandenburg: I'll have to suppress SA1402 but I think it makes sense in this instance. The Type-less version only exists so we can have a base-type to give the dictionary.

src/JsonAsynchronousNodeKit/LargeJsonParser.cs
ryanbrandenburg: Good catch. I had originally been going to do ParserUtils.SkipProperty for "extra" array parameters, but I ended up deciding it made more sense to make people explicitly skip the properties they didn't want since otherwise it's difficult to explain why we don't return extra arrays but DO return extra objects. That being the case I think the solution is to just remove this outdated comment entirely.

src/JsonAsynchronousNodeKit/Exceptions/ParserException.cs
```diff
@@ -1,10 +1,9 @@
-﻿// Copyright (c) Microsoft. All rights reserved.
+// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-using System;
 using System.Runtime.Serialization;
 
-namespace Microsoft.Sbom.Exceptions;
+namespace JsonAsynchronousNodeKit.Exceptions;
```
ryanbrandenburg, line 6: I should change the namespace to start with `Microsoft.`

ByAgenT (approved):

PR #401, opened 2023-09-25
Update Microsoft.Sbom.Api README.md

sebasgomez238:


src/Microsoft.Sbom.Api/README.md
```diff
@@ -1,9 +1,27 @@
...
 
 #### Scan Sample
 ```C#
-            var generator = new SBOMGenerator();
+using Microsoft.Extensions.Hosting;
+using Microsoft.Sbom.Contracts;
+
+namespace SBOMApiExample
+{
+    public class GenerationService: IHostedService
```
JamieMagee, line 13: Why change this from a generic example to an ASP.NET specific example?

src/Microsoft.Sbom.Api/README.md
```diff
@@ -24,20 +42,60 @@
                 Verbosity = System.Diagnostics.Tracing.EventLevel.Verbose,
             };
 
-            await generator.GenerateSBOMAsync(scanPath, scanPath, metadata, null, configuration, outputPath);
+            await Task.Run(async () =>
```
JamieMagee, line 45: Is `Task.Run` necessary? The general advice is that it's only a good practice for CPU-bound code[^1]

[^1]: https://blog.stephencleary.com/2013/10/taskrun-etiquette-and-proper-usage.html

src/Microsoft.Sbom.Api/README.md
sebasgomez238: The old example showed the generator being instantiated without arguments. This was possible when we were using ninject in v0.3.3. Now that we are using Microsoft Dependency injection libraries these arguments need to be passed into the SBOMGenerator object. It is technically possible for users to do this manually but would be much more complex than this approach.

src/Microsoft.Sbom.Api/README.md
sebasgomez238: You're right it's not necessary. Fixed.

src/Microsoft.Sbom.Api/README.md
JamieMagee: My issue wasn't about the arguments, but `IHostedService` is ASP.NET. Generating SBOMs in ASP.NET is possible, but I wouldn't see it as a common scenario.

ryanbrandenburg (approved):

PR #422, opened 2023-10-10
Target net6.0 and net8.0

sebasgomez238:
Could use review on a few comments. This PR also updates the sdk to 8.0.100-rc.2.23502.2 in order to pick up some bug fixes. There are also a few miscellaneous fixes for IDE errors that popped up when I opened the file for the first time.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -811,10 +811,6 @@ dotnet_diagnostic.CA1852.severity = suggestion
 # CA1854: Prefer 'TryGetValue' over 'ContainsKey' and 'Item' when accessing dictionary items
 dotnet_diagnostic.CA1854.severity = suggestion
 
-# https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1859
-# CA1859: Use culture-aware string operations
-dotnet_diagnostic.CA1859.severity = suggestion
-
 # https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1860
 # CA1860: Avoid using 'Enumerable.Any()' extension method.
 dotnet_diagnostic.CA1860.severity = suggestion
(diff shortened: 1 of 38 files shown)


src/Microsoft.Sbom.Api/Exceptions/ComponentDetectorException.cs
```diff
@@ -26,6 +26,9 @@ public ComponentDetectorException(string message, Exception innerException)
     {
     }
 
+#if NET8_0_OR_GREATER
+    [Obsolete(DiagnosticId = "SYSLIB0051")] // SerializationInfo type is obsolete as of .NET8
```
sebasgomez238, line 30: Documentation on this error suggests this as the fix if you NEED the exception to be serializable. I after testing I couldn't find any issues by removing the [Serializable] attribute and the corresponding method. Could use suggestions on which approach to take in case I'm missing something.

src/Microsoft.Sbom.Extensions/Exceptions/MissingHashValueException.cs
```diff
@@ -19,9 +18,4 @@ public MissingHashValueException(string message)
 
     public MissingHashValueException(string message, Exception inner)
         : base(message, inner) { }
-
-    protected MissingHashValueException(
-        SerializationInfo info,
-        StreamingContext context)
-        : base(info, context) { }
 }
```
sebasgomez238, line 21: This is the other approach to handling the error mentioned above.

.editorconfig
```diff
@@ -474,6 +474,16 @@ dotnet_diagnostic.CA1859.severity = none
 # https://learn.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu5104
 dotnet_diagnostic.NU5104.severity = none
 
+# SYSLIB1045: Use 'GeneratedRegexAttribute' to generate the regular expression implemenatation at compile-time
+# Justification: Would not have much impact on us / is only available in .NET8
+# https://learn.microsoft.com/dotnet/fundamentals/syslib-diagnostics/syslib1045
+dotnet_diagnostic.SYSLIB1045.severity = none
```
sebasgomez238, line 480: A few of the analyzers being suppressed here are because they involve using functions exclusive to .NET8. My thought process is that splitting up the logic would be harder to maintain. Let me know if this is the wrong approach. The other errors being suppressed here come from #406 which I had reverted but just brought back.

src/Microsoft.Sbom.Api/Exceptions/ComponentDetectorException.cs
JamieMagee: Reading the documentation[^1], it seems like we can likely drop this altogether:

> If you created a custom type derived from [System.Exception](https://learn.microsoft.com/en-us/dotnet/api/system.exception), consider whether you really need it to be serializable. It's likely that you don't need it to be serializable, as exception serialization is primarily intended to support remoting, and support for remoting was dropped in .NET Core 1.0. 

[^1]: https://learn.microsoft.com/en-us/dotnet/fundamentals/syslib-diagnostics/syslib0051

src/Microsoft.Sbom.Common/UnixFileSystemUtils.cs
sebasgomez238, line 1: Key changes to fix ARM issues.

PR #425, opened 2023-10-11
Update ValidateAsync & throw exception if SBOM not found

micyunmsft:
Updates SBOMValidator's recently added ValidateAsync method to return a list of errors rather than just true/false whether the validation succeeded or not. Also updates behavior of validation such that an exception is thrown if the SBOM is not found. Previously, if the SBOM wasn't found, validation would return false but nothing would be written to the output file, making it hard to tell what the problem was.

How validated:
Tested locally by giving sbom-tool a folder with no SBOM and tried validating a few SBOMs where file content validation fails (missing file, etc.).


src/Microsoft.Sbom.Api/Workflows/SBOMParserBasedValidationWorkflow.cs
```diff
@@ -171,6 +171,13 @@ public async Task<bool> RunAsync()
                 recorder.RecordException(e);
                 log.Error("Encountered an error while validating the drop.");
                 log.Error($"Error details: {e.Message}");
+
+                if (e is FileNotFoundException || e is DirectoryNotFoundException)
+                {
+                    // Bubble up exception if user provided an invalid path/SBOM wasn't found
+                    throw;
```
edgarrs, line 178: My guess is that this will break the existing behavior. Did you test your changes calling sbom-tool and compare the behavior before and after? Maybe the validation should happen within the new API method and not down here to avoid changing the existing behavior

src/Microsoft.Sbom.Api/SBOMValidator.cs
```diff
@@ -45,7 +45,7 @@ public async Task<bool> ValidateSbomAsync()
         return isSuccess;
     }
 
-    public async Task<bool> ValidateSbomAsync(
+    public async Task<IList<EntityError>> ValidateSbomAsync(
```
edgarrs, line 48: It would be better if you could return a new object that has a property for validation result (true/false) and the list of Errors

edgarrs (commented):

sebasgomez238 (approved):

src/Microsoft.Sbom.Api/SBOMValidator.cs
```diff
@@ -45,7 +57,7 @@ public async Task<bool> ValidateSbomAsync()
         return isSuccess;
     }
 
-    public async Task<bool> ValidateSbomAsync(
+    public async Task<SBOMValidationResult> ValidateSbomAsync(
```
edgarrs, line 60: Unfortunately, using inheritance of SBOMValidationResult to define success and failed validation might not be a good API design pattern. The caller of ValidateSbomAsync would need to check for the type of the SBOMValidationResult object to see if the validation failed or not. Can we add the Errors property to SBOMValidationResult, add also another boolean property to define the Result of the validation and remove the abstract modifier from the class. Then we could get rid of SBOMValidationSuccess and SBOMValidationFailure classes

edgarrs (changes requested):

src/Microsoft.Sbom.Api/SBOMValidator.cs
```diff
@@ -23,16 +26,25 @@ public class SbomValidator : ISBOMValidator
...
+    private readonly IFileSystemUtils fileSystemUtils;
 
     public SbomValidator(
         IWorkflow<SbomParserBasedValidationWorkflow> sbomParserBasedValidationWorkflow,
         IRecorder recorder,
         IEnumerable<ConfigValidator> configValidators,
-        ConfigSanitizer configSanitizer)
+        ConfigSanitizer configSanitizer,
+        IConfiguration configuration,
+        ISbomConfigProvider sbomConfigs,
```
edgarrs, line 39: This is a breaking change, please add a new constructor that takes the new parameters

src/Microsoft.Sbom.Api/SBOMValidator.cs
```diff
@@ -77,12 +89,18 @@ public async Task<bool> ValidateSbomAsync()
 
         inputConfig.ToConfiguration();
 
+        var sbomConfig = sbomConfigs.Get(configuration.ManifestInfo.Value.FirstOrDefault());
```
sebasgomez238, line 92: You should be able to use inputConfig.ManifestPath(). This would remove the need for an ISbomConfigProvider in the constructor.

src/Microsoft.Sbom.Api/SBOMValidator.cs
```diff
@@ -23,16 +26,25 @@ public class SbomValidator : ISBOMValidator
...
+    private readonly IConfiguration configuration;
+    private readonly ISbomConfigProvider sbomConfigs;
+    private readonly IFileSystemUtils fileSystemUtils;
 
     public SbomValidator(
         IWorkflow<SbomParserBasedValidationWorkflow> sbomParserBasedValidationWorkflow,
         IRecorder recorder,
         IEnumerable<ConfigValidator> configValidators,
-        ConfigSanitizer configSanitizer)
+        ConfigSanitizer configSanitizer,
```
sebasgomez238, line 37: We can also remove this configSanitizer. Looks like I accidentally left this in. But it doesn't actually get used anywhere.

src/Microsoft.Sbom.Api/SBOMValidator.cs
micyunmsft: When I try to use this, I get a build error saying that ManifestPath is obsolete.

sebasgomez238: Using inputConfig.ManifestDirPath() instead should work. This will give you the path to the _manifest directory under which the SBOM will always be found.

micyunmsft: But it only gives you the directory, right? Not the full path to the actual manifest file? In the case that the directory exists but the file doesn't exist, I think we would still want to throw an error.

sebasgomez238: You're right, even though this should never be the case. In theory it is possible for the _manifest to be present the spdx not be present. So we can be safe and check for it directly using the sbomConfig in the new constructor.

edgarrs (approved):

PR #436, opened 2023-10-24
Update to latest version of Component-Detection Libraries.

sebasgomez238:
Updates to latest version of cd and adds support for Conan packages.


Directory.Build.props
```diff
@@ -10,7 +10,7 @@
         <AnalysisLevel>latest</AnalysisLevel>
         <AnalysisMode>AllEnabledByDefault</AnalysisMode>
         <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
-        <NoWarn>$(NoWarn);CA1014;CS8002;CA1416</NoWarn>
+        <NoWarn>$(NoWarn);CA1014;CS8002;CA1416;NU5104</NoWarn>
```
sebasgomez238, line 13: This is a warning about the use of a preview package. This is the same Spectre package that CD is using in their latest version. It was needed to be able to access the ScanSettings Type.

src/Microsoft.Sbom.Adapters/Adapters/ComponentDetection/ConanComponentExtensions.cs
```diff
@@ -0,0 +1,45 @@
...
+using Microsoft.ComponentDetection.Contracts.TypedComponent;
+using Microsoft.Sbom.Contracts;
+
+namespace Microsoft.Sbom.Adapters.ComponentDetection;
+
+internal static class ConanComponentExtensions
+{
+    public static SbomPackage? ToSbomPackage(this ConanComponent conanComponent)
+    {
+        var lst = new List<Checksum>();
```
ByAgenT, line 14: Maybe rename to `checksums`?

src/Microsoft.Sbom.Api/Utils/ComponentDetectorCachedExecutor.cs
```diff
@@ -5,7 +5,9 @@
 using System.Collections.Concurrent;
 using System.Threading.Tasks;
 using Microsoft.ComponentDetection.Contracts.BcdeModels;
+using Microsoft.ComponentDetection.Orchestrator.Commands;
 using Serilog;
+using Spectre.Console.Cli;
```
ByAgenT, line 10: I assume you're specifying this `using` to access `ScanSettings` object, but it seems like for `IComponentDetector.cs` only `using Microsoft.ComponentDetection.Orchestrator.Commands` is sufficient. Are you sure we have to add and use `Spectre.Console.Cli` explicitly?

sebasgomez238: It won't let me do ScanSettings.ToString() If I don't include this. Not entirely sure why. I tried to get the hash of the ScanSettings object directly but then the cachedExecutor wasn't working properly.

PR #441, opened 2023-11-08
Extract package info from local package cache for NuGet and Maven

sebasgomez238:
Creates PackageDetailsFactory that will handle calling protocol specific logic for locating/parsing information from local package cache. Relevant telemetry is recorded to the telemetry file and the feature was placed behind a flag that can be passed as (-pm true)


src/Microsoft.Sbom.Adapters/Adapters/ComponentDetection/ScannedComponentExtensions.cs
```diff
@@ -10,31 +10,30 @@
...
 /// </summary>
 public static class ScannedComponentExtensions
 {
     /// <summary>
-    /// Converts a <see cref="ScannedComponentWithLicense"/> to an <see cref="SbomPackage"/>.
+    /// Converts a <see cref="ExtendedScannedComponent"/> to an <see cref="SbomPackage"/>.
     /// </summary>
-    public static SbomPackage? ToSbomPackage(this ScannedComponentWithLicense component, AdapterReport report)
+    public static SbomPackage? ToSbomPackage(this ExtendedScannedComponent component, AdapterReport report)
     {
```
sebasgomez238, line 21: Here I renamed the ScannedComponentWithLicense class because now it holds more than just a license I also now pass the entire component down in order to extract the properties within the individual ToSbomPackage() methods.

src/Microsoft.Sbom.Api/PackageDetails/ComponentDetailsUtils/MavenUtils.cs
```diff
@@ -0,0 +1,201 @@
...
+
+    private readonly string userDefinedLocalRepositoryPath;
+
+    public MavenUtils(IFileSystemUtils fileSystemUtils, ILogger log, IRecorder recorder)
+    {
+        this.fileSystemUtils = fileSystemUtils ?? throw new ArgumentNullException(nameof(fileSystemUtils));
+        this.log = log ?? throw new ArgumentNullException(nameof(log));
+        this.recorder = recorder ?? throw new ArgumentNullException(nameof(recorder));
+
+        userDefinedLocalRepositoryPath = GetLocalRepositoryPath();
```
sebasgomez238, line 38: I moved this call here in order to avoid a call to this method for every Maven component detected. If there is a better way to do this let me know 😄

Directory.Build.props
```diff
@@ -10,7 +10,7 @@
         <AnalysisLevel>latest</AnalysisLevel>
         <AnalysisMode>AllEnabledByDefault</AnalysisMode>
         <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
-        <NoWarn>$(NoWarn);CA1014;CS8002;CA1416</NoWarn>
+        <NoWarn>$(NoWarn);CA1014;CS8002;IDE0290;IDE0300;JSON002;CA1416</NoWarn>
```
sebasgomez238, line 13: Will address these unrelated warnings after PR is approved to avoid noise in PR

src/Microsoft.Sbom.Api/PackageDetails/ComponentDetailsUtils/MavenUtils.cs
```diff
@@ -0,0 +1,187 @@
...
+
+        try
+        {
+            var pomBytes = fileSystemUtils.ReadAllBytes(pomLocation);
+            using var pomStream = new MemoryStream(pomBytes, false);
+
+            var doc = new XmlDocument();
+            doc.Load(pomStream);
+
+            XmlNode developersNode = doc["project"]?["developers"];
```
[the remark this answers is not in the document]
ryanbrandenburg: Looking below it seems we DO expect at least some of these to actually be null so you can ignore the above (except for the part about `#nullable enable`)

src/Microsoft.Sbom.Api/PackageDetails/ComponentDetailsUtils/MavenUtils.cs
```diff
@@ -0,0 +1,187 @@
...
+            }
+            else if (developersNode != null)
+            {
+                // Take the first developer name and use it as the supplier when there is no organization listed.
+                var developerName = developersNode["developer"]?["name"]?.InnerText;
+                if (!string.IsNullOrEmpty(developerName))
+                {
+                    supplierField = $"Person: {developerName}";
+                }
+            }
```
ryanbrandenburg, line 121: Should we have a final case here, perhaps at least a debug for "We don't know who produced this"?

src/Microsoft.Sbom.Api/PackageDetails/ComponentDetailsUtils/MavenUtils.cs
```diff
@@ -0,0 +1,187 @@
...
+                    if (!string.IsNullOrEmpty(licenseName))
+                    {
+                        licenseField = licenseName;
+                    }
+                }
+            }
+
+            return (name, version, new PackageDetails(licenseField, supplierField));
+        }
+        catch (Exception e)
```
ryanbrandenburg, line 138: Generally when you catch `Exception` instead of something more specific it's good to either justify in a comment why you can't allow failure here or catch something more specific.

src/Microsoft.Sbom.Api/PackageDetails/ComponentDetailsUtils/MavenUtils.cs
```diff
@@ -0,0 +1,187 @@
...
+        }
+    }
+
+    /// <summary>
+    /// Gets the local repository path for the Maven protocol. Returns null if a settings.xml is not found.
+    /// </summary>
+    /// <returns>The path to the local repository path defined in the settings.xml</returns>
+    private static string GetLocalRepositoryPath()
+    {
+        var m2Path = $"{HomePath}/.m2";
```
ryanbrandenburg, line 153: Path.Join (and below).

src/Microsoft.Sbom.Api/PackageDetails/ComponentDetailsUtils/NugetUtils.cs
```diff
@@ -0,0 +1,114 @@
...
+    {
+        var nuspecLocation = string.Empty;
+
+        var componentName = scannedComponent.Component.PackageUrl?.Name.ToLower();
+        var componentVersion = scannedComponent.Component.PackageUrl?.Version;
+        var componentType = scannedComponent.Component.PackageUrl?.Type.ToLower();
+
+        if (componentType == "nuget")
+        {
+            nuspecLocation = $"{NugetPackagesPath}/{componentName}/{componentVersion}/{componentName}.nuspec";
```
ryanbrandenburg, line 45: Looks like you can apply most of my comments from MavenUtils to this one as well.

src/Microsoft.Sbom.Api/PackageDetails/PackageDetailsFactory.cs
```diff
@@ -0,0 +1,102 @@
...
+                case ".pom" when !string.IsNullOrEmpty(path):
+                    var pomDetails = mavenUtils.ParsePom(path);
+                    if (!string.IsNullOrEmpty(pomDetails.packageDetails.License) || !string.IsNullOrEmpty(pomDetails.packageDetails.Supplier))
+                    {
+                        packageDetailsDictionary.TryAdd((pomDetails.Name, pomDetails.Version), pomDetails.packageDetails);
+                    }
+
+                    break;
+                default:
+                    break;
```
ryanbrandenburg, line 89: Probably log here.

src/Microsoft.Sbom.Api/PackageDetails/PackageDetailsFactory.cs
```diff
@@ -0,0 +1,102 @@
...
+
+    private IDictionary<(string Name, string Version), PackageDetails> ExtractPackageDetailsFromFiles(List<string> packageDetailsPaths)
+    {
+        var packageDetailsDictionary = new ConcurrentDictionary<(string, string), PackageDetails>();
+
+        foreach (var path in packageDetailsPaths)
+        {
+            switch (Path.GetExtension(path)?.ToLowerInvariant())
+            {
+                case ".nuspec" when !string.IsNullOrEmpty(path):
```
ryanbrandenburg, line 72: Can we move the `IsNullOrEmpty` check up out of the case statement?

test/Microsoft.Sbom.Api.Tests/Executors/SBOMComponentsWalkerTests.cs
```diff
@@ -23,16 +25,14 @@
...
-using Microsoft.Sbom.Adapters.ComponentDetection;
-
 [TestClass]
 public class SBOMComponentsWalkerTests
 {
     private readonly Mock<ILogger> mockLogger = new Mock<ILogger>();
     private readonly Mock<IConfiguration> mockConfiguration = new Mock<IConfiguration>();
     private readonly Mock<ISbomConfigProvider> mockSbomConfigs = new Mock<ISbomConfigProvider>();
     private readonly Mock<IFileSystemUtils> mockFileSystem = new Mock<IFileSystemUtils>();
+    private readonly Mock<IPackageDetailsFactory> mockPackageDetailsFactory = new Mock<IPackageDetailsFactory>();
```
ryanbrandenburg, line 35: Not for this PR since it's an existing pattern, but when possible it's best to avoid storing mocks on the class since this breaks parallel execution of tests.

ryanbrandenburg (approved):
I have NitPicks but structurally this looks great.

src/Microsoft.Sbom.Api/PackageDetails/ComponentDetailsUtils/MavenUtils.cs
```diff
@@ -0,0 +1,187 @@
...
+        // Check for file permissions on the .m2 directory before attempting to check if the file exists
+        if (MavenPackagesPathHasReadPermissions)
+        {
+            if (fileSystemUtils.FileExists(pomLocation))
+            {
+                return pomLocation;
+            }
+            else
+            {
+                log.Debug($"Pom location could not be found at: {pomLocation}");
```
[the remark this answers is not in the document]
sebasgomez238: This would mean that for the specific package in question there is no pom.xml at the local package cache. This could be either because it doesn't exist or because we're it's not in the place we expect it to be. I'll change this to verbose.

src/Microsoft.Sbom.Api/PackageDetails/ComponentDetailsUtils/MavenUtils.cs
sebasgomez238: At the Generator.cs there is a check for when this is null it gets populated as "NOASSERTION" https://github.com/microsoft/sbom-tool/blob/main/src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Generator.cs#L164 This also applies to other fields such as LicenseDeclared and LicenseConcluded

PR #453, opened 2023-11-22
Add supplier/license info support for rubygems.

sebasgomez238:


src/Microsoft.Sbom.Api/PackageDetails/ComponentDetailsUtils/RubyGemsUtils.cs
```diff
@@ -0,0 +1,278 @@
...
+    {
+        this.fileSystemUtils = fileSystemUtils ?? throw new ArgumentNullException(nameof(fileSystemUtils));
+        this.log = log ?? throw new ArgumentNullException(nameof(log));
+        this.recorder = recorder ?? throw new ArgumentNullException(nameof(recorder));
+
+        this.rubyGemsPath = GetRubyGemsSpecificationsPath();
+    }
+
+    // Takes in a scanned component and attempts to find the associated gemspec file. If it is not found then it returns null.
+    public string? GetMetadataLocation(ScannedComponent scannedComponent)
```
sebasgomez238, line 41: Having a hard time coming up with a way to unit test this since it relies on finding the gem executable which may not be present in test environments.

src/Microsoft.Sbom.Adapters/Adapters/ComponentDetection/NuGetComponentExtensions.cs
```diff
@@ -25,10 +25,10 @@ public static SbomPackage ToSbomPackage(this NuGetComponent nuGetComponent, Exte
         PackageName = nuGetComponent.Name,
         PackageVersion = nuGetComponent.Version,
         Supplier = nuGetComponent.Authors?.Any() == true ? $"Organization: {nuGetComponent.Authors.First()}" : component.Supplier,
```
ryanbrandenburg, line 27: Personal preference, but when predicates and assignments start getting longer like this I tend to prefer putting them in their own variable first since it makes it easier to tell what's going on in the debugger.

src/Microsoft.Sbom.Adapters/Adapters/ComponentDetection/RubyGemsComponentExtensions.cs
```diff
@@ -24,9 +24,11 @@ public static SbomPackage ToSbomPackage(this RubyGemsComponent rubyGemsComponent
         PackageName = rubyGemsComponent.Name,
         PackageVersion = rubyGemsComponent.Version,
         PackageSource = rubyGemsComponent.Source,
-        LicenseInfo = string.IsNullOrWhiteSpace(component.LicenseConcluded) ? null : new LicenseInfo
+        Supplier = string.IsNullOrEmpty(component.Supplier) ? null : $"Person: {component.Supplier}",
```
ryanbrandenburg, line 27: Is there a standard we're trying to follow here? `Person` implies to me that there's an individual human on the right side of the `:`, when it's more likely that it's a corporation or other entity.

src/Microsoft.Sbom.Api/PackageDetails/ComponentDetailsUtils/RubyGemsUtils.cs
```diff
@@ -0,0 +1,291 @@
...
+
+/// <summary>
+/// Utilities for retrieving information from rubygems packages that may not be present on the buildDropPath.
+/// </summary>
+public class RubyGemsUtils : IPackageManagerUtils<RubyGemsUtils>
+{
+    private const int ExecutableIndex = 1;
+
+    private readonly IFileSystemUtils fileSystemUtils;
+    private readonly ILogger log;
```
ryanbrandenburg, line 28: Nit: `logger` instead of `log` (which would imply to me that it's a single log message).

src/Microsoft.Sbom.Api/PackageDetails/ComponentDetailsUtils/RubyGemsUtils.cs
```diff
@@ -0,0 +1,291 @@
...
+    private readonly string rubyGemsPath;
+    private string[] potentialGemEnvPaths;
+
+    public RubyGemsUtils(IFileSystemUtils fileSystemUtils, ILogger log, IRecorder recorder)
+    {
+        this.fileSystemUtils = fileSystemUtils ?? throw new ArgumentNullException(nameof(fileSystemUtils));
+        this.log = log ?? throw new ArgumentNullException(nameof(log));
+        this.recorder = recorder ?? throw new ArgumentNullException(nameof(recorder));
+
+        this.rubyGemsPath = GetRubyGemsSpecificationsPath();
```
ryanbrandenburg, line 40: I would advise not calling methods that do `Process.Start` from the constructor. There's a lot of failure-cases and caveats for `Process.Start` and people generally treat a constructor more reliable than that. Maybe make a property for RubyGemsPath which calls GetRubyGemsSpecificationsPath if the underlying variable hasn't already been populated.

src/Microsoft.Sbom.Api/PackageDetails/ComponentDetailsUtils/RubyGemsUtils.cs
```diff
@@ -0,0 +1,291 @@
...
+
+        if (string.IsNullOrEmpty(gemspecLocation))
+        {
+            return null;
+        }
+
+        var componentName = scannedComponent.Component.PackageUrl?.Name.ToLower();
+        var componentVersion = scannedComponent.Component.PackageUrl?.Version;
+
+        var fullGemspecPath = Path.Join(gemspecLocation, $"{componentName}-{componentVersion}.gemspec");
```
ryanbrandenburg, line 56: Is this format a requirement or a convention?

src/Microsoft.Sbom.Api/PackageDetails/ComponentDetailsUtils/RubyGemsUtils.cs
```diff
@@ -0,0 +1,291 @@
...
+        if (fileSystemUtils.FileExists(fullGemspecPath))
+        {
+            return fullGemspecPath;
+        }
+        else
+        {
+            // Enumerate directories in the gemspec location and then search for the gemspec file in each directory
+            var directories = fileSystemUtils.GetDirectories(gemspecLocation);
+
+            foreach (var directory in directories)
```
ryanbrandenburg, line 67: Are there any static directory names we ought to ignore or ways we can pare the number of directories we search down?

src/Microsoft.Sbom.Api/PackageDetails/ComponentDetailsUtils/RubyGemsUtils.cs
```diff
@@ -0,0 +1,291 @@
...
+        return null;
+    }
+
+    public ParsedPackageInformation ParseMetadata(string gemspecLocation)
+    {
+        var supplierField = string.Empty;
+        var licenseField = string.Empty;
+
+        // Regex patterns for finding the relevant properties of the .gemspec file.
+        var namePattern = @"s.name\s*=\s*""([^""]+)""";
```
ryanbrandenburg, line 87: Are these single-lines delimited by newline? If so I'd suggest something like `fileContent.Split(Newline).Where(l => l.StartsWith("s.name")).Split('=')` (of course you'd need to harden it a bit). When possible I suggest avoiding Regex because it has a lot of caveats and security issues (RegexDoS) that make it hard to maintain and debug.

Even better would be if there was a C# library which already parses GemSpec for us.

src/Microsoft.Sbom.Api/PackageDetails/ComponentDetailsUtils/RubyGemsUtils.cs
```diff
@@ -0,0 +1,291 @@
...
+        var versionPattern = @"s.version\s*=\s*""([^""]+)""";
+        var authorsPattern = @"s.authors\s*=\s*\[([^\]]+)\]";
+        var licensesPattern = @"s\.licens(?:e|es)\s*=\s*\[([^\]]+)\]";
```
ryanbrandenburg, line 90: Why is this `.` escaped but the others are not?

src/Microsoft.Sbom.Api/PackageDetails/ComponentDetailsUtils/RubyGemsUtils.cs
```diff
@@ -0,0 +1,291 @@
...
+                    UseShellExecute = false,
+                    CreateNoWindow = true
+                };
+            }
+
+            using var processFindGem = new Process();
+            processFindGem.StartInfo = processStartFindGem;
+
+            processFindGem.Start();
+            var gemExecutablePath = processFindGem.StandardOutput.ReadToEnd()?.Trim().ToString();
```
ryanbrandenburg, line 208: Have you tested this locally? This code feels race-conditiony.

src/Microsoft.Sbom.Api/PackageDetails/ComponentDetailsUtils/RubyGemsUtils.cs
sebasgomez238: gemspec files are automatically placed in this structure whenever the user installs gems. So I would say this is a requirement.

src/Microsoft.Sbom.Api/PackageDetails/ComponentDetailsUtils/RubyGemsUtils.cs
sebasgomez238: I believe the only nesting that should occur in this scenario is the "default" directory which we could change specifically look for but I figured this was safer in case there are other cases I'm not aware of. But there are none we should ignore.

src/Microsoft.Sbom.Api/PackageDetails/ComponentDetailsUtils/RubyGemsUtils.cs
```diff
@@ -0,0 +1,291 @@
...
+
+            var name = GetMatchesFromPattern(fileContent, namePattern);
+            var version = GetMatchesFromPattern(fileContent, versionPattern);
+
+            // We only take the first author to maintain consistency with what we do with the other component types.
+            var supplierList = GetMatchesFromPattern(fileContent, authorsPattern);
+
+            if (supplierList.Any())
+            {
+                supplierField = supplierList.First().Replace("\".freeze", string.Empty);
```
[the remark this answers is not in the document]
sebasgomez238: Ruby has a required authors field that looks like "s.authors = ["Nobu Nakada".freeze]" After reading the value I get something like "Nobu Nakada\".freeze" so even if the name was "anti.freeze it would be fine because we would end up with "anti.freeze\".freeze"

src/Microsoft.Sbom.Api/PackageDetails/ComponentDetailsUtils/RubyGemsUtils.cs
sebasgomez238: Haven't had issues on either Mac or Windows. Should I add logic to wait for the process to finish + include a timeout?

src/Microsoft.Sbom.Api/PackageDetails/ComponentDetailsUtils/RubyGemsUtils.cs
sebasgomez238: I couldn't find one in C# but there is one in Ruby. We could use something like:

using IronRuby;
using IronRuby.Runtime;
using Microsoft.Scripting.Hosting;

// Create a Ruby engine
var engine = Ruby.CreateEngine();

// Execute the .gemspec file and get the Gem::Specification object
var spec = engine.ExecuteFile("my_gem.gemspec") as RubyObject;

but how much overhead this Ruby engine would add.

ryanbrandenburg: Their last release was 2011 so I wouldn't want to take a dependency on them. Either the Regex or a split-and-search should be fine. If you stick with Regex I'd suggest adding a timeout to the Regex.Match call just as a precaution. These patterns don't look vulnerable to catastrophic backtracking, but I'm not a Regex wizard so I'm not 100% sure.

src/Microsoft.Sbom.Api/PackageDetails/ComponentDetailsUtils/RubyGemsUtils.cs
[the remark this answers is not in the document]
ryanbrandenburg: Ah, I missed the `\"` in the front.

src/Microsoft.Sbom.Api/PackageDetails/ComponentDetailsUtils/RubyGemsUtils.cs
ryanbrandenburg: Definitely supply a timeout. If I understand correctly the standardout.ReadToEnd should only complete once the process completes (one way or another) so I don't THINK additional waiting should be needed.

ryanbrandenburg (commented):
Looks good after my suggestions are addressed.

ryanbrandenburg (approved):

PR #455, opened 2023-12-04
Bump sdk, pin dependency versions.

sebasgomez238:
Bumps sdk to latest release version of dotnet. Pins dependency that was updated through component-detection update to avoid package downgrade errors (Version in question is coming through a Spectre package in the CD repo which we do not use). The update of CD libraries is to pick up this fix: https://github.com/microsoft/component-detection/issues/862. Also adds timer around package metadata parsing.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -21,7 +21,7 @@
         <PackageVersion Include="Microsoft.ComponentDetection.Detectors" Version="$(ComponentDetectionPackageVersion)" />
         <PackageVersion Include="Microsoft.ComponentDetection.Orchestrator" Version="$(ComponentDetectionPackageVersion)" />
         <PackageVersion Include="Microsoft.Extensions.DependencyInjection" Version="7.0.0" />
-        <PackageVersion Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="8.0.0-rc.2.23479.6" />
+        <PackageVersion Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="8.0.0" />
         <PackageVersion Include="Microsoft.Extensions.Hosting" Version="7.0.1" />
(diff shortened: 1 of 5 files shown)


Directory.Packages.props
```diff
@@ -21,6 +21,7 @@
         <PackageVersion Include="Microsoft.ComponentDetection.Detectors" Version="$(ComponentDetectionPackageVersion)" />
         <PackageVersion Include="Microsoft.ComponentDetection.Orchestrator" Version="$(ComponentDetectionPackageVersion)" />
         <PackageVersion Include="Microsoft.Extensions.DependencyInjection" Version="7.0.0" />
+        <PackageVersion Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="8.0.0-rc.2.23479.6" />
```
JamieMagee, line 24: Does this need to be updated upstream? I'd like to avoid using an RC version if `8.0.0` is available.

sebasgomez238: Yes, bumping to 8.0.0 in the meantime.

src/Microsoft.Sbom.Api/PackageDetails/PackageDetailsFactory.cs
```diff
@@ -34,9 +35,12 @@ public PackageDetailsFactory(ILogger log, IRecorder recorder, IPackageManagerUti
 
     public IDictionary<(string Name, string Version), PackageDetails> GetPackageDetailsDictionary(IEnumerable<ScannedComponent> scannedComponents)
     {
-        var packageDetailsLocations = GetPackageDetailsLocations(scannedComponents);
+        using (recorder.TraceEvent(Events.SBOMParseMetadata))
```
sebasgomez238, line 38: Threw in package metadata parsing timer as a part of this PR.

JamieMagee (approved):

PR #472, opened 2024-01-08
Fix IsSuccess return value in SBOMValidator

micyunmsft:
The current return value is flipped from what it should be.


jlperkins (approved):
Thanks for the catch

sfoslund (approved):

PR #475, opened 2024-01-10
Refresh SbomConfigProvider when config values change

micyunmsft:
Currently SbomConfigProvider doesn't pick up updated config values when the global Configuration changes.


jlperkins:
Closing in favor of the type of fix described in #480.

PR #535, opened 2024-04-01
Convert SBOM Tool Main Build to 1ESPT

sfoslund:
Successful test run: https://dev.azure.com/mseng/AzureDevOps/_build/results?buildId=27586716&view=results

### DIFF
--- a/pipelines/sbom-tool-main-build.yaml
+++ b/pipelines/sbom-tool-main-build.yaml
@@ -75,33 +75,33 @@ extends:
               ConnectedServiceName: 'Sbom tool CodeSigning ESRP  service connection'
               FolderPath: '$(Build.SourcesDirectory)/src'
               Pattern: |
-              **/Release/**/*.exe
-              **/Release/**/*.dll
-              UseMinimatch: true
-              signConfigType: inlineSignParams
-              inlineOperation: |
-              [
-                      {
-                          "KeyCode" : "CP-230012",
-                          "OperationCode" : "SigntoolSign",
-                          "Parameters" : {
-                              "OpusName" : "Microsoft",
-                              "OpusInfo" : "http://www.microsoft.com",
-                              "FileDigest" : "/fd \"SHA256\"",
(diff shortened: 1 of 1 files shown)


pipelines/sbom-tool-main-build.yaml
```diff
@@ -0,0 +1,360 @@
...
+                            "OperationCode" : "SigntoolVerify",
+                            "Parameters" : {},
+                            "ToolName" : "sign",
+                            "ToolVersion" : "1.0"
+                        }
+                ]
+            condition: and(succeeded(), startswith(variables['Build.SourceBranch'], 'refs/tags/'))
+
+          - task: DotNetCoreCLI@2
+            displayName: 'Pack NuGet package'
```
pownkel, line 109: This task is disabled (both in the original pipeline and here), do we need it in the yaml?

pownkel:
I noticed that the structure of the published build artifacts is different between the two pipelines - the original puts all binaries in SBOMTool/bin, but the new YAML Pipeline separates the outputs into different directories by platform. Do we expect this to work with our current release pipeline?

sfoslund:
> I noticed that the structure of the published build artifacts is different between the two pipelines - the original puts all binaries in SBOMTool/bin, but the new YAML Pipeline separates the outputs into different directories by platform. Do we expect this to work with our current release pipeline, or will changes also be needed there?

Because of the way 1ESPT publishes artifacts we won't be able to publish all of the artifacts from all of the jobs to the same directory (see [this failed run as an example](https://dev.azure.com/mseng/AzureDevOps/_build/results?buildId=27601213&view=results)). But we will need to update the release pipeline anyways with the following changes:

- Point to the new yaml pipeline instead of the old classic one (which I will be deleting once this PR is merged)
- Account for new placement of SBOM in artifact dir structure
- Account for new placement of linux/ mac artifacts in new artifact dir structure

pownkel (approved):

PR #540, opened 2024-04-09
fix: Improve visibility of logging from inside exception handlers

DaveTryon:
https://github.com/microsoft/dropvalidator/issues/874 calls out some scenarios where error messages were being logged only at Debug verbosity. This PR changes all `ILogger.Debug` calls in exception handlers to `ILogger.Warning` to increase the visibility.


sfoslund (approved):
Changes themselves LGTM. Have you run this version of the tool against any real drops? I would be curious how many of these warnings are surfaced in a normal run, if any

DaveTryon:
> Changes themselves LGTM. Have you run this version of the tool against any real drops? I would be curious how many of these warnings are surfaced in a normal run, if any

As we discussed offline, local testing was against the sbom-tool repo where the change was made

Issue #500, opened 2024-02-13
Disable the `Warning: Manifest directory path was explicitly defined`


jozefizso:
We use the `sbom-tool` to generate SBOM for our .NET project output.

This is the command line call:

```pwsh
sbom-tool generate -b "${{ github.workspace }}/dist" `
    -bc "${{ github.workspace }}/src" `
    -pn "Acme Product Name" `
    -pv "${{ steps.build.outputs.app_full_version }}" `
    -ps "Acme Company" `
    -m "${{ github.workspace }}"
```

It will emit this warning:

```
Warning: Manifest directory path was explicitly defined. Will not attempt to delete any existing _manifest directory.
```

As we have zero warnings policy in the GitHub Actions build pipeline, this will fail our builds.

And the message is not event actionable. We are basically using the tool with normal set of parameters.

DaveTryon:
Fix has been merged and will be included in our next release

PR #542, opened 2024-04-09
fix: Don't throw a warning if an output folder is specified

DaveTryon:
As called out in https://github.com/microsoft/sbom-tool/issues/500, the code returns a warning if the manifest output folder is explicitly specified by the `-m` parameter. Logging a warning for normal and supported operation is bad form. It was causing grief for a team that wanted to control the output location but had a zero-warning policy on their pipeline.

Side note: The default verbosity level for logging (defined at https://github.com/microsoft/sbom-tool/blob/main/src/Microsoft.Sbom.Common/Constants.cs#L16) is `Warning`, so this message will only be displayed if a non-default verbosity is specified. Presumably a higher verbosity would be the first diagnostic step if a problem were to occur, so this feels unlikely to cause any real-world problems.


pownkel (approved):
LGTM

Issue #520, opened 2024-03-05
unexpected input argument will prompt --help but exit_code still be 0


zongtaol:
I am using run_subprocess function in python to invoke sbom-tool generate command. which given invalid input, it will prompt --help function, but exit_code is 0, which run_subprocess believe this command is executed successfully and continue to the next step.

Is there a way to tell the command got unexpected input and throw an exception instead of an exit_code = 0 return?
![Screenshot 2024-03-05 134828](https://github.com/microsoft/sbom-tool/assets/91494184/3a01b808-2d62-49dc-a447-29e5add3bffe)

pownkel:
For some reason I can't reproduce the issue on powershell unless I'm using Python's subprocess.run() method, but it seems to happen consistently on cmd. It looks like the issue is that we [don't explicitly set the exit code](https://github.com/microsoft/sbom-tool/blob/main/src/Microsoft.Sbom.Tool/Program.cs#L40) if the tool encounters an argument parsing error.

DaveTryon:
Fix has been merged and will be included in our next release

PR #544, opened 2024-04-09
fix: Return failing error code on invalid parameter

DaveTryon:
As called out in https://github.com/microsoft/sbom-tool/issues/520, the sbom generation tool is a little inconsistent in how it handles return codes, returning a value of 0 if invalid parameters are passed in. This PR adds 2 lines to explicitly set the return values after displaying help or after successful completion. The successful completion is probably not necessary since we're setting to the default value, but being explicit here is low cost and makes the code (IMO) a little clearer. There was no obvious need to have a specific error code for invalid parameters, so we reuse `ExitCode.GeneralError` for this case. It's trivial to add a different error code if we later decide that we need to expose more information.

Testing was simple--I ran the following script to launch the app and capture the return code. This needs to be a .bat file
```
@echo off
Microsoft.Sbom.DotNetTool.exe
echo %ERRORLEVEL%
```

Before the change, `ERRORLEVEL` is 0. After the change, `ERRORLEVEL` is 1

I also tested it with correct parameters and ensured that `ERRORLEVEL` was correctly set to 0

### DIFF
--- a/src/Microsoft.Sbom.Tool/Program.cs
+++ b/src/Microsoft.Sbom.Tool/Program.cs
@@ -37,6 +37,7 @@ internal class Program
         var result = await Args.InvokeActionAsync<SbomToolCmdRunner>(args);
         if (result.HandledException != null || (result.ActionArgs is not CommonArgs))
         {
+            Environment.ExitCode = (int)ExitCode.GeneralError;
             return;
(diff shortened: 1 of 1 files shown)


pownkel (approved):
LGTM

Issue #444, opened 2023-11-10
Missing logging and false reports in folder permission check


DelphinRP:
Hello everyone,

we are attempting to generate SBOM using sbom-tool in our GitLab jobs using powershell within windows docker container, but receive the following error message:

```
$ dotnet restore ProfiSignal-Neu/ProfiSignal.sln --configfile "${NuGetConfig}" --verbosity quiet
$ dotnet publish ProfiSignal-Neu/PSDesktop/PSDesktopWPF/PSDesktopWPF.csproj --output "wpf_client" --runtime win10-x64 --no-self-contained --version-suffix $CI_COMMIT_SHA --no-restore --configuration Release --verbosity quiet /p:WarningLevel=0 /p:DefineConstants=STABLE
$ dotnet tool install Microsoft.Sbom.DotNetTool --tool-path .tools
You can invoke the tool using the following command: sbom-tool
Tool 'microsoft.sbom.dotnettool' (version '2.0.0') was successfully installed.
 $ .tools\sbom-tool generate -b ${CI_PROJECT_DIR}\wpf_client -bc ${CI_PROJECT_DIR}\ProfiSignal-Neu -pn PS20WPF -pv $version -ps Delphin -nsb https://sbom.delphin.com -V Verbose
[...]

PR #548, opened 2024-04-11
fix: Add logging to Windows permissions checks

DaveTryon:
As called out in https://github.com/microsoft/sbom-tool/issues/444, we need to provide a little more help in debugging some directory permission failures. This PR adds an `ILogger` to the `WindowsFileSystemUtils` class, so that we can log the Exceptions that are caught in that class. 

There was a //TODO in the original code, suggesting that we would log this at `Debug` level. This makes total sense, but I encountered a problem when testing this--the `Config` that exposes the user-configured verbosity level only becomes available _after_ we need to create the `WindowsFileSystemUtils`, and none of the variations I tried were successful in altering that order. Due to this, I opted to log the error at `Warning` level, so that it would show up with the default settings but not show up as the key problem.

Testing
I inserted a test Exception inside the `DirectoryHasRights` method and tested locally to confirm that the Exception appears as a warning in the output. I then removed the test Exception and tested the following scenarios locally to ensure that our "expected" scenarios work without displaying an additional Exception:
- Trying to write to a protected directory (C:\Windows)
- Trying to write to a non-existent directory (C:\NotThere)
- Trying to write to a non-existent network drive (Q:\)
- Trying to write to a non-existent UNC path (\\FOO\BAR)

The net result is that the new warning displays only in cases where an Exception is thrown inside `WindowsFileSystemUtils.DirectoryHasRights`. Flow continues as it did before.

### DIFF
--- a/src/Microsoft.Sbom.Extensions.DependencyInjection/ServiceCollectionExtensions.cs
+++ b/src/Microsoft.Sbom.Extensions.DependencyInjection/ServiceCollectionExtensions.cs
@@ -64,28 +64,11 @@ public static class ServiceCollectionExtensions
     {
         services
             .AddSingleton<IConfiguration, Configuration>()
-            .AddTransient(_ => FileSystemUtilsProvider.CreateInstance())
+            .AddTransient(_ => FileSystemUtilsProvider.CreateInstance(CreateLogger(logLevel)))
             .AddTransient(x =>
             {
                 logLevel = x.GetService<InputConfiguration>()?.Verbosity?.Value ?? logLevel;
-                return Log.Logger = new LoggerConfiguration()
-                    .MinimumLevel.ControlledBy(new LoggingLevelSwitch { MinimumLevel = logLevel })
-                    .Filter.ByExcluding(Matching.FromSource("System.Net.Http.HttpClient"))
-                    .Enrich.With<LoggingEnricher>()
-                    .Enrich.FromLogContext()
-                    .WriteTo.Map(
-                        LoggingEnricher.LogFilePathPropertyName,
-                        (logFilePath, wt) => wt.Async(x => x.File($"{logFilePath}")),
-                        1) // sinkMapCountLimit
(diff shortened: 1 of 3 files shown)


src/Microsoft.Sbom.Extensions.DependencyInjection/ServiceCollectionExtensions.cs
```diff
@@ -64,28 +64,11 @@ public static IServiceCollection AddSbomTool(this IServiceCollection services, L
     {
         services
             .AddSingleton<IConfiguration, Configuration>()
-            .AddTransient(_ => FileSystemUtilsProvider.CreateInstance())
+            .AddTransient(_ => FileSystemUtilsProvider.CreateInstance(CreateLogger(logLevel)))
             .AddTransient(x =>
             {
                 logLevel = x.GetService<InputConfiguration>()?.Verbosity?.Value ?? logLevel;
```
sfoslund, line 70: Do we need to use this line above to determine the log level of the FileSystemUtilsProvider logger too? It seems like the way this is currently written that new logger would be ignoring any user inputted log level

sfoslund (approved):

src/Microsoft.Sbom.Extensions.DependencyInjection/ServiceCollectionExtensions.cs
DaveTryon: My first iteration was exactly what you describe, since I wanted to log the Exception at `Debug` level as suggested in the previously existing TODO comment. Unfortunately, those attempts failed because the lambda was being called on a different `IServiceProvider` than the one that actually contains the user config. This disconnect is why I ended up logging the Exception at `Warning` level, so it would always show up if it occurred.

sfoslund: Makes sense!

Issue #523, opened 2024-03-07
Handle filename collisions in validation


pownkel:
Linux filesystems are case-sensitive, but Windows filesystems are not, and the SBOM validator follows the case sensitivity of whichever filesystem it runs on. As a result, if an SBOM is generated on Linux and contains files called, for example, `filename` and `FileName`, and we attempt to validate that SBOM on Windows, the validator tool sees the same filename twice with different hashes.

Currently if this happens, we [add `null` to a dictionary of file hashes](https://github.com/microsoft/sbom-tool/blob/main/src/Microsoft.Sbom.Api/Executors/ConcurrentSha256HashValidator.cs#L66), but since we [do not check for null when that dictionary is used](https://github.com/microsoft/sbom-tool/blob/main/src/Microsoft.Sbom.Api/Workflows/Helpers/FilesValidator.cs#L97), the user sees the message `Object reference not set to an instance of an object`, which doesn't give any clues about what happened.

When this happens, we should output a more meaningful error, ideally one prompting the user to make sure the validator runs on the same OS that the SBOM was generated on.

DaveTryon:
This has been merged to main and will be included in our next release

PR #550, opened 2024-04-12
fix: Improve case-sensitive handling

DaveTryon:
As called out in #523, we have some issues if a manifest file contains files whose names vary only by case. When this happens, we create a `NullReferenceException` that causes the validator to terminate prematurely without providing information to allow the person to address the issue. This PR does the following:

It leaves the overall validation logic unchanged, but it detects the case that was causing the `NullReferenceException` and reports it as an `AdditionalFile` error. This is consistent with the effect of copying 2 files whose names vary only by case to a file case-insensitive drive. This change allows the validation to complete with errors, rather than exiting prematurely

If errors are found when scanning on a case-insensitive system, it adds a brief warning to inform the users that if they created the manifest on a case-sensitive system, they may need to validate it on a case-sensitive system. We already have a mechanism to identify case-sensitivity, but it was private inside a class. It is now exposed on the associated interface. Unit tests have been updated to account for the case sensitivity check.

I simulated this case locally by generating a manifest, then manually creating some cases where file entries differed by case and by hash. When the validator runs, the order in which the manifest entries is processed is **nondeterministic**. As a result, the same manifest file can produce different errors. Depending on the order of processing, the errors may be reported as invalid hashes or as additional files. Because of this, I made the case sensitivity warning general and displayed it at the top of the output results. If _any_ errors are found when validation is performed on a case-insensitive OS, the error header now reads as follows:

```
------------------------------------------------------------
Individual file validation results
  Note: If the manifest file was originally created using a
        case-sensitive OS, you may also need to validate it
        using a case-sensitive OS.
------------------------------------------------------------
```

There was also a private member in the `OSUtils` class that was not `readonly` and the compiler complained about it. I fixed it while I was touching the file.

### DIFF
--- a/test/Microsoft.Sbom.Api.Tests/Workflows/SbomParserBasedValidationWorkflowTests.cs
+++ b/test/Microsoft.Sbom.Api.Tests/Workflows/SbomParserBasedValidationWorkflowTests.cs
@@ -151,6 +151,8 @@ public class SbomParserBasedValidationWorkflowTests : ValidationWorkflowTestsBas
         var rootFileFilterMock = new DownloadedRootPathFilter(configurationMock.Object, fileSystemMock.Object, mockLogger.Object);
         rootFileFilterMock.Init();
 
+        var osUtilsMock = new Mock<IOSUtils>(MockBehavior.Strict);
+
         var hashValidator = new ConcurrentSha256HashValidator(FileHashesDictionarySingleton.Instance);
         var enumeratorChannel = new EnumeratorChannel(mockLogger.Object);
         var fileConverter = new SbomFileToFileInfoConverter(new FileTypeUtils());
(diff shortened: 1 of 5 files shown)


test/Microsoft.Sbom.Api.Tests/Workflows/SbomParserBasedValidationWorkflowTests.cs
```diff
@@ -151,6 +151,8 @@ public async Task SbomParserBasedValidationWorkflowTests_ReturnsSuccessAndValida
         var rootFileFilterMock = new DownloadedRootPathFilter(configurationMock.Object, fileSystemMock.Object, mockLogger.Object);
         rootFileFilterMock.Init();
 
+        var osUtilsMock = new Mock<IOSUtils>(MockBehavior.Strict);
```
sfoslund, line 154: Thank you for updating the tests when they exist! Hopefully eventually we'll be able to add much more test coverage to this repo.

sfoslund (approved):

PR #551, opened 2024-04-13
fix: Improve logging on a corrupted manifest file

DaveTryon:
As called out in https://github.com/microsoft/dropvalidator/issues/857, it's hard to debug a case where the JSON inside a manifest file is corrupted. When https://github.com/microsoft/dropvalidator/issues/857 was opened, the only logging occurred at `Debug` level and it was not obvious what was happening. Since then, #540 bumped the logging to `Warning` level, which is better but not ideal:

```
##[warning]Encountered an unknown error while enumerating: Expected a '{' token at position 5444 but got String
##[error]Encountered an error while validating the drop.
##[error]Error details: Value cannot be null. (Parameter 'key')
```

The final error is actually thrown by the `Dictionary` class because we're trying to add an entry with a `null` key. This is also the error reported in telemetry.

This PR does the following:
- It defines an `ErrorType` for malformed files
- It silently catches a `ParserException` to distinguish a JSON error from some other random error, and sets the `ErrorType`. It also captures the error message for later handling. (Side note: It does this by stuffing the `Message` into the `Path` parameter, which also avoids the Exception from adding a `null` key to the `Dictionary`
- It detects the presence of the new `ErrorType` and passes the message to the top-level error handling for reporting and recording

With these changes, the output of validating a file with malformed JSON is as follows:
```
##[error]Encountered an error while validating the drop.
##[error]Error details: Your manifest file is malformed. Expected a '{' token at position 5444 but got String
```

The telemetry includes the following:
```
  "Exceptions": {
    "System.IO.InvalidDataException": "Your manifest file is malformed. Expected a \u0027{\u0027 token at position 5444 but got String"
  },
```

### DIFF
--- a/src/Microsoft.Sbom.Api/Entities/ErrorType.cs
+++ b/src/Microsoft.Sbom.Api/Entities/ErrorType.cs
@@ -47,5 +47,8 @@ public enum ErrorType
     NoPackagesFound = 11,
 
     [EnumMember(Value = "Manifest file signing error")]
-    ManifestFileSigningError = 12
+    ManifestFileSigningError = 12,
+
+    [EnumMember(Value = "Invalid input file")]
+    InvalidInputFile = 13
 }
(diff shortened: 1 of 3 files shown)


sfoslund (approved):
Much clearer error message, happy to see this change come in!!

PR #553, opened 2024-04-15
Clean apt cache in Dockerfile to reduce image size

PeterDaveHello:
This is just a minor improvement.


DaveTryon (approved):
This follows a recommended docker best practice at https://docs.docker.com/build/building/best-practices/

DaveTryon:
This was done in #939 because we didn't have a fresh branch

PeterDaveHello:
@DaveTryon With all due respect, I believe that Microsoft will honor credits and contributions from the community rather than just copying their work, even if it's small. You could request a branch refresh, use rebase or cherry-pick, or at least include co-author info in the git commit, but none of them was done. I'm starting to wonder if we should send any pull requests anymore ;)

Issue #505, opened 2024-02-23
Improve component detection error handling


sfoslund:
We have been seeing support requests in which the component detection library throws errors in a scan, usually due to malformed packages present on the machine, and the SBOM tool surfaces those errors before continuing the scan. This can be confusing to users, who see these errors (which are displayed in red, along with a full stack trace) and think that the SBOM generation failed. Further, this can obscure other, more critical errors. 

We should wrap component detection errors such that they are only displayed as warnings when they are not critical and the scan can continue.

DaveTryon:
Component detection logs directly, so doing this will require either that we change component detection or find a way to replace the default implementation of `Serilog.ILogger` with our own version.

DaveTryon:
This has been merged into main and will be included in our next release

PR #554, opened 2024-04-15
fix: Remap errors to warnings if logged inside ComponentDetection

DaveTryon:
As called out in https://github.com/microsoft/sbom-tool/issues/505, `ComponentDetection` will sometimes log things that might be errors in some scenarios, but not in ours. These error messages show up in the log file and potentially confuse the user to pursue the wrong problem. We could either suppress these errors completely or remap them to warnings. This PR takes the latter approach.

Summary of changes:
- Create a new implementation of the `ILogger` interface.
    - This implementation is a decorator around a previously existing `ILogger` implementation
    - This implementation is called `RemapComponentDetectionErrorsToWarningsLogger`
    - This implementation uses `Environment.StackTrace` to detect the presence of "Microsoft.ComponentDetection." in the stack trace, and remaps any log events at `LogEventLevel.Error` to `LogEventLevel.Warning`
- Update the `CreateLogger` method to wrap the existing logger in a `RemapComponentDetectionErrorsToWarningsLogger` object

Tested locally by changing the `LogEventLevel.Error` to `LogEventLevel.Debug` and generating an SBOM. Here is a portion of that output showing the remapping in effect (these lines map as Debug events and are not normally output). It doesn't show in the PR, but the field properties are correctly retained in the logs:

```
##[warning]Gradle dev-only lockfiles ""
##[warning]Gradle dev-only configurations ""
##[warning]Finished applying restrictions to detectors.
##[warning]Removing "NewPipDetector" from active experiments
```

### DIFF
--- a/src/Microsoft.Sbom.Extensions.DependencyInjection/RemapComponentDetectionErrorsToWarningsLogger.cs
+++ b/src/Microsoft.Sbom.Extensions.DependencyInjection/RemapComponentDetectionErrorsToWarningsLogger.cs
@@ -1,0 +1,62 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Sbom;
+
+using System;
+using Serilog;
+using Serilog.Events;
+
+/// <summary>
+/// A class to remap Errors logged from ComponentDetection assemblies to Warnings.
+/// </summary>
+public class RemapComponentDetectionErrorsToWarningsLogger : ILogger
+{
+    private readonly ILogger logger;
+    private readonly Func<string?> stackTraceProvider;
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="RemapComponentDetectionErrorsToWarningsLogger"/> class.
+    /// Production constructor.
+    /// </summary>
+    public RemapComponentDetectionErrorsToWarningsLogger(ILogger logger)
+        : this(logger, () => Environment.StackTrace)
+    {
+    }
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="RemapComponentDetectionErrorsToWarningsLogger"/> class.
+    /// Testable constructor.
+    /// </summary>
+    /// <exception cref="ArgumentNullException"></exception>
+    public RemapComponentDetectionErrorsToWarningsLogger(ILogger logger, Func<string?> stackTraceProvider)
+    {
+        this.logger = logger ?? throw new ArgumentNullException(nameof(logger));
+        this.stackTraceProvider = stackTraceProvider ?? throw new ArgumentNullException(nameof(stackTraceProvider));
+    }
+
+    /// <inheritdoc />
+    /// <remarks>If ComponentDetection is on the stack, remap and log Errors to Warnings.</remarks>
+    public void Write(LogEvent logEvent)
+    {
+        // For performance reasons, bypass StackTrace work for non-errors.
+        if (logEvent.Level == LogEventLevel.Error)
+        {
+            var stackTrace = stackTraceProvider();
+            if (stackTrace is not null && stackTrace.Contains("Microsoft.ComponentDetection."))
+            {
+                var warningLogEvent = new LogEvent(
+                    logEvent.Timestamp,
+                    LogEventLevel.Warning,
+                    logEvent.Exception,
+                    logEvent.MessageTemplate,
+                    logEvent.Properties.Select(kvp => new LogEventProperty(kvp.Key, kvp.Value)));
+
+                logger.Write(warningLogEvent);
+                return;
+            }
+        }
+
+        logger.Write(logEvent);
+    }
+}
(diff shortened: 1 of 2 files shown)


src/Microsoft.Sbom.Extensions.DependencyInjection/RemapComponentDetectionErrorsToWarningsLogger.cs
```diff
@@ -0,0 +1,62 @@
...
+namespace Microsoft.Sbom;
+
+using System;
+using Serilog;
+using Serilog.Events;
+
+/// <summary>
+/// A class to remap Errors logged from ComponentDetection assemblies to Warnings.
+/// </summary>
+public class RemapComponentDetectionErrorsToWarningsLogger : ILogger
```
sfoslund, line 13: Much of this codebase doesn't have any unit tests, but IMO we should be trying to fix that as we add new classes as much as possible/is reasonable. Would you be able to add unit tests for this class?

src/Microsoft.Sbom.Extensions.DependencyInjection/RemapComponentDetectionErrorsToWarningsLogger.cs
```diff
@@ -0,0 +1,62 @@
...
+        : this(logger, () => Environment.StackTrace)
+    {
+    }
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="RemapComponentDetectionErrorsToWarningsLogger"/> class.
+    /// Testable constructor.
+    /// </summary>
+    /// <exception cref="ArgumentNullException"></exception>
+    public RemapComponentDetectionErrorsToWarningsLogger(ILogger logger, Func<string?> stackTraceProvider)
```
sfoslund, line 32: nit: can we make the testable constructor internal? This also doesn't appear to be used right now, can we either add tests that exercise it or remove it?

src/Microsoft.Sbom.Extensions.DependencyInjection/RemapComponentDetectionErrorsToWarningsLogger.cs
DaveTryon: Unit tests have been added

src/Microsoft.Sbom.Extensions.DependencyInjection/RemapComponentDetectionErrorsToWarningsLogger.cs
DaveTryon: Unit tests have been added and the test constructor has been made internal

test/Microsoft.Sbom.Extensions.DependencyInjection.Tests/RemapComponentDetectionErrorsToWarningsLoggerTests.cs
```diff
@@ -0,0 +1,149 @@
...
+        var nonErrorLevels = Enum.GetValues(typeof(LogEventLevel)).Cast<LogEventLevel>().Where((x) => x != LogEventLevel.Error);
+        foreach (var level in nonErrorLevels)
+        {
+            var logEvent = GetLogEvent(level);
+            loggerMock.Setup(x => x.Write(logEvent)).Verifiable();
+            var logger = new RemapComponentDetectionErrorsToWarningsLogger(loggerMock.Object, stackTraceProvider);
+
+            logger.Write(logEvent);
+
+            loggerMock.VerifyAll();
```
sfoslund, line 73: nit: our usual pattern is to have a test clean up method that verifies all of the mocks, so we don't have to remember to call verifyAll in every test separately. 
I also don't think the loggerMock.Reset is necessary, as the loggerMock is recreated before every test is run, right?

test/Microsoft.Sbom.Extensions.DependencyInjection.Tests/RemapComponentDetectionErrorsToWarningsLoggerTests.cs
```diff
@@ -0,0 +1,149 @@
...
+
+    [TestMethod]
+    public void Write_LogEventLevelIsNotError_DoesNotCallStackTraceProvider_LogsEventAsSpecified()
+    {
+        var nonErrorLevels = Enum.GetValues(typeof(LogEventLevel)).Cast<LogEventLevel>().Where((x) => x != LogEventLevel.Error);
+        foreach (var level in nonErrorLevels)
+        {
+            var logEvent = GetLogEvent(level);
+            loggerMock.Setup(x => x.Write(logEvent)).Verifiable();
+            var logger = new RemapComponentDetectionErrorsToWarningsLogger(loggerMock.Object, stackTraceProvider);
```
sfoslund, line 69: nit: we also usually initialize the test subject in the test init method so it doesn't have to be duplicated across all the tests

DaveTryon: Updated

test/Microsoft.Sbom.Extensions.DependencyInjection.Tests/RemapComponentDetectionErrorsToWarningsLoggerTests.cs
DaveTryon: Updated. In this case, since `loggerMock` is being recycled in a loop, the `VerifyAll` and `Reset` calls are needed. I suppose we could skip them on the last iteration, but that would just add code clutter without changing what is being tested. I added a comment about why the lines are present

sfoslund (approved):

PR #575, opened 2024-05-17
Add redact verb to CLI

sfoslund:
Adding the new `redact` verb to the sbom tool CLI. For now this verb is excluded from the tool help since the actual redaction logic is not yet present. With these changes running redact simply prints the arguments provided for the command, but redaction logic will be added in a future PR. 

Once we turn this feature on it will show up in the help text as follows: 
```  
  Redact -options - Redact sensitive file information from given SBOM(s).

    Option            Description
    SbomPath (-sp)    The file path of the SBOM to redact.
    SbomDir (-sd)     The directory containing the sbom(s) to redact.
    OutputPath (-o)   Gets or sets the directory where the redacted SBOM file(s) will be generated.
    Verbosity (-V)    Display this amount of detail in the logging output.
                      Verbose
                      Debug
                      Information
                      Warning
                      Error
                      Fatal

  Version  - Displays the version of the tool being used. Can be used as '--version'
```

This PR also updates the style linter to be less strict about some comment style rules.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -502,6 +502,9 @@ dotnet_diagnostic.SA1005.severity = none
 # SA1515: Single-line comment should be preceded by blank line
 dotnet_diagnostic.SA1515.severity = none
 
+# SA1512: Single-line comments should not be followed by blank line
+dotnet_diagnostic.SA1512.severity = none
+
 # https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1101.md
 # SA1101: Prefix local calls with this
 dotnet_diagnostic.SA1101.severity = suggestion
--- a/.editorconfig
+++ b/.editorconfig
@@ -496,6 +496,12 @@ dotnet_diagnostic.CA1863.severity = none
 # https://github.com/DotNetAnalyzers/StyleCopAnalyzers
 ##########################################
 
+# SA1005: Single line comments should begin with single space
+dotnet_diagnostic.SA1005.severity = none
+
+# SA1515: Single-line comment should be preceded by blank line
+dotnet_diagnostic.SA1515.severity = none
+
 # https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1101.md
 # SA1101: Prefix local calls with this
 dotnet_diagnostic.SA1101.severity = suggestion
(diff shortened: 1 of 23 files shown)


src/Microsoft.Sbom.DotNetTool/RedactService.cs
```diff
@@ -0,0 +1,56 @@
...
+    private readonly IRecorder recorder;
+
+    private readonly IHostApplicationLifetime hostApplicationLifetime;
+
+    public RedactService(
+        IWorkflow<SbomRedactionWorkflow> redactionWorkflow,
+        IRecorder recorder,
+        IHostApplicationLifetime hostApplicationLifetime)
+    {
+        this.redactionWorkflow = redactionWorkflow;
```
DaveTryon, line 29: Should these be validated against null values?

src/Microsoft.Sbom.Tool/RedactService.cs
```diff
@@ -0,0 +1,52 @@
...
+    private readonly IHostApplicationLifetime hostApplicationLifetime;
+
+    public RedactService(
+        IWorkflow<SbomRedactionWorkflow> redactionWorkflow,
+        IRecorder recorder,
+        IHostApplicationLifetime hostApplicationLifetime)
+    {
+        this.redactionWorkflow = redactionWorkflow;
+        this.recorder = recorder;
+        this.hostApplicationLifetime = hostApplicationLifetime;
```
DaveTryon, line 27: Should these arguments be validated against null?

.editorconfig
```diff
@@ -496,6 +496,15 @@ dotnet_diagnostic.CA1863.severity = none
...
 
+# SA1005: Single line comments should begin with single space
+dotnet_diagnostic.SA1005.severity = none
+
+# SA1515: Single-line comment should be preceded by blank line
+dotnet_diagnostic.SA1515.severity = none
+
+# SA1512: Single-line comments should not be followed by blank line
+dotnet_diagnostic.SA1512.severity = none
+
```
DaveTryon, line 507: Are these related to the changes? I don't notice any changes that stand out as being obvious matches

DaveTryon (approved):
Added a couple of questions, but seems OK to me

.editorconfig
sfoslund: No, this change just disables some very strict linting rules surrounding comments that were impeding my and Alison's dev loops.

alisonlomaka: IMO if we want strict linting rules like these, we should be automatically applying them at some point, instead of either interrupting flow or breaking compile and forcing engs to fix them one at a time. So painful.

alisonlomaka (approved):
LGTM!

PR #577, opened 2024-05-20
Add validate-format verb with placeholder for future validation

alisonlomaka:
Wire up a validate-format verb that runs the format validation logic for a specified file. This is a hidden verb that I imagine being used during development, and in future by DRI to explore an SBOM or troubleshoot a validation failure. This change is just the boilerplate required to add and call the new verb; actual validation logic will be in the next PR.

### DIFF
--- a/src/Microsoft.Sbom.Api/Config/ConfigSanitizer.cs
+++ b/src/Microsoft.Sbom.Api/Config/ConfigSanitizer.cs
@@ -147,17 +147,6 @@ public class ConfigSanitizer
                 throw new ValidationArgException($"Please provide a value for the BuildDropPath (-b) parameter to generate the SBOM.");
             }
         }
-        else if (configuration.ManifestToolAction == ManifestToolActions.ValidateFormat)
-        {
-            if (configuration.SbomPath.Value != null)
-            {
-                return;
-            }
-            else
-            {
-                throw new ValidationArgException($"Please provide a value for the SbomPath (-sp) parameter to validate the SBOM.");
-            }
-        }
         else
         {
             throw new ValidationArgException($"Please provide a value for the BuildDropPath (-b) parameter.");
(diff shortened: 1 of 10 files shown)


src/Microsoft.Sbom.Api/Config/ConfigSanitizer.cs
```diff
@@ -132,6 +147,17 @@ private void ValidateBuildDropPathConfiguration(IConfiguration configuration)
...
         }
+        else if (configuration.ManifestToolAction == ManifestToolActions.ValidateFormat)
+        {
+            if (configuration.SbomPath.Value != null)
+            {
+                return;
+            }
+            else
+            {
+                throw new ValidationArgException($"Please provide a value for the SbomPath (-sp) parameter to validate the SBOM.");
```
sfoslund, line 158: nit: it seems like we have this check/ throw this error in 2 places, are they both necessary?

src/Microsoft.Sbom.Common/Config/ManifestToolActions.cs
```diff
@@ -12,5 +12,6 @@ public enum ManifestToolActions
     Validate = 1,
     Generate = 2,
     Redact = 4,
-    All = Validate | Generate | Redact
+    ValidateFormat = 8,
```
sfoslund, line 15: Why 8 here?

sfoslund (approved):

src/Microsoft.Sbom.Common/Config/ManifestToolActions.cs
alisonlomaka: It looks like the enum was constructed as a bitmask, so our next binary flag would be 8.

This made me curious to go look and see why we even conceive of this enum as a bitmask - since in practice you could not OR together different actions to run them. Turns out we use it as an applicability filter on custom attributes that can be applied to the different command line flags.

src/Microsoft.Sbom.Api/Config/ConfigSanitizer.cs
alisonlomaka: Lol, nope, I pulled the logic out to a more sensibly named function and then didn't go back and delete it. Good catch!

PR #580, opened 2024-05-21
Add Validate Format functionality

alisonlomaka:
Implement a format validation service in support of redacting the Files section from an SPDX 2.x format SBOM.

This is a naive implementation that does ***not*** leverage the JSON object streaming mechanism that is used for drop validation. The reason for this is that it is fairly complex to amend that logic to customize the validation behavior. Since we are operating on a very tight timeline, the preliminary redaction implementation will rely on deserializing the entire JSON file at once, on the assumption that will cover a large proportion of our in-the-wild use cases. A follow-on investigation will be conducted to determine the practical limits of this technique and necessity for extending to support larger SBOMs.

The criteria for a "valid" SBOM include:
- well-formed JSON document
- includes required SPDX elements
- SPDX v2.x
- Packages and Relationships sections present

Validating SBOMs for SPDX elements covering the NTIA definition of an SBOM is out of scope for this change. Such validation may not be practical when processing 3P SBOMs since an SBOM creator could choose to use different attributes to satisfy the NTIA requirements.

In context of redaction, we will skip deserializing the files section (since removing it is the purpose of redaction anyway - this is also expected to reduce the size of an SBOM by ~50%). We will also require Packages and Relationships, since we need to operate on those in order to successfully redact. These ignore/require expectations are set at runtime but currently hardcoded; making them configurable is out of scope for this feature but definitely feasible.

The JSON validation and SPDX required elements are explicitly and implicitly enforced by the JSON model definition and deserialization. This also implicitly enforces the SPDX version since v3 is so different in format that the deserialization would fail before we ever got to a version check. Nonetheless, we make our expectations explicit by also parsing and verifying the spdxVersion value.

Expanding support for all SPDX 2.x (previously only SPDX 2.2.1) was in scope for this feature, and was accomplished by eliminating the use of enums for deserialization (an earlier PR) and adding to the JSON model all properties found in the SPDX 2.3 exemplar document (https://github.com/spdx/spdx-spec/blob/development/v2.3.1/examples/SPDXJSONExample-v2.3.spdx.json). SPDX 2.3 is backwards compatible with earlier 2.x versions. Note that this expanded SPDX version support ***only*** applies to the Validate Format and Redact functionality, ***not*** to Generate or Drop Validation. Adding all documented properties to the model also ensures that we do not lose data when deserializing/re-serializing after redaction.

### DIFF
--- a/src/Microsoft.Sbom.DotNetTool/FormatValidationService.cs
+++ b/src/Microsoft.Sbom.DotNetTool/FormatValidationService.cs
@@ -2,10 +2,13 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Collections.Generic;
+using System.IO;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Hosting;
 using Microsoft.Sbom.Api;
+using Microsoft.Sbom.Api.FormatValidator;
 using Microsoft.Sbom.Api.Output.Telemetry;
 using Microsoft.Sbom.Common.Config;
 
@@ -26,11 +29,26 @@ public class FormatValidationService : IHostedService
         this.hostApplicationLifetime = hostApplicationLifetime;
     }
 
+    private void PrintLines(List<string> lines)
+    {
+        foreach (var line in lines)
+        {
+            Console.WriteLine(line);
+        }
+    }
+
     public async Task StartAsync(CancellationToken cancellationToken)
     {
         try
         {
-            Console.WriteLine($"Format validation service called for {config.SbomPath.Value}");
+            using (var sbomStream = new StreamReader(config.SbomPath.Value))
+            {
+                var validatedSbom = new ValidatedSBOM(sbomStream.BaseStream);
+                var details = await validatedSbom.GetValidationResults();
+                var sbom = await validatedSbom.GetRawSPDXDocument();
+
+                PrintLines(validatedSbom.MultilineSummary());
+            }
 
             await recorder.FinalizeAndLogTelemetryAsync();
             Environment.ExitCode = true ? (int)ExitCode.Success : (int)ExitCode.ValidationError;
--- a/src/Microsoft.Sbom.DotNetTool/FormatValidationService.cs
+++ b/src/Microsoft.Sbom.DotNetTool/FormatValidationService.cs
@@ -44,10 +44,7 @@ public class FormatValidationService : IHostedService
             using (var sbomStream = new StreamReader(config.SbomPath.Value))
             {
                 var validatedSbom = new ValidatedSBOM(sbomStream.BaseStream);
-                var details = await validatedSbom.GetValidationResults();
-                var sbom = await validatedSbom.GetRawSPDXDocument();
-
-                PrintLines(validatedSbom.MultilineSummary());
+                PrintLines(await validatedSbom.MultilineSummary());
             }
 
             await recorder.FinalizeAndLogTelemetryAsync();
--- a/src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Utils/SPDXVersionParser.cs
+++ b/src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Utils/SPDXVersionParser.cs
@@ -1,0 +1,35 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Sbom.Utils;
+
+using System;
+
+public static class SPDXVersionParser
+{
+    // SPDX versions are of the form "SPDX-m.n". We only care about the major version.
+    public static bool VersionMatchesRequiredVersion(string spdxVersionString, int requiredMajorVersion)
+    {
+        if (string.IsNullOrEmpty(spdxVersionString))
+        {
+            return false;
+        }
+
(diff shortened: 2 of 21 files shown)


src/Microsoft.Sbom.DotNetTool/FormatValidationService.cs
```diff
@@ -26,11 +29,26 @@ public class FormatValidationService : IHostedService
...
     public async Task StartAsync(CancellationToken cancellationToken)
     {
         try
         {
-            Console.WriteLine($"Format validation service called for {config.SbomPath.Value}");
+            using (var sbomStream = new StreamReader(config.SbomPath.Value))
+            {
+                var validatedSbom = new ValidatedSBOM(sbomStream.BaseStream);
+                var details = await validatedSbom.GetValidationResults();
+                var sbom = await validatedSbom.GetRawSPDXDocument();
```
sfoslund, line 48: nit: we never appear to use the raw SBOM, why is this line necessary?

src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Utils/SPDXVersionParser.cs
```diff
@@ -0,0 +1,35 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Sbom.Utils;
+
+using System;
+
+public static class SPDXVersionParser
+{
+    // SPDX versions are of the form "SPDX-m.n". We only care about the major version.
+    public static bool VersionMatchesRequiredVersion(string spdxVersionString, int requiredMajorVersion)
```
sfoslund, line 11: This seems like a good candidate for some unit tests

test/Microsoft.Sbom.Api.Tests/FormatValidator/FormatValidatorTests.cs
```diff
@@ -0,0 +1,185 @@
...
+            var rawspdx = await sbom.GetRawSPDXDocument();
+            var details = await sbom.GetValidationResults();
+
+            Assert.AreEqual(FormatValidationStatus.Valid, details.Status);
+            Assert.IsTrue(details.Errors.Count == 0);
+            Assert.IsNotNull(rawspdx);
+            Assert.AreEqual("SPDX-2.2", rawspdx.Version);
+            Assert.AreEqual("CC0-1.0", rawspdx.DataLicense);
+            Assert.AreEqual("sbom-tool 1.0.0", rawspdx.Name);
+            Assert.AreEqual("https://microsoft.com/sbom-tool/alisontest/sbom-tool/1.0.0/cuK7iCCPVEuSmgBfeFPc-g", rawspdx.DocumentNamespace);
```
sfoslund, line 33: nit: can we search and replace `alisontest` in these strings with something more generic?

test/Microsoft.Sbom.Api.Tests/FormatValidator/SpdxExemplars.cs
```diff
@@ -0,0 +1,306 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+internal readonly struct SpdxExemplars
+{
+    public const string JsonSpdx23Exemplar = /*lang=json,strict*/ @"{
```
sfoslund, line 14: nit: no strong feelings about this, but wondering if it makes more sense to save this as its own spdx file that we use as a test asset instead of keeping it as a string like this.

sfoslund (approved):

src/Microsoft.Sbom.DotNetTool/FormatValidationService.cs
alisonlomaka: It's not, nor the preceding. Good catch!

src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Utils/SPDXVersionParser.cs
alisonlomaka: Unit tests are in Microsoft.Sbom.Api.Tests.

test/Microsoft.Sbom.Api.Tests/FormatValidator/SpdxExemplars.cs
alisonlomaka: I considered it, but didn't want to have to fiddle with locating the file relative to the path of wherever the test harness executes. Or I guess embedding it as a resource in the test assembly would be another option. I don't feel strongly enough about it to spend the effort, but I will if you have strong feelings.

src/Microsoft.Sbom.DotNetTool/FormatValidationService.cs
alisonlomaka: Oh hey, that exposed a nice usage bug! The ValidatedSBOM is lazy-loaded but MultilineSummary was not loading it, so removing these 2 lines caused the validate-format command to stop outputting results to the CLI.

test/Microsoft.Sbom.Api.Tests/FormatValidator/SpdxExemplars.cs
sfoslund: Nope, that's reasonable, works for me!

sfoslund (approved):

sfoslund (approved):

PR #581, opened 2024-05-21
Add redaction workflow logic

sfoslund:
This PR is a follow up to https://github.com/microsoft/sbom-tool/pull/575, which added the CLI plumbing for the `redact` verb. This PR populates the redact workflow with the redaction logic. The current redaction logic involves the following components:
- Remove the entire files section
- Remove any relationship entries which include a file reference
- Remove any `packages.hasFiles` entries 
- Remove any `packages.sourceInfo` entries (note that this property is not populated for SBOMs generated by this tool, but we have seen syft SBOMs in which this property contains file information)
- If the document namespace indicates that the SBOM was generated via this sbom-tool, generate a new unique guid section of the document namespace

### DIFF
--- a/src/Microsoft.Sbom.Api/FormatValidator/IValidatedSBOM.cs
+++ b/src/Microsoft.Sbom.Api/FormatValidator/IValidatedSBOM.cs
@@ -11,4 +11,6 @@ public interface IValidatedSBOM
     public Task<FormatValidationResults> GetValidationResults();
 
     public Task<FormatEnforcedSPDX2> GetRawSPDXDocument();
+
+    public void Dispose();
 }
--- a/src/Microsoft.Sbom.Api/FormatValidator/IValidatedSBOM.cs
+++ b/src/Microsoft.Sbom.Api/FormatValidator/IValidatedSBOM.cs
@@ -1,0 +1,14 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Sbom.Api.FormatValidator;
+
+using System.Threading.Tasks;
+using Microsoft.Sbom.Parsers.Spdx22SbomParser.Entities;
+
+public interface IValidatedSBOM
+{
+    public Task<FormatValidationResults> GetValidationResults();
+
+    public Task<FormatEnforcedSPDX2> GetRawSPDXDocument();
+}
--- a/src/Microsoft.Sbom.Api/FormatValidator/IValidatedSBOM.cs
+++ b/src/Microsoft.Sbom.Api/FormatValidator/IValidatedSBOM.cs
@@ -3,14 +3,13 @@
 
 namespace Microsoft.Sbom.Api.FormatValidator;
 
+using System;
 using System.Threading.Tasks;
 using Microsoft.Sbom.Parsers.Spdx22SbomParser.Entities;
 
-public interface IValidatedSBOM
+public interface IValidatedSBOM: IDisposable
 {
     public Task<FormatValidationResults> GetValidationResults();
 
     public Task<FormatEnforcedSPDX2> GetRawSPDXDocument();
-
-    public void Dispose();
 }
--- a/src/Microsoft.Sbom.Api/FormatValidator/ValidatedSBOMFactory.cs
+++ b/src/Microsoft.Sbom.Api/FormatValidator/ValidatedSBOMFactory.cs
@@ -1,0 +1,16 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Sbom.Api.FormatValidator;
+
+using System.IO;
+
+public class ValidatedSBOMFactory
+{
+    public virtual IValidatedSBOM CreateValidatedSBOM(string sbomFilePath)
+    {
+        var sbomStream = new StreamReader(sbomFilePath);
+        var validatedSbom = new ValidatedSBOM(sbomStream.BaseStream);
+        return validatedSbom;
+    }
+}
--- a/src/Microsoft.Sbom.Api/Workflows/Helpers/SbomRedactor.cs
+++ b/src/Microsoft.Sbom.Api/Workflows/Helpers/SbomRedactor.cs
@@ -1,0 +1,88 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.Sbom.Api.FormatValidator;
+using Microsoft.Sbom.Common.Utils;
+using Microsoft.Sbom.Parsers.Spdx22SbomParser.Entities;
+
+namespace Microsoft.Sbom.Api.Workflows.Helpers;
+
+/// <summary>
+/// SBOM redactor that removes file information from SBOMs
+/// </summary>
+public class SbomRedactor
+{
+    private const string SpdxFileRelationshipPrefix = "SPDXRef-File-";
+
+    public virtual async Task<FormatEnforcedSPDX2> RedactSBOMAsync(IValidatedSBOM sbom)
+    {
+        var spdx = await sbom.GetRawSPDXDocument();
+
+        if (spdx.Files != null)
+        {
+            spdx.Files = null;
+        }
+
+        RemovePackageFileRefs(spdx);
+        RemoveRelationshipsWithFileRefs(spdx);
+        UpdateDocumentNamespace(spdx);
+
+        return spdx;
+    }
+
+    private void RemovePackageFileRefs(FormatEnforcedSPDX2 spdx)
+    {
+        if (spdx.Packages != null)
+        {
+            foreach (var package in spdx.Packages)
+            {
+                if (package.HasFiles != null)
+                {
+                    package.HasFiles = null;
+                }
+
+                if (package.SourceInfo != null)
+                {
+                    package.SourceInfo = null;
+                }
+            }
+        }
+    }
+
+    private void RemoveRelationshipsWithFileRefs(FormatEnforcedSPDX2 spdx)
+    {
+        if (spdx.Relationships != null)
+        {
+            var relationshipsToRemove = new List<SPDXRelationship>();
+            foreach (var relationship in spdx.Relationships)
+            {
+                if (relationship.SourceElementId.Contains(SpdxFileRelationshipPrefix) || relationship.TargetElementId.Contains(SpdxFileRelationshipPrefix))
+                {
+                    relationshipsToRemove.Add(relationship);
+                }
+            }
+
+            if (relationshipsToRemove.Any())
+            {
+                spdx.Relationships = spdx.Relationships.Except(relationshipsToRemove);
+            }
+        }
+    }
+
+    private void UpdateDocumentNamespace(FormatEnforcedSPDX2 spdx)
+    {
+        if (string.IsNullOrWhiteSpace(spdx.DocumentNamespace) || !spdx.DocumentNamespace.Contains("microsoft"))
+        {
+            return;
+        }
+
+        var existingNamespaceComponents = spdx.DocumentNamespace.Split('/');
+        var uniqueComponent = IdentifierUtils.GetShortGuid(Guid.NewGuid());
+        existingNamespaceComponents[^1] = uniqueComponent;
+        spdx.DocumentNamespace = string.Join("/", existingNamespaceComponents);
+    }
+}
--- a/src/Microsoft.Sbom.Api/Workflows/Helpers/SbomRedactor.cs
+++ b/src/Microsoft.Sbom.Api/Workflows/Helpers/SbomRedactor.cs
@@ -15,7 +15,7 @@ namespace Microsoft.Sbom.Api.Workflows.Helpers;
 /// <summary>
 /// SBOM redactor that removes file information from SBOMs
 /// </summary>
-public class SbomRedactor
+public class SbomRedactor: ISbomRedactor
 {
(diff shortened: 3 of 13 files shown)


src/Microsoft.Sbom.Api/FormatValidator/ValidatedSBOMFactory.cs
```diff
@@ -0,0 +1,16 @@
...
+
+namespace Microsoft.Sbom.Api.FormatValidator;
+
+using System.IO;
+
+public class ValidatedSBOMFactory
+{
+    public virtual IValidatedSBOM CreateValidatedSBOM(string sbomFilePath)
+    {
+        var sbomStream = new StreamReader(sbomFilePath);
```
DaveTryon, line 12: Does this need a `using` statement to properly dispose of the Stream?

src/Microsoft.Sbom.Api/Workflows/Helpers/SbomRedactor.cs
```diff
@@ -0,0 +1,88 @@
...
+            if (relationshipsToRemove.Any())
+            {
+                spdx.Relationships = spdx.Relationships.Except(relationshipsToRemove);
+            }
+        }
+    }
+
+    private void UpdateDocumentNamespace(FormatEnforcedSPDX2 spdx)
+    {
+        if (string.IsNullOrWhiteSpace(spdx.DocumentNamespace) || !spdx.DocumentNamespace.Contains("microsoft"))
```
DaveTryon, line 78: The guard, as written, seems to intend to prevent changing the namespaces for non-Microsoft files. That said, what if a non-Microsoft file had a namespace like "wow-microsoft-would-love-this" or similar? Would something like `StartsWith("microsoft.")` be a better guard?

src/Microsoft.Sbom.Api/FormatValidator/ValidatedSBOMFactory.cs
alisonlomaka: Hm, the lifespan of the stream is fuzzy here since both the caller and the ValidatedSBOM touch it. The ValidatedSBOM needs that stream to live at least long enough to deserialize it - which since it happens lazily, could be sometime after the ValidatedSBOM is created. So a using(stream) here could cause the stream to be disposed before the ValidatedSBOM has actually deserialized it.

Maybe the ValidatedSBOM should own the stream and responsibility for disposing it when it is disposed itself.

src/Microsoft.Sbom.Api/Workflows/SBOMRedactionWorkflow.cs
```diff
@@ -17,17 +24,133 @@ public class SbomRedactionWorkflow : IWorkflow<SbomRedactionWorkflow>
...
+            catch (Exception ex)
+            {
+                throw new Exception($"Failed to redact {sbomPath}: {ex.Message}", ex);
+            }
+        }
+
+        return true;
+    }
+
+    private async Task<IDictionary<string, IValidatedSBOM>> GetValidSbomsAsync()
```
DaveTryon, line 75: If the caller doesn't need to alter the contents, maybe
```suggestion
    private async Task<IReadOnlyDictionary<string, IValidatedSBOM>> GetValidSbomsAsync()
```

src/Microsoft.Sbom.Api/Workflows/Helpers/SbomRedactor.cs
```diff
@@ -0,0 +1,88 @@
...
+    public virtual async Task<FormatEnforcedSPDX2> RedactSBOMAsync(IValidatedSBOM sbom)
+    {
+        var spdx = await sbom.GetRawSPDXDocument();
+
+        if (spdx.Files != null)
+        {
+            spdx.Files = null;
+        }
+
+        RemovePackageFileRefs(spdx);
```
alisonlomaka, line 30: I wonder if we might want some tracing of what we've done in here, either for telemetry or troubleshooting.

src/Microsoft.Sbom.Api/Workflows/SBOMRedactionWorkflow.cs
```diff
@@ -17,17 +24,133 @@ public class SbomRedactionWorkflow : IWorkflow<SbomRedactionWorkflow>
...
+
+        var outputDir = configuration.OutputPath.Value;
+        if (fileSystemUtils.GetFullPath(outputDir).Equals(fileSystemUtils.GetFullPath(inputDir)))
+        {
+            throw new ArgumentException("Output path cannot be the same as input SBOM directory.");
+        }
+
+        if (!fileSystemUtils.DirectoryExists(outputDir))
+        {
+            fileSystemUtils.CreateDirectory(outputDir);
```
DaveTryon, line 142: Assuming `CreateDirectory` fails (bad drive, no permissions, no disk space, etc.), is the default Exception handling enough or should we provide a nicer error message?

src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Entities/CreationInfo.cs
```diff
@@ -28,5 +29,6 @@ public class CreationInfo
     public IEnumerable<string> Creators { get; set; }
 
     [JsonPropertyName("licenseListVersion")]
+    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
```
DaveTryon, line 32: Maybe flip the order of the attributes, for consistency with the others that you're adding?
 ```suggestion
    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
    [JsonPropertyName("licenseListVersion")]
```

File: src/Microsoft.Sbom.Api/Workflows/Helpers/SbomRedactor.cs (hunk 2 above)
Comment (sfoslund) [reply src/Microsoft.Sbom.Api/Workflows/Helpers/SbomRedactor.cs:78]: Yeah, I was tossing around ideas of how to detect when our tool produced a SBOM since it seems like there are differing namespaces (in some examples I'm seeing `"https://sbom.microsoft/sbom-tool/....` and in others I'm seeing `https://microsoft.com/sbom-tool/...`). If we want to guard more carefully, it might make sense to check for `sbom-tool` as well, what do you think?

File: src/Microsoft.Sbom.Api/Workflows/SBOMRedactionWorkflow.cs (hunk 5 above)
Comment (sfoslund) [reply src/Microsoft.Sbom.Api/Workflows/SBOMRedactionWorkflow.cs:142]: IMO default error handling is enough- any exceptions thrown here are caught higher up in the code and wrapped in a cleaner error message

File: src/Microsoft.Sbom.Api/Workflows/SBOMRedactionWorkflow.cs [hunk 7]
```diff
@@ -17,17 +24,133 @@ public class SbomRedactionWorkflow : IWorkflow<SbomRedactionWorkflow>
...
+        IConfiguration configuration,
+        IFileSystemUtils fileSystemUtils,
+        ValidatedSBOMFactory validatedSBOMFactory = null,
+        SbomRedactor sbomRedactor = null)
     {
         this.log = log ?? throw new ArgumentNullException(nameof(log));
         this.configuration = configuration ?? throw new ArgumentNullException(nameof(configuration));
+        this.fileSystemUtils = fileSystemUtils ?? throw new ArgumentNullException(nameof(fileSystemUtils));
+        this.validatedSBOMFactory = validatedSBOMFactory ?? new ValidatedSBOMFactory();
+        this.sbomRedactor = sbomRedactor ?? new SbomRedactor();
```
DaveTryon, line 44: This sort of pattern to optionally override parameters feels a bit out of place in an app that uses Dependency Injection as a key part of its structure. If we don't want to use DI, would it make sense to split this constructor into 2 constructors, one of which is used for testing?  Something like this:

```suggestion
    // production constructor
    public SbomRedactionWorkflow(
        ILogger log,
        IConfiguration configuration,
        IFileSystemUtils fileSystemUtils)
        : this (log, configuration, fileSystemUtils, new ValidatedSBOMFactory(), new SbomRedactor)
    {
    }
    
    // unit test constructor
    public SbomRedactionWorkflow(
        ILogger log,
        IConfiguration configuration,
        IFileSystemUtils fileSystemUtils,
        ValidatedSBOMFactory validatedSBOMFactory,
        SbomRedactor sbomRedactor)
    {
        this.log = log ?? throw new ArgumentNullException(nameof(log));
        this.configuration = configuration ?? throw new ArgumentNullException(nameof(configuration));
        this.fileSystemUtils = fileSystemUtils ?? throw new ArgumentNullException(nameof(fileSystemUtils));
        this.validatedSBOMFactory = validatedSBOMFactory ?? throw new ArgumentNullException(nameof(validatedSBOMFactory));
        this.sbomRedactor = sbomRedactor ?? throw new ArgumentNullException(nameof(sbomRedactor));
    }
```

src/Microsoft.Sbom.Api/Workflows/Helpers/SbomRedactor.cs
alisonlomaka: Could we use the Creators Tool value? I think we write this to a specific string when we create an SBOM.
"creators": [
      "Organization: AlisonTest",
      "Tool: Microsoft.SBOMTool-2.2.5"
    ]

sfoslund: Good call, works for me!

src/Microsoft.Sbom.Api/FormatValidator/ValidatedSBOMFactory.cs
DaveTryon: Relying on the caller to provide cleanup of an `IDisposable` resource (in this case, the `Stream`) is less than ideal--in fact, [CA1001](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1001) is designed to avoid just this scenario.

I think any "proper" solution would involve deriving `IValidatedSBOM` from `IDisposable` and implementing the associated `Dispose` method. That may or may not be worth the small memory leak that is represented here, given the short program lifetime and the current time constraints.

If `ValidatedSBOM` implements `IDisposable`, then it might also make sense to revisit the constructor parameter. Instead of a `Stream`, it could represent a file path or a string that represents the file contents, and then the `Stream` object is fully contained inside `ValidatedSBOM`.

sfoslund:
@DaveTryon @alisonlomaka thanks for the feedback, I believe I have made all the requested changes, let me know if I missed anything!

src/Microsoft.Sbom.Api/Workflows/SBOMRedactionWorkflow.cs
```diff
@@ -50,44 +50,61 @@ public virtual async Task<bool> RunAsync()
...
+
+        try
         {
-            try
+            foreach (var (sbomPath, validatedSbom) in sboms)
             {
-                var outputPath = GetOutputPath(sbomPath);
-                var redactedSpdx = await this.sbomRedactor.RedactSBOMAsync(validatedSbom);
-                using (var outStream = fileSystemUtils.OpenWrite(outputPath))
+                try
```
DaveTryon: Nested `try` blocks in the same method are something of a code smell. Would it make sense to refactor lines 58-72 into a private method, then just call that method from inside the loop?

src/Microsoft.Sbom.Api/Workflows/SBOMRedactionWorkflow.cs
```diff
@@ -50,44 +50,61 @@ public virtual async Task<bool> RunAsync()
         var sboms = await GetValidSbomsAsync();
```
DaveTryon: Should the call to `GetValidSbomsAsync` be inside the `try` block?

src/Microsoft.Sbom.Api/FormatValidator/ValidatedSBOMFactory.cs
alisonlomaka: Yeah, I agree that the caller should not need to know how long to keep the stream around and when it's ok to dispose it. The lifetime must be controlled by the only thing that has a stake in how long it lives, and that's the ValidatedSBOM. But, it's maybe a little unexpected that something else creates it, and hands off ownership.

I think we _don't_ want to be passing a string around because of the potential for very big SBOMs in memory. And I moved away from a file path in order to facilitate unit testing.

Maybe ValidatedSBOM should expose a Load(stream) method. Then the caller explicitly calls load, inside a using(stream) block. ValidatedSBOM no longer needs to store the stream at all, and the time when the load occurs is completely deterministic, controlled by the caller, and separate from object construction.

src/Microsoft.Sbom.Api/Workflows/SBOMRedactionWorkflow.cs
sfoslund: Yep, it looks like you might be looking at an older commit, I didn't like this either so I refactored in a subsequent commit.

src/Microsoft.Sbom.Api/FormatValidator/IValidatedSBOM.cs
```diff
@@ -11,4 +11,6 @@ public interface IValidatedSBOM
     public Task<FormatValidationResults> GetValidationResults();
 
     public Task<FormatEnforcedSPDX2> GetRawSPDXDocument();
+
+    public void Dispose();
```
DaveTryon, line 15: Suggestion: Rather than adding the `Dispose` method explicitly, derive `IValidatedSBOM` from `IDisposable`. That buys you the syntactic sugar of `using` blocks

src/Microsoft.Sbom.Api/Workflows/SBOMRedactionWorkflow.cs
sfoslund: Also appears to be out of date with the latest commits :)

src/Microsoft.Sbom.Api/FormatValidator/IValidatedSBOM.cs
sfoslund: Done!

DaveTryon (approved):

PR #600, opened 2024-06-25
Create template tool task

vpatakottu:
Create a Tool Task that invokes the SBOM CLI tool. This PR will only have an outline of the tool task without any implemented methods. It also has logic for conditionally compiling the GenerateSbomTask vs SbomCLIToolTask based on the target framework.

### DIFF
--- a/src/Microsoft.Sbom.Targets/GenerateSbom.cs
+++ b/src/Microsoft.Sbom.Targets/GenerateSbom.cs
@@ -15,85 +15,91 @@ using Microsoft.Build.Framework;
 public partial class GenerateSbom
 {
     /// <summary>
-    /// The path to the drop directory for which the SBOM will be generated.
+    /// Gets or sets the path to the drop directory for which the SBOM will be generated.
     /// </summary>
     [Required]
     public string BuildDropPath { get; set; }
 
     /// <summary>
-    /// Supplier of the package the SBOM represents.
+    /// Gets or sets the supplier of the package the SBOM represents.
     /// </summary>
     [Required]
     public string PackageSupplier { get; set; }
 
     /// <summary>
-    /// Name of the package the SBOM represents.
+    /// Gets or sets the name of the package the SBOM represents.
     /// </summary>
     [Required]
     public string PackageName { get; set; }
 
     /// <summary>
-    /// Version of the package the SBOM represents.
+    /// Gets or sets the version of the package the SBOM represents.
     /// </summary>
     [Required]
     public string PackageVersion { get; set; }
 
     /// <summary>
-    /// The base path of the SBOM namespace uri.
+    /// Gets or sets the base path of the SBOM namespace uri.
     /// </summary>
     [Required]
     public string NamespaceBaseUri { get; set; }
 
     /// <summary>
-    /// The path to the directory containing build components and package information.
+    /// Gets or sets the path to the directory containing build components and package information.
     /// For example, path to a .csproj or packages.config file.
     /// </summary>
     public string BuildComponentPath { get; set; }
 
     /// <summary>
-    /// A unique URI part that will be appended to NamespaceBaseUri.
+    /// Gets or sets a unique URI part that will be appended to NamespaceBaseUri.
     /// </summary>
     public string NamespaceUriUniquePart { get; set; }
 
     /// <summary>
-    /// The path to a file containing a list of external SBOMs that will be appended to the
+    /// Gets or sets the path to a file containing a list of external SBOMs that will be appended to the
     /// SBOM that is being generated.
     /// </summary>
     public string ExternalDocumentListFile { get; set; }
 
     /// <summary>
-    /// If true, it will fetch licensing information for detected packages.
+    /// Indicates whether licensing information will be fetched for detected packages.
     /// </summary>
     public bool FetchLicenseInformation { get; set; }
 
     /// <summary>
-    /// If true, it will parse licensing and supplier information from a packages metadata file.
+    /// Indicates whether to parse licensing and supplier information from a packages metadata file.
     /// </summary>
     public bool EnablePackageMetadataParsing { get; set; }
 
     /// <summary>
-    /// Determines how detailed the outputed logging will be.
+    /// Gets or sets the verbosity level for logging output.
     /// </summary>
     public string Verbosity { get; set; }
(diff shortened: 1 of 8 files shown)


src/Microsoft.Sbom.Targets/GenerateSbom.cs
```diff
@@ -0,0 +1,100 @@
...
+using Microsoft.Build.Framework;
+
+/// <summary>
+/// This partial class defines and sanitizes the arguments that will be passed
+/// into the SBOM API and CLI tool for generation.
+/// </summary>
+public partial class GenerateSbom
+{
+    /// <summary>
+    /// The path to the drop directory for which the SBOM will be generated.
```
nschwerzler, line 18: Would add Gets or sets to the summary doc

Gets or sets the path to the drop directory...

src/Microsoft.Sbom.Targets/GenerateSbom.cs
```diff
@@ -0,0 +1,100 @@
...
+    /// Version of the package the SBOM represents.
+    /// </summary>
+    [Required]
+    public string PackageVersion { get; set; }
+
+    /// <summary>
+    /// The base path of the SBOM namespace uri.
+    /// </summary>
+    [Required]
+    public string NamespaceBaseUri { get; set; }
```
nschwerzler, line 45: This should be a Uri to protect the input

src/Microsoft.Sbom.Targets/GenerateSbom.cs
```diff
@@ -0,0 +1,100 @@
...
+    public string Verbosity { get; set; }
+
+    /// <summary>
+    /// A list of the name and version of the manifest format being used.
+    /// </summary>
+    public string ManifestInfo { get; set; }
+
+    /// <summary>
+    /// If true, it will delete the previously generated SBOM manifest directory before
+    /// generating a new SBOM in ManifestDirPath.
```
nschwerzler, line 86: Suggestion:
```
/// <summary>
/// Indicates whether the previously generated SBOM manifest directory should be deleted 
/// before generating a new SBOM in the directory specified by ManifestDirPath. 
/// Defaults to true.
/// </summary>
```

src/Microsoft.Sbom.Targets/Microsoft.Sbom.Targets.csproj
```diff
@@ -39,6 +41,15 @@
     </ItemGroup>
   </Target>
 
+   <Target Name="AddSbomToolToPackage" AfterTargets="Build">
```
nschwerzler, line 44: We should figure out if there is a better AfterTargets for SBOM generation to only do this when they need an SBOM. Can do in a follow-up change once that is figured out.

src/Microsoft.Sbom.Targets/SbomCLIToolTask.cs
```diff
@@ -0,0 +1,53 @@
...
+/// </summary>
+public partial class GenerateSbom : ToolTask
+{
+    /// <summary>
+    /// Path to the SBOM CLI tool
+    /// </summary>
+    [Required]
+    public string SbomToolPath { get; set; }
+
+    protected override string ToolName => "Microsoft.Sbom.Tool";
```
nschwerzler, line 25: Comment

src/Microsoft.Sbom.Targets/SbomCLIToolTask.cs
```diff
@@ -0,0 +1,53 @@
...
+
+    /// <summary>
+    /// Get full path to SBOM CLI tool.
+    /// </summary>
+    /// <returns></returns>
+    protected override string GenerateFullPathToTool()
+    {
+        return $"{this.SbomToolPath}\\{this.ToolName}.exe";
```
nschwerzler, line 33: Use Path.Combine to guard against extra slashes

src/Microsoft.Sbom.Targets/SbomCLIToolTask.cs
```diff
@@ -0,0 +1,53 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Sbom.Targets;
+
+using System;
```
nschwerzler, line 6: Double check you need all these usings. Some look like they are extra.

src/Microsoft.Sbom.Targets/Microsoft.Sbom.Targets.csproj
gustavoaca1997: This target is for including the SBOM CLI Tool into the Task's Nuget Package. I tried making the target run only during the packing step without success :(

src/Microsoft.Sbom.Targets/GenerateSbom.cs
vpatakottu: Updated!

src/Microsoft.Sbom.Targets/GenerateSbom.cs
vpatakottu: The MSBuild Task seems to only allow strings and bool's as parameters when we are passing them from the XML <Target> command, and the SBOM API accepts this value as a string as well. I believe the SBOM API does some URI checking, but should we implement some extra validation for this input as well?

src/Microsoft.Sbom.Targets/GenerateSbom.cs
vpatakottu: Updated!

src/Microsoft.Sbom.Targets/SbomCLIToolTask.cs
vpatakottu: Moved this property to GenerateSbom.cs and updated the comment!

src/Microsoft.Sbom.Targets/SbomCLIToolTask.cs
vpatakottu: Updated!

gustavoaca1997 (approved):
Good job!

vpatakottu:
@microsoft-github-policy-service agree company="Microsoft"

nschwerzler (approved):

Issue #483, opened 2024-01-17
Add E2E test in PR pipeline


pownkel:
One of the [dependabot PRs](https://github.com/microsoft/sbom-tool/pull/342) passes the PR build, but causes the tool to fail. If we had an E2E test in the PR pipeline, this would help us catch these issues before merging other dependabot PRs

DaveTryon:
This has been fixed

PR #606, opened 2024-07-06
Add simple integration tests

DaveTryon:
As called out in #483, our PR pipelines include no E2E tests. This PR adds E2E tests to cover 5 simple cases (more can be added at a later time):
1. Runs on all platforms. App exists where it is expected
    - This exists mainly as a step toward cross-platform testing. It can be removed, if desired.
2. Runs only on Windows. App is run with no parameters. The following checks are made:
    - App provides a non-zero return code
    - Nothing is sent to `stderr`
    - Help text is displayed to `stdout`
3. Runs only on Windows. App is run to generate an SBOM. The following checks are made:
    - App provides a zero return code from generation
    - A non-empty SBOM file is created in the specified location
    - A non-empty SHA file is created in the specified location
4. Runs only on Windows. App is run to generate an SBOM, then validate it. This includes all checks from test <span>#</span>3, plus the following:
    - App provides a zero return code from validation
    - Output file is written to the specified location
    - Output file includes `"Result":"Success"` - it just looks for the snippet, rather than parsing the entire JSON file
5. Runs only on Windows. App is run to generate an SBOM, then redact it. This includes all checks from test <span>#</span>3, plus the following:
    - App provides a zero return code from redaction
    - Telemetry (based on verbose output) includes `Result=Success`
    - The redacted SBOM is created in the specified location and has a length > 0
    - The redacted SBOM file is smaller than the original SBOM file

On my build machine, these tests add about 30 seconds to the test execution time. If we wish, I'm happy (in a future PR) to add some sort of filtering to limit them to pipeline builds only.

Running tests 2-5 on non-Windows platforms fail, potentially because the tool doesn't have the execute bit set. I've tried some variations of using `chmod u+x` and the .NET 8 [File.SetUnixFileMode](https://learn.microsoft.com/en-us/dotnet/api/system.io.file.setunixfilemode?view=net-8.0) function without success, so I took the approach of "some improvement is better than none" and made the tests Windows-only.

### DIFF
--- a/test/Microsoft.Sbom.Tool.Tests/IntegrationTests.cs
+++ b/test/Microsoft.Sbom.Tool.Tests/IntegrationTests.cs
@@ -20,23 +20,23 @@ public class IntegrationTests
 
     public TestContext TestContext { get; set; }
 
-    private static string testResultsDirectory;
+    private static string testRunDirectory;
 
     [ClassInitialize]
     public static void Setup(TestContext context)
     {
-        testResultsDirectory = context.ResultsDirectory;
+        testRunDirectory = context.TestRunDirectory;
     }
 
     [ClassCleanup]
     public static void TearDown()
     {
         // Clean up test directories
-        if (testResultsDirectory is not null)
+        if (testRunDirectory is not null)
         {
-            if (Directory.Exists(testResultsDirectory))
+            if (Directory.Exists(testRunDirectory))
             {
-                Directory.Delete(testResultsDirectory, true);
+                Directory.Delete(testRunDirectory, true);
             }
         }
     }
@@ -90,7 +90,7 @@ public class IntegrationTests
         var testFolderPath = CreateTestFolder();
         GenerateManifestAndValidateSuccess(testFolderPath);
 
-        var outputFile = Path.Combine(TestContext.ResultsDirectory, TestContext.TestName, "validation.json");
+        var outputFile = Path.Combine(TestContext.TestRunDirectory, TestContext.TestName, "validation.json");
         var manifestRootFolderName = Path.Combine(testFolderPath, ManifestRootFolderName);
         var arguments = $"validate -m \"{manifestRootFolderName}\" -b . -o \"{outputFile}\" -mi spdx:2.2";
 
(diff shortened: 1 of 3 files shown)


test/Microsoft.Sbom.Tool.Tests/IntegrationTests.cs
```diff
@@ -0,0 +1,219 @@
...
+
+            process.Start();
+            process.BeginOutputReadLine();
+            process.BeginErrorReadLine();
+            process.WaitForExit();
+        }
+        catch (Exception ex)
+        {
+            // Handle any exceptions here
+            Console.WriteLine("Error: " + ex.Message);
```
sfoslund, line 201: I would expect the test to fail on an exception- why would we not want that to be the case?

test/Microsoft.Sbom.Tool.Tests/IntegrationTests.cs
```diff
@@ -0,0 +1,219 @@
...
+        var shaFilePath = Path.Combine(manifestFolderPath, "manifest.spdx.json.sha256");
+        Assert.IsTrue(File.Exists(jsonFilePath));
+        Assert.IsTrue(File.Exists(shaFilePath));
+        Assert.AreEqual(0, exitCode.Value);
+    }
+
+    private string CreateTestFolder()
+    {
+        var testFolderPath = Path.GetFullPath(Path.Combine(testContext.ResultsDirectory, testContext.TestName));
+        Directory.CreateDirectory(testFolderPath);
```
sfoslund, line 136: Can we add a test cleanup method to delete these test dirs once the tests finish?

test/Microsoft.Sbom.Tool.Tests/IntegrationTests.cs
DaveTryon: Good point. I wrote this via Copilot and wasn't thinking in terms of errors occurring here. I _think_ upstream tests would fail, but an `Assert.Fail` here will guarantee it.

sfoslund (approved):

PR #607, opened 2024-07-07
Implement SBOM CLI ToolTask

vpatakottu:
Added implementation for ToolTask to invoke the SBOM CLI tool. This involved the following changes:

- Implemented the methods required to invoke the SBOM CLI tool in SbomCLIToolTask
- Added a new SbomInputValidator partial class to share the argument validation/sanitization logic between the SbomCLIToolTask and GenerateSbomTask. 
- Added logic within Microsoft.Sbom.Targets.targets to copy the _manifest contents into the consumer's NuGet package for the SbomCLIToolTask scenario.

Another thing to note is that the SBOM API and SBOM CLI accept slightly different arguments for Verbosity. The SBOM API expects values from the EventLevel enum while the CLI accepts these specific args:
- Verbose
- Debug
- Information
- Fatal
- Error
- Warning

Therefore, I made it so that we accept the inputs for the SBOM CLI, but converted them to the respective EventLevel value for the API.

### DIFF
--- a/src/Microsoft.Sbom.Targets/SbomCLIToolTask.cs
+++ b/src/Microsoft.Sbom.Targets/SbomCLIToolTask.cs
@@ -3,8 +3,8 @@
 
 namespace Microsoft.Sbom.Targets;
 
+using System.Diagnostics.Tracing;
 using System.IO;
-using Microsoft.Build.Framework;
 using Microsoft.Build.Utilities;
 
 /// <summary>
@@ -29,7 +29,41 @@ public partial class GenerateSbom : ToolTask
     /// <returns>string list of args</returns>
     protected override string GenerateCommandLineCommands()
     {
-        return "Command";
+        var arguments =
+                "generate " +
+                $"-BuildDropPath {this.BuildDropPath} " +
+                $"-BuildComponentPath {this.BuildComponentPath} " +
+                $"-PackageName {this.PackageName} " +
+                $"-PackageVersion {this.PackageVersion} " +
+                $"-PackageSupplier {this.PackageSupplier} " +
+                $"-NamespaceUriBase {this.NamespaceBaseUri} " +
+                $"-DeleteManifestDirIfPresent {this.DeleteManifestDirIfPresent} " +
+                $"-FetchLicenseInformation {this.FetchLicenseInformation} " +
+                $"-EnablePackageMetadataParsing {this.EnablePackageMetadataParsing} " +
+                $"-Verbosity {this.Verbosity} ";
+
+        // For optional arguments, append them only if they are specified by the user
+        if (!string.IsNullOrEmpty(this.ManifestDirPath))
+        {
+            arguments += $"-ManifestDirPath {this.ManifestDirPath} ";
+        }
+
+        if (!string.IsNullOrEmpty(this.ExternalDocumentListFile))
+        {
+            arguments += $"-ExternalDocumentListFile {this.ExternalDocumentListFile} ";
+        }
+
+        if (!string.IsNullOrEmpty(this.NamespaceUriUniquePart))
+        {
+            arguments += $"-NamespaceUriUniquePart {this.NamespaceUriUniquePart} ";
+        }
+
+        if (!string.IsNullOrEmpty(this.ManifestInfo))
+        {
+            arguments += $"-ManifestInfo {this.ManifestInfo} ";
+        }
+
+        return arguments;
     }
 
     /// <summary>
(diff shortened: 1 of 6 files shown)


src/Microsoft.Sbom.Targets/SbomCLIToolTask.cs
```diff
@@ -38,6 +72,30 @@ protected override string GenerateCommandLineCommands()
...
+    /// <param name="eventLevel"></param>
+    private void SetOutputImportance(EventLevel eventLevel)
+    {
+        this.StandardOutputImportance = "High";
+
+        if (eventLevel == EventLevel.Critical)
+        {
+            this.StandardOutputImportance = "Low";
+        }
+    }
```
gustavoaca1997, line 100: Should we map the Verbosity to an `EventLevel` value for the CLI tool? It looks like we only need to check if the verbosity is _Fatal_ for setting the `StandardOutputImportante` to `"Low"`

src/Microsoft.Sbom.Targets/SbomCLIToolTask.cs
```diff
@@ -29,7 +29,41 @@ protected override string GenerateFullPathToTool()
...
+                $"-BuildDropPath {this.BuildDropPath} " +
+                $"-BuildComponentPath {this.BuildComponentPath} " +
+                $"-PackageName {this.PackageName} " +
+                $"-PackageVersion {this.PackageVersion} " +
+                $"-PackageSupplier {this.PackageSupplier} " +
+                $"-NamespaceUriBase {this.NamespaceBaseUri} " +
+                $"-DeleteManifestDirIfPresent {this.DeleteManifestDirIfPresent} " +
+                $"-FetchLicenseInformation {this.FetchLicenseInformation} " +
+                $"-EnablePackageMetadataParsing {this.EnablePackageMetadataParsing} " +
+                $"-Verbosity {this.Verbosity} ";
```
gustavoaca1997, line 43: Enclose the paths in quotes. Something like ` $"-BuildDropPath \"{this.BuildDropPath}\" "`

src/Microsoft.Sbom.Targets/SbomCLIToolTask.cs
```diff
@@ -29,7 +29,41 @@ protected override string GenerateFullPathToTool()
...
+
+        if (!string.IsNullOrEmpty(this.NamespaceUriUniquePart))
+        {
+            arguments += $"-NamespaceUriUniquePart {this.NamespaceUriUniquePart} ";
+        }
+
+        if (!string.IsNullOrEmpty(this.ManifestInfo))
+        {
+            arguments += $"-ManifestInfo {this.ManifestInfo} ";
+        }
```
gustavoaca1997, line 64: Consider using StringBuilder rather than concatenating strings: https://learn.microsoft.com/en-us/dotnet/standard/base-types/stringbuilder

gustavoaca1997 (changes requested):

src/Microsoft.Sbom.Targets/SbomCLIToolTask.cs
vpatakottu: Updated to remove the EventLevel logic for the CLI

src/Microsoft.Sbom.Targets/SbomCLIToolTask.cs
vpatakottu: Updated! The CommandLineBuilder automatically appends quotes for the arguments

src/Microsoft.Sbom.Targets/SbomCLIToolTask.cs
vpatakottu: Good idea! I saw there was a [CommandLineBuilder](https://learn.microsoft.com/en-us/dotnet/api/microsoft.build.utilities.commandlinebuilder?view=msbuild-17-netcore) class we can use to construct command line arguments. I saw this was used in the SDK project example as well

gustavoaca1997 (approved):

src/Microsoft.Sbom.Targets/SbomInputValidator.cs
```diff
@@ -0,0 +1,135 @@
...
+public partial class GenerateSbom
+{
+    /// <summary>
+    /// Ensure all required arguments are non-null/empty,
+    /// and do not contain whitespaces, tabs, or newline characters.
+    /// </summary>
+    /// <returns>True if the required parameters are valid. False otherwise.</returns>
+    public bool ValidateAndSanitizeRequiredParams()
+    {
+        if (string.IsNullOrWhiteSpace(this.BuildDropPath))
```
nschwerzler, line 22: Suggestion:

```
var propertiesToCheck = new Dictionary<string, string>
{
    { nameof(this.BuildDropPath), this.BuildDropPath },
    { nameof(this.PackageSupplier), this.PackageSupplier },
    { nameof(this.PackageName), this.PackageName },
    { nameof(this.PackageVersion), this.PackageVersion },
    { nameof(this.NamespaceBaseUri), this.NamespaceBaseUri }
};

foreach (var property in propertiesToCheck)
{
    if (string.IsNullOrWhiteSpace(property.Value))
    {
        Log.LogError($"SBOM generation failed: Empty argument detected for {property.Key}. Please provide a valid value.");
        return false;
    }
}
```

nschwerzler (approved):

src/Microsoft.Sbom.Targets/SbomInputValidator.cs
vpatakottu: Updated!

PR #619, opened 2024-07-17
Update NuGet Package Format and Surface Errors

vpatakottu:
Modified our .targets files to use <TargetsForTfmSpecificContentInPackage> in order to copy the SBOM and sha256 hash into the user's nuget package near the binaries. For scenarios where users append "-windows" to the target framework, we also need to append the <TargetPlatformVersion> as well because NuGet renames the folder to "net8.0-windows7.0" in the nuget package.

These changes also set a ToolTask property that will log any errors from the CLI tool as ToolTask errors, so the messages will be surfaced to Visual Studio's output. There is also extra logic added to create the ManifestDirPath directory if it doesn't exist, minimizing the need for users to create it themselves. This is automatically done by the SBOM API, but is not done by the CLI tool; therefore, this logic was added to the ToolTask.

### DIFF
--- a/src/Microsoft.Sbom.Targets/Microsoft.Sbom.Targets.targets
+++ b/src/Microsoft.Sbom.Targets/Microsoft.Sbom.Targets.targets
@@ -74,10 +74,15 @@
 
   <!-- Specify the SBOM files to copy into the nuget package -->
   <Target Name="CopySbomOutput" DependsOnTargets="GenerateSbomTarget">
+    <PropertyGroup>
+      <!--When building frameworks such as net8.0-windows, the platform version is appended to the framework in the NuGet package-->
+      <TargetFrameworkWithPlatformVersion Condition="$(TargetPlatformVersion) != ''">$(TargetFramework)$(TargetPlatformVersion)</TargetFrameworkWithPlatformVersion>
+      <TargetFrameworkWithPlatformVersion Condition="$(TargetPlatformVersion) == ''">$(TargetFramework)</TargetFrameworkWithPlatformVersion>
+    </PropertyGroup>
     <ItemGroup>
       <!--Add manifest and SHA file from the GenerateSbom target execution-->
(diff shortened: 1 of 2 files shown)


nschwerzler (approved):

src/Microsoft.Sbom.Targets/Microsoft.Sbom.Targets.targets
```diff
@@ -58,23 +67,34 @@
...
 
-    <!-- Include the generated SBOM contents within the consumer's nuget package -->
-    <ItemGroup >
-      <Content Condition=" '$(MSBuildRuntimeType)' == 'Core'" Include="$(SbomPathResult)\**">
-        <Pack>true</Pack>
-        <PackagePath>_manifest</PackagePath>
-      </Content>
+      <!--Add manifest and SHA file from the SbomCLIToolTask execution from the default manifest path -->
+      <BuildOutputInPackage Condition=" '$(MSBuildRuntimeType)' == 'Full' And '$(SbomGenerationManifestDirPath)' == ''"
+								  Include="$(SbomGenerationBuildDropPath)\$(ManifestFolderName)\$(SbomSpecification)\manifest.spdx.json"
```
gustavoaca1997, line 85: Actually, this should be set to `ManifestDirPath` if it's not null or to $(SbomGenerationBuildDropPath)\$(ManifestFolderName) if it's null. We should return it from the ToolTask to make it consistent with the Task.

gustavoaca1997 (changes requested):

src/Microsoft.Sbom.Targets/Microsoft.Sbom.Targets.targets
vpatakottu: Updated! Changed this to make the ToolTask set the SbomPath output variable we have

nschwerzler (approved):

gustavoaca1997 (approved):

PR #656, opened 2024-08-08
Workaround for generating a SBOM manifest at the root level of the Nuget Package

gustavoaca1997:
# Issue
Ideally, we want a SBOM manifest that would include all the content of the Nuget Package, however it was concluded that the best way of doing this is by waiting for [the .NET Team to implement a pre-compress event that we can hook in](https://github.com/NuGet/Home/issues/13690). In the meantime, it was decided to at least generate smaller SBOMs that include the binaries that are produced during build. 

Unfortunately, we have encountered multiple issues, including dealing with file extensions that are not included by default in the Nuget Packages (such as .exe or .deps.json) but that were included in the SBOM manifest. This makes **the SBOM not able to be validated**.

# Workaround
Having said that, we propose a workaround for generating a SBOM at the root level, that is able to be validated, while we wait for the pre-compress hook. The proposal consists of unzipping the Nuget Package, generating the SBOM, and zipping again. 

### .NET Standard project
In average, the workaround has a better performance than the original solution. I ran some performance tests using **a .NET Standard** project we use in production:
| | Generate SBOM per framework | Generate SBOM per package (with workaround) |
| -------- | ------- | -------- |
| Avg. Time | 2.95 | 2.85 |

However, in one of the performance tests, there was one outlier of 3.767 seconds for the workaround, probably due to a cold-start.

### Multi-framework project
I also ran some performance tests using a dummy multi-framework project:
| | Generate SBOM per framework | Generate SBOM per package (with workaround) |
| -------- | ------- | -------- |
| Avg. Time | 3.15 | 3.16 |

However, there were two cold-start runs that gave outlier results:
| | Generate SBOM per framework | Generate SBOM per package (with workaround) |
| -------- | ------- | -------- |
| Cold-Start Time | 4.98 | 4.22 |

### DIFF
--- a/src/Microsoft.Sbom.Targets/Microsoft.Sbom.Targets.targets
+++ b/src/Microsoft.Sbom.Targets/Microsoft.Sbom.Targets.targets
@@ -39,6 +39,7 @@
     <SbomGenerationVerbosity Condition=" '$(SbomGenerationVerbosity)' == '' ">information</SbomGenerationVerbosity>
     <SbomGenerationManifestInfo Condition=" '$(SbomGenerationManifestInfo)' == '' ">SPDX:2.2</SbomGenerationManifestInfo>
     <SbomGenerationDeleteManifestDirIfPresent Condition=" '$(SbomGenerationDeleteManifestDirIfPresent)' == '' ">true</SbomGenerationDeleteManifestDirIfPresent>
+    <UnzipGuid>$([System.Guid]::NewGuid())</UnzipGuid>
   </PropertyGroup>
 
   <!-- After the Nuget Package is generated, we will unzip, scan, generate the SBOM and zip again. -->
@@ -51,7 +52,7 @@
         $(PackageOutputPath)\$(PackageId).$(PackageVersion).nupkg
       </NugetPackage>
       <NugetPackageUnzip>
-        $(PackageOutputPath)\$(PackageId).$(PackageVersion).nupkg.unzip
+        $(PackageOutputPath)\$(PackageId).$(PackageVersion).$(UnzipGuid).temp
       </NugetPackageUnzip>
     </PropertyGroup>
     <Unzip DestinationFolder="$(NugetPackageUnzip)" SourceFiles="$(NugetPackage)" OverwriteReadOnlyFiles="true" />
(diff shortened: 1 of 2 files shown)


src/Microsoft.Sbom.Targets/Microsoft.Sbom.Targets.targets
```diff
@@ -70,20 +76,9 @@
...
-      <TargetFrameworkWithPlatformVersion Condition="$(TargetPlatformVersion) == ''">$(TargetFramework)</TargetFrameworkWithPlatformVersion>
-    </PropertyGroup>
-    <ItemGroup>
-      <!--Add manifest and SHA file from the GenerateSbom target execution-->
-      <TfmSpecificPackageFile Include="$([System.IO.Path]::Combine($(SbomPathResult),$(SbomSpecification)))\**">
-        <PackagePath>$([System.IO.Path]::Combine($(BuildOutputTargetFolder),$(TargetFrameworkWithPlatformVersion),$(ManifestFolderName),$(SbomSpecification)))</PackagePath>
-      </TfmSpecificPackageFile>
-    </ItemGroup>
+    <!-- Zip the Nuget package back up and delete the temporary unzipped package. -->
+    <ZipDirectory SourceDirectory="$(NugetPackageUnzip)" DestinationFile="$(NugetPackage)" Overwrite="true" />
```
vpatakottu, line 81: Out of curiosity, about how much longer does the unzip/zip process take compared to our original method?

vpatakottu (approved):

src/Microsoft.Sbom.Targets/Microsoft.Sbom.Targets.targets
gustavoaca1997: I have updated the description with the performance:

### .NET Standard project
In average, the workaround has a better performance than the original solution. I ran some performance tests using **a .NET Standard** project we use in production:
| | Generate SBOM per framework | Generate SBOM per package (with workaround) |
| -------- | ------- | -------- |
| Avg. Time | 2.95 | 2.85 |

However, in one of the performance tests, there was one outlier of 3.767 seconds for the workaround, probably due to a cold-start.

### Multi-framework project
I also ran some performance tests using a dummy multi-framework project:
| | Generate SBOM per framework | Generate SBOM per package (with workaround) |
| -------- | ------- | -------- |
| Avg. Time | 3.15 | 3.16 |

However, there were two cold-start runs that gave outlier results:
| | Generate SBOM per framework | Generate SBOM per package (with workaround) |
| -------- | ------- | -------- |
| Cold-Start Time | 4.98 | 4.22 |

gustavoaca1997:
I've created this issue to the Nuget team: https://github.com/NuGet/Home/issues/13690

PR #658, opened 2024-08-10
Add E2E tests for Microsoft.Sbom.Targets project

vpatakottu:
Added E2E unit test project using the MSBuild API to dynamically restore, build, and pack a sample project with our SBOM generation imported. These tests are only supported on Windows.

This PR also contains changes to remove support for ManifestDirPath, as we want the SBOM to only be generated at the root of the NuGet package.


nschwerzler (approved):

test/Microsoft.Sbom.Targets.Tests/GenerateSbomE2ETests.cs
```diff
@@ -0,0 +1,458 @@
...
+    {
+        //Arrange
+        // Create a Project object for ProjectSample1
+        var sampleProject = (Project)project;
+
+        // Get all the expected default properties
+        GetDefaultProperties(sampleProject);
+
+        // Set the TargetFrameworks property to empty. By default, it sets this property to net6.0 and net8.0, which fails for net8.0 builds.
+        sampleProject.SetProperty("TargetFrameworks", string.Empty);
```
gustavoaca1997, line 164: If you need to do these steps for all the test cases, try to bundle them into one set-up method, that returns the `sampleProject`, or sets a private property `this.sampleProject` that then you use in the test cases.

test/Microsoft.Sbom.Targets.Tests/GenerateSbomE2ETests.cs
```diff
@@ -0,0 +1,458 @@
...
+        // Restore, build, and pack the project
+        RestoreBuildPack(sampleProject);
+
+        // Extract the NuGet package
+        ExtractPackage();
+
+        manifestPath = Path.Combine(manifestDirPath, "_manifest", sbomSpecificationDirectoryName, "manifest.spdx.json");
+
+        // Check if the SBOM exists in the ManifestDirPath
+        Assert.IsTrue(File.Exists(manifestPath));
```
gustavoaca1997, line 208: I think this should test that: If we set the `manifestDirPath` to a relative path, the relative manifest path inside the Nuget package is the same and the content is valid.

We could also test what happens when we set `manifestDirPath` to an absolute path.

gustavoaca1997 (changes requested):
Two important things:
- Review better alternatives for adapting the Targets file for our tests. I see the E2E tests as the tests for the Targets file layer, while the other tests were for the back-end layer.
- Review the ManifestDirPath E2E tests. I left some comments there.

test/Microsoft.Sbom.Targets.Tests/AbstractGenerateSbomTaskTests.cs
```diff
@@ -58,6 +59,16 @@ private void CleanupManifestDirectory()
 
     [TestInitialize]
     public void Startup()
+    {
+        if (MSBuildLocator.CanRegister)
```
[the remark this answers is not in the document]
vpatakottu: Yes, the Microsoft.Build.Locator package apparently contains all the other Microsoft.Build.* assemblies within it. If we remove these lines, then we get missing assembly errors. I was following this guidance: https://learn.microsoft.com/en-us/visualstudio/msbuild/find-and-use-msbuild-versions?view=vs-2022#use-nuget-packages-preferred

test/Microsoft.Sbom.Targets.Tests/Microsoft.Sbom.Targets.Tests.csproj
```diff
@@ -15,12 +16,24 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Build.Framework" />
-    <PackageReference Include="Microsoft.Build.Utilities.Core" />
+    <PackageReference Include="Microsoft.Build" ExcludeAssets="Runtime" />
+    <PackageReference Include="Microsoft.Build.Framework" ExcludeAssets="Runtime" />
```
gustavoaca1997, line 20: Why do we need `ExcludeAssets="Runtime"`?

test/Microsoft.Sbom.Targets.Tests/AbstractGenerateSbomTaskInputTests.cs
```diff
@@ -35,7 +36,16 @@ public abstract class AbstractGenerateSbomTaskInputTests
     [TestInitialize]
     public void Startup()
     {
-        // Setup the build engine
+        if (MSBuildLocator.CanRegister)
```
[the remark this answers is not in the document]
vpatakottu: Updated it so that the E2E project is separate from the unit tests so we don't need to use the MSBuildLocator for these

test/Microsoft.Sbom.Targets.Tests/ProjectSamples/Microsoft.Sbom.Targets.Test.targets
```diff
@@ -0,0 +1,76 @@
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
```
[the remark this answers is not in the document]
vpatakottu: Updated to copy this file during runtime!

test/Microsoft.Sbom.Targets.Tests/GenerateSbomE2ETests.cs
```diff
@@ -0,0 +1,458 @@
...
+            ProjectCollection.GlobalProjectCollection.UnloadAllProjects();
+            project = null;
+        }
+        catch (Exception ex)
+        {
+            Assert.Fail($"Failed to cleanup output directories. {ex}");
+        }
+    }
+
+    private void GetDefaultProperties(Project sampleProject)
```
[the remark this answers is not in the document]
vpatakottu: Updated!

test/Microsoft.Sbom.Targets.Tests/GenerateSbomE2ETests.cs
vpatakottu: Updated!

test/Microsoft.Sbom.Targets.Tests/GenerateSbomE2ETests.cs
vpatakottu: I removed the ability to set ManifestDirPath altogether, as the desired location for the SBOM is at the root of the nuget package aka, BuildDropPath. For now I left the ManifestDirPath parameter empty in case people want to implement this in the future, but let me know if I should set this to something.

src/Microsoft.Sbom.Targets/Microsoft.Sbom.Targets.targets
```diff
@@ -39,13 +30,11 @@
     <SbomGenerationVerbosity Condition=" '$(SbomGenerationVerbosity)' == '' ">information</SbomGenerationVerbosity>
     <SbomGenerationManifestInfo Condition=" '$(SbomGenerationManifestInfo)' == '' ">SPDX:2.2</SbomGenerationManifestInfo>
     <SbomGenerationDeleteManifestDirIfPresent Condition=" '$(SbomGenerationDeleteManifestDirIfPresent)' == '' ">true</SbomGenerationDeleteManifestDirIfPresent>
-    <UnzipGuid>$([System.Guid]::NewGuid())</UnzipGuid>
+    <UnzipGuid>$([System.DateTime]::Now.Ticks)</UnzipGuid>
```
nschwerzler, line 33: UnzipDirectoryName

nschwerzler (approved):

nschwerzler (approved):

PR #670, opened 2024-08-15
Bump CD version to support performance improvement on pip report

JoseRenan:
Component Detection 4.9.3 added the environment variable `PipReportPersistReports` that enables storing the pip reports for tasks running CD more than once inside it.


DaveTryon:
This is superseded by #755

Issue #714, opened 2024-09-19
[Microsoft.Sbom.Targets] Mark as DevelopmentDependency


KalleOlaviNiemitalo:
`Microsoft.Sbom.Targets.csproj` should set `<DevelopmentDependency>true</DevelopmentDependency>` so that, when `dotnet add package` adds the Microsoft.Sbom.Targets package to a project, the PackageReference item gets this kind of metadata by default:

```XML
    <PackageReference Include="Microsoft.Sbom.Targets" Version="2.2.8">
      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
      <PrivateAssets>all</PrivateAssets>
    </PackageReference>
```

Then, any NuGet package built from that project will not declare a dependency on Microsoft.Sbom.Targets.  The DevelopmentDependency setting is documented in <https://learn.microsoft.com/nuget/reference/msbuild-targets#pack-target> and <https://github.com/NuGet/Home/wiki/DevelopmentDependency-support-for-PackageReference>.

DevelopmentDependency does not matter if the PackageReference is automatically added by .NET SDK as in <https://github.com/dotnet/sdk/pull/43151>, because `dotnet add package` is not used in this case; but it can be useful if the package is used with an older SDK that does not do so, or perhaps if the developer wants to use a specific version of Microsoft.Sbom.Targets instead of the SDK default.

_Originally posted by @KalleOlaviNiemitalo in <https://github.com/microsoft/sbom-tool/pull/674#discussion_r1751708750> and then <https://github.com/microsoft/sbom-tool/issues/693#issuecomment-2340388481>_

Issue #713, opened 2024-09-19
[Microsoft.Sbom.Targets] Spaces between words of $(SbomGenerationPackageSupplier) are removed


KalleOlaviNiemitalo:
Spaces between words of `$(SbomGenerationPackageSupplier)` are removed, e.g. "Contoso Catering GmbH" becomes "ContosoCateringGmbH".   I guess it's done here: <https://github.com/microsoft/sbom-tool/blob/08ba73d303228eb4d92a6a5f75350d78230bca30/src/Microsoft.Sbom.Targets/SbomInputValidator.cs#L44>

Why was it implemented like that?  The `sbom-tool -ps` option preserves internal spaces; and if the NuGetComponentDetector finds a `*.nuspec` file and reads the `authors` element, that too preserves internal spaces.

_Originally posted by @KalleOlaviNiemitalo in https://github.com/microsoft/sbom-tool/issues/693#issuecomment-2341446243_

PR #726, opened 2024-09-26
Make targets package a dev dependency and fix package supplier trimming

sfoslund:
Fixes https://github.com/microsoft/sbom-tool/issues/713
Fixes https://github.com/microsoft/sbom-tool/issues/714

### DIFF
--- a/src/Microsoft.Sbom.Targets/SbomInputValidator.cs
+++ b/src/Microsoft.Sbom.Targets/SbomInputValidator.cs
@@ -41,7 +41,7 @@ public partial class GenerateSbom
             }
         }
 
-        this.PackageSupplier =this.PackageSupplier.Trim();
+        this.PackageSupplier = this.PackageSupplier.Trim();
         this.PackageName = Remove_Spaces_Tabs_Newlines(this.PackageName);
         this.PackageVersion = Remove_Spaces_Tabs_Newlines(this.PackageVersion);
         this.NamespaceBaseUri = this.NamespaceBaseUri.Trim();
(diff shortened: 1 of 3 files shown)


src/Microsoft.Sbom.Targets/SbomInputValidator.cs
```diff
@@ -41,18 +41,18 @@ public bool ValidateAndSanitizeRequiredParams()
             }
         }
 
-        this.PackageSupplier = Remove_Spaces_Tabs_Newlines(this.PackageSupplier);
-        this.PackageName = Remove_Spaces_Tabs_Newlines(this.PackageName);
-        this.PackageVersion = Remove_Spaces_Tabs_Newlines(this.PackageVersion);
+        this.PackageSupplier = Remove_Tabs_Newlines(this.PackageSupplier);
+        this.PackageName = Remove_Tabs_Newlines(this.PackageName);
+        this.PackageVersion = Remove_Tabs_Newlines(this.PackageVersion);
```
DaveTryon, line 46: PackageSupplier and PackageName should definitely allow inner spaces, but I'm less convinced about PackageVersion. Based on the docs at https://learn.microsoft.com/en-us/nuget/concepts/package-versioning and the parsing tests at https://github.com/NuGet/NuGet.Client/blob/dev/test/NuGet.Core.Tests/NuGet.Versioning.Test/NuGetVersionTest.cs#L90-L166 (this library is recommended from https://learn.microsoft.com/en-us/nuget/concepts/package-versioning?tabs=semver20sort#where-nugetversion-diverges-from-semantic-versioning), I would expect that the PackageVersion string would _not_ allow inner spaces

DaveTryon (approved):

PR #734, opened 2024-10-03
Remove static keyword from Configuration class objects

DaveTryon:
When I run the unit tests inside Visual Studio, the .NET 8.0 build of `ConfigSanitizerTests` has a persistent quirk. When I run the full suite of tests, `ShouldGetPackageSupplierFromAsseblyConfig_Succeeds` _always_ fails. When I run this specific test by itself, it _always_ succeeds.

After some sleuthing, I determined that when the tests get run as a group, `ShouldGetPackageSupplierFromAsseblyConfig_Succeeds` contains a non-default value for `configuration.PackageSupplier`, while the same field has a default value when the test runs by itself. We build a new `Configuration` object for each test, so we should expect to have the default values. Digging into the `Configuration` class, its backing objects all use the `static` keyword, meaning that the properties are _shared_ between different `Configuration` instances. The `ShouldNotOverridePackageSupplierIfProvided_Succeeds` happens to run first, and it sets `config.PackageSupplier` to the non-default value. The non-default value then leaks to `ShouldGetPackageSupplierFromAsseblyConfig_Succeeds`, causing the test to fail. Oops!

I could have fixed the unit test by just resetting the property, but that would leave us with this shared `Configuration` state, which I don't believe was ever the intent of this class. This change removes the `static` keyword so each new `Configuration` object is created with a known default state.


DaveTryon:
Thanks for the input, @KalleOlaviNiemitalo! As it turns out, this problem no longer occurs in the current main branch. I think it may have been an artifact of .NET 6.0 behavior. We won't be merging this in.

PR #737, opened 2024-10-04
Update Sbom targets E2E tests to .net8

vpatakottu:
Added some NuGet packages that were required to make the E2E tests compatible with .NET8

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -1,63 +1,65 @@
 <!-- All csproj package references should not include version numbers. The version numbers are set using this props file. -->
 <Project>
-    <ItemDefinitionGroup>
-        <PackageVersion>
-            <!-- Do not share compile-time dependencies transitively.  This requires that all projects reference all packages -->
-            <PrivateAssets>Compile</PrivateAssets>
-        </PackageVersion>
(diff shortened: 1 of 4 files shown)


Directory.Packages.props
```diff
@@ -1,63 +1,65 @@
...
-        <PackageVersion Include="System.Reactive" Version="5.0.0" />
-        <PackageVersion Include="System.Runtime.Loader" Version="4.3.0" />
-        <PackageVersion Include="System.Text.Json" Version="8.0.4" />
-        <PackageVersion Include="System.Threading.Channels" Version="6.0.0" />
-        <PackageVersion Include="System.Threading.Tasks.Dataflow" Version="4.11.1" />
-        <PackageVersion Include="System.Threading.Tasks.Extensions" Version="4.5.4" />
-    </ItemGroup>
-</Project>
+  <ItemDefinitionGroup>
+    <PackageVersion>
```
DaveTryon, line 4: I'm not sure why this file uses 4-spaces for each indent, but let's maintain that if possible. If/when we decide to change the indents, let's put that into a separate PR.

vpatakottu: Updated to fix the indents!

PR #754, opened 2024-10-15
Add dependency graph support to remaining ecosystems

jalkire:
This PR builds off of #746 to add support for ancestral dependencies of the remaining ecosystems supported by CD. The list of ecosystems supporting graph creation [can be found here](https://github.com/microsoft/component-detection?tab=readme-ov-file#features).

### DIFF
--- a/src/Microsoft.Sbom.Adapters/Adapters/ComponentDetection/CargoComponentExtensions.cs
+++ b/src/Microsoft.Sbom.Adapters/Adapters/ComponentDetection/CargoComponentExtensions.cs
@@ -3,6 +3,7 @@
 
 namespace Microsoft.Sbom.Adapters.ComponentDetection;
 
+using System.Linq;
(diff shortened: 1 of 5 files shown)


DaveTryon (approved):
Are there any concerns caused by a diamond pattern? For instance, if my app has direct dependencies on package X and package Y, and those packages each have a direct dependency on package Z, does package Z get included as a dependency of _both X and Y_, or _either X or Y_ ?

jalkire:
> Are there any concerns caused by a diamond pattern? For instance, if my app has direct dependencies on package X and package Y, and those packages each have a direct dependency on package Z, does package Z get included as a dependency of _both X and Y_, or _either X or Y_ ?

That's a good question! As it stands in this and the previous PR, the latter would occur (either X or Y). We primarily care about differentiating between direct and indirect dependencies, but I will double check if the both X and Y behavior is preferable.

PR #756, opened 2024-10-16
Add support for osx-arm64

DaveTryon:
This adds the ability to produce an `osx-arm64` version of the tool. It also includes a small doc change about building locally--local builds of arm-64 should not be treated as official since we will be publishing this as a separate package.


jalkire (approved):
Is there any kind of signing that should occur for Mac?

PR #767, opened 2024-10-29
Blunt update of SPDX version 2.2 -> 2.2.2

### LINKED ISSUE #738: Support German BSI TR-03183 2.0.0 by supporting SPDX 2.2.1
From [Version 2.0.0](https://www.bsi.bund.de/EN/Themen/Unternehmen-und-Organisationen/Standards-und-Zertifizierung/Technische-Richtlinien/TR-nach-Thema-sortiert/tr03183/TR-03183_node.html) of the BSI TR-03183 guideline which is just released on 20 Sep 2024, the minimum required SPDX version is now 2.2.1 (page 9):

> 4 SBOM formats
>
> A newly generated or updated SBOM MUST be in JSON- or XML-format that meets one of the following
specifications in one of the specified versions.
>
> - CycloneDX9, version 1.5 or higher
> - Software Package Data eXchange (SPDX), version 2.2.1 or higher
>
> The transitional system of this Technical Guideline is specified in section 7

SPDX 2.2.1 is the one that is ISO standard: https://www.iso.org/standard/81870.html

**There's no technical differences between V2.2 and V2.2.1**, according to SPDX's [Differences between V2.2.1 and V2.2](https://github.com/spdx/using/blob/main/docs/diffs-from-previous-editions.md#a4-differences-between-v221-and-v22-) documentation.

> A.4 Differences between V2.2.1 and V2.2
>
> There were no technical differences; V2.2.1 is V2.2 reformatted for submission to ISO via the PAS process. As a result, new clauses were added causing the previous clause-numbering sequence to change. Also, Annexes went from having Roman numbers to Latin letters. Here is the translation between numbering in V2.2.1 and the version that came before it:

So it is very likely that SBOM Tool can generate the same SBOM (2.2) and merely change "spdxVersion" to "SPDX-2.2.1" to support SPDX 2.2.1.

bact:
- This is a blunt and aggressive approach of making sbom-tool supporting SPDX 2.2.2 (in order to fix #738).
- It does not add 2.2.2 support AND keep 2.2 support, it is replacing 2.2 with 2.2.2. There will be no 2.2.
- It is basically replacing the version number (`SPDXConstants.SPDXVersion`) from "2.2" to "2.2.2" when creating `SbomSpecification`.
- As there's no technical differences between 2.2, 2.2.1, and 2.2.2, this can be a minimal way to support 2.2.2.
- Considering 2.2.2 is a patch version of 2.2 (and there's no technical differences), we may able to say it is still "supporting" 2.2.

### DIFF
--- a/docs/sbom-tool-api-reference.md
+++ b/docs/sbom-tool-api-reference.md
@@ -106,7 +106,7 @@ The `SBOMSpecificiation` object represents a SBOM format. Each `SBOMSpecificatio
 ```C#
 using Microsoft.Sbom.Contracts;
 
-var spdx22Specification = new SBOMSpecification("SPDX", "2.2");
+var spdx22Specification = new SBOMSpecification("SPDX", "2.2.2");
 ```
 
 While this API supports the creation of a SBOM output file in multiple formats, it currently only supports the SPDX version 2.2 architecture. Users looking to implement other SBOM architectures can use this API call, which provides the full list of all supported formats.
@@ -118,7 +118,7 @@ var specifications = generator.GetSupportedSBOMSpecifications();
 
 Assert.True(specifications.Count() == 1);
 Assert.Equal("SPDX", specifications.First().Name);
-Assert.Equal("2.2", specifications.First().Version);
+Assert.Equal("2.2.2", specifications.First().Version);
 ```
 
 ### GetRequiredAlgorithms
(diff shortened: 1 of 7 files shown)


bact:
@KalleOlaviNiemitalo  I assume and expect that this command

> `sbom-tool Generate -ManifestInfo SPDX:2.2`

will just generate 2.2.2 (and any 2.2.x, as a feature-compatible version of 2.2).

DaveTryon:
@KalleOlaviNiemitalo and @bact, we're discussing this internally to determine our best path forward. We have internal validation tooling that expects the `spdx_2.2` folder, so that will break if we suddenly change the folder name to `spdx_2.2.2` We haven't yet determined exactly how we want to proceed to provide the optimal experience.

@THEJRRR is the person driving the requirements, so I'm adding him to the thread.

DaveTryon:
It's unlikely that we'll merge this PR. May I suggest moving the discussion to #738?

bact:
Please move the discussion over. Thank you.

Issue #584, opened 2024-05-27
Timeout when fetching license information


erri120:
The timeout of 30 seconds is a bit too short for larger projects:

```log
##[debug]Retrieving license information for 500 components...
##[warning]Error encountered while fetching license information from API, resulting SBOM may have incomplete license information: The request was canceled due to the configured HttpClient.Timeout of 30 seconds elapsing.
```

Would it be possible to increase the timeout or have a configuration option?

https://github.com/microsoft/sbom-tool/blob/0b4db48ca4f235cb75d9deaff2e094f710d55b2d/src/Microsoft.Sbom.Api/Executors/LicenseInformationService.cs#L23

https://github.com/microsoft/sbom-tool/blob/0b4db48ca4f235cb75d9deaff2e094f710d55b2d/src/Microsoft.Sbom.Api/Executors/LicenseInformationService.cs#L41

LucaBlackDragon:
I'm facing the same issue in what I'd consider a small project (102 components in total, 27 explicitly referenced).
Maybe using a lower `batchSize` would help preventing timeouts:
https://github.com/microsoft/sbom-tool/blob/0b4db48ca4f235cb75d9deaff2e094f710d55b2d/src/Microsoft.Sbom.Api/Executors/LicenseInformationService.cs#L34
But it would also increase the risk of hitting the [documented rate limits](https://docs.clearlydefined.io/docs/get-involved/using-data#rest-api-usage-considerations), so all things considered, I think a configurable timeout would be the best solution.

georg-eckert-zeiss:
Same here. But I don't know the root cause for the timeout. Very large project.

philipp-naused:
We have a project with 121 and it times out every single time. Please make this configurable.

kidcline1:
I've submitted a PR that adds a CLI argument to configure this.

erri120:
PR is #773.

DaveTryon:
This was released in version 3.1.0, released on 4 February 2025. I forgot to close this out after the release.

PR #773, opened 2024-11-01
Made the Timeout in LicenseInformationService configurable via CLI argument (#584)

kidcline1:
- Added new command line argument for generator (lto)
- Modified LicenseInformationService and LicenseInformationFetcher to allow passing timeout as an argument
- Removed hardcoded limit of 30 seconds

### DIFF
--- a/src/Microsoft.Sbom.Api/Config/Args/GenerationArgs.cs
+++ b/src/Microsoft.Sbom.Api/Config/Args/GenerationArgs.cs
@@ -117,6 +117,15 @@ public class GenerationArgs : GenerationAndValidationCommonArgs
     public bool? FetchLicenseInformation { get; set; }
 
     /// <summary>
+    /// Specifies the timeout in seconds for fetching the license information. Defaults to 30 seconds. Has no effect if
+    /// FetchLicenseInformation (li) argument is false or not provided.
+    /// </summary>
+    [ArgShortcut("lto")]
+    [ArgDescription("Specifies the timeout in seconds for fetching the license information. Defaults to 30 seconds. " +
+        "Has no effect if the FetchLicenseInformation (li) argument is false or not provided. ")]
+    public int? LicenseInformationTimeout { get; set; }
+
+    /// <summary>
     /// If set to true, we will attempt to parse license and supplier info from the packages metadata file.
     /// </summary>
     [ArgShortcut("pm")]
--- a/src/Microsoft.Sbom.Api/Config/Args/GenerationArgs.cs
+++ b/src/Microsoft.Sbom.Api/Config/Args/GenerationArgs.cs
@@ -118,12 +118,13 @@ public class GenerationArgs : GenerationAndValidationCommonArgs
 
     /// <summary>
     /// Specifies the timeout in seconds for fetching the license information. Defaults to 30 seconds. Has no effect if
-    /// FetchLicenseInformation (li) argument is false or not provided.
+    /// FetchLicenseInformation (li) argument is false or not provided. A negative value corresponds to an infinite timeout.
     /// </summary>
     [ArgShortcut("lto")]
     [ArgDescription("Specifies the timeout in seconds for fetching the license information. Defaults to 30 seconds. " +
-        "Has no effect if the FetchLicenseInformation (li) argument is false or not provided. ")]
-    public int? LicenseInformationTimeout { get; set; }
+        "Has no effect if the FetchLicenseInformation (li) argument is false or not provided. A negative value corresponds" +
+        "to an infinite timeout.")]
+    public int? LicenseInformationTimeoutInSeconds { get; set; }
 
     /// <summary>
     /// If set to true, we will attempt to parse license and supplier info from the packages metadata file.
--- a/src/Microsoft.Sbom.Api/Config/ConfigSanitizer.cs
+++ b/src/Microsoft.Sbom.Api/Config/ConfigSanitizer.cs
@@ -83,11 +83,21 @@ public class ConfigSanitizer
         configuration.PackageSupplier = GetPackageSupplierFromAssembly(configuration, logger);
 
         // Prevent null value for LicenseInformationTimeoutInSeconds.
-        // Values from int.MinValue to int.MaxValue are allowed. Negative values are simply treated as InfiniteTimespan in LicenseInformationService, and
-        // from what I can tell there are no upper limits higher than int.MaxValue in Timespan.FromSeconds or httpClient.Timeout
+        // Values of (0, Constants.MaxLicenseFetchTimeoutInSeconds] are allowed. Negative values are replaced with the default, and
+        // the higher values are truncated to the maximum of Common.Constants.MaxLicenseFetchTimeoutInSeconds
         if (configuration.LicenseInformationTimeoutInSeconds is null)
         {
-            configuration.LicenseInformationTimeoutInSeconds = new(LicenseInformationFetcher.DefaultTimeoutInSeconds, SettingSource.Default);
+            configuration.LicenseInformationTimeoutInSeconds = new(Common.Constants.DefaultLicenseFetchTimeoutInSeconds, SettingSource.Default);
+        }
+        else if (configuration.LicenseInformationTimeoutInSeconds.Value <= 0)
+        {
+            logger.Warning($"Negative and Zero Values not allowed for timeout. Using the default {Common.Constants.DefaultLicenseFetchTimeoutInSeconds} seconds instead.");
+            configuration.LicenseInformationTimeoutInSeconds.Value = Common.Constants.DefaultLicenseFetchTimeoutInSeconds;
+        }
+        else if (configuration.LicenseInformationTimeoutInSeconds.Value > Common.Constants.MaxLicenseFetchTimeoutInSeconds)
+        {
+            logger.Warning($"Specified timeout exceeds maximum allowed. Truncating the timeout to {Common.Constants.MaxLicenseFetchTimeoutInSeconds} seconds.");
+            configuration.LicenseInformationTimeoutInSeconds.Value = Common.Constants.MaxLicenseFetchTimeoutInSeconds;
         }
 
         // Check if arg -lto is specified but -li is not
--- a/src/Microsoft.Sbom.Api/Config/ConfigSanitizer.cs
+++ b/src/Microsoft.Sbom.Api/Config/ConfigSanitizer.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Reflection;
+using Microsoft.Sbom.Api.Executors;
 using Microsoft.Sbom.Api.Hashing;
 using Microsoft.Sbom.Api.Utils;
 using Microsoft.Sbom.Common;
@@ -80,6 +81,20 @@ public class ConfigSanitizer
 
         // Set default package supplier if not provided in configuration.
         configuration.PackageSupplier = GetPackageSupplierFromAssembly(configuration, logger);
+
+        // Prevent null value for LicenseInformationTimeoutInSeconds.
+        // Values from int.MinValue to int.MaxValue are allowed. Negative values are simply treated as InfiniteTimespan in LicenseInformationService, and
+        // from what I can tell there are no upper limits higher than int.MaxValue in Timespan.FromSeconds or httpClient.Timeout
+        if (configuration.LicenseInformationTimeoutInSeconds is null)
+        {
+            configuration.LicenseInformationTimeoutInSeconds = new(LicenseInformationFetcher.DefaultTimeoutInSeconds, SettingSource.Default);
+        }
+
+        // Check if arg -lto is specified but -li is not
+        if (configuration.FetchLicenseInformation?.Value != true && !configuration.LicenseInformationTimeoutInSeconds.IsDefaultSource)
+        {
+            logger.Warning("A license fetching timeout is specified (argument -lto), but this has no effect when FetchLicenseInfo is unspecified or false (argument -li)");
+        }
 
         // Replace backslashes in directory paths with the OS-sepcific directory separator character.
         PathUtils.ConvertToOSSpecificPathSeparators(configuration);
--- a/src/Microsoft.Sbom.Api/Executors/ILicenseInformationFetcher.cs
+++ b/src/Microsoft.Sbom.Api/Executors/ILicenseInformationFetcher.cs
@@ -17,12 +17,16 @@ public interface ILicenseInformationFetcher
     /// <returns></returns>
     List<string> ConvertComponentsToListForApi(IEnumerable<ScannedComponent> scannedComponents);
 
+    /// <inheritdoc cref="FetchLicenseInformationAsync(List{string}, int)"/>>
+    Task<List<string>> FetchLicenseInformationAsync(List<string> listOfComponentsForApi);
+
     /// <summary>
     /// Calls the ClearlyDefined API to get the license information for the list of components.
     /// </summary>
     /// <param name="listOfComponentsForApi"> A list of strings formatted into a list of strings that can be used to call the batch ClearlyDefined API.</param>
+    /// <param name="timeout">Timeout in seconds to use when making web requests</param>
     /// <returns></returns>
-    Task<List<string>> FetchLicenseInformationAsync(List<string> listOfComponentsForApi);
+    Task<List<string>> FetchLicenseInformationAsync(List<string> listOfComponentsForApi, int timeout);
 
     /// <summary>
     /// Gets the dictionary of licenses that were fetched from the ClearlyDefined API.
--- a/src/Microsoft.Sbom.Api/Executors/ILicenseInformationFetcher.cs
+++ b/src/Microsoft.Sbom.Api/Executors/ILicenseInformationFetcher.cs
@@ -17,16 +17,12 @@ public interface ILicenseInformationFetcher
     /// <returns></returns>
     List<string> ConvertComponentsToListForApi(IEnumerable<ScannedComponent> scannedComponents);
 
-    /// <inheritdoc cref="FetchLicenseInformationAsync(List{string}, int)"/>>
-    Task<List<string>> FetchLicenseInformationAsync(List<string> listOfComponentsForApi);
-
     /// <summary>
-    /// Calls the ClearlyDefined API to get the license information for the list of components.
+    /// Calls the ClearlyDefined API to get the license information for the list of components. Uses a default timeout specified in implementation
     /// </summary>
     /// <param name="listOfComponentsForApi"> A list of strings formatted into a list of strings that can be used to call the batch ClearlyDefined API.</param>
-    /// <param name="timeout">Timeout in seconds to use when making web requests</param>
     /// <returns></returns>
-    Task<List<string>> FetchLicenseInformationAsync(List<string> listOfComponentsForApi, int timeout);
+    Task<List<string>> FetchLicenseInformationAsync(List<string> listOfComponentsForApi);
 
     /// <summary>
     /// Gets the dictionary of licenses that were fetched from the ClearlyDefined API.
--- a/src/Microsoft.Sbom.Api/Executors/ILicenseInformationService2.cs
+++ b/src/Microsoft.Sbom.Api/Executors/ILicenseInformationService2.cs
@@ -6,7 +6,7 @@ using System.Threading.Tasks;
 
 namespace Microsoft.Sbom.Api.Executors;
 
-public interface ILicenseInformationService
+public interface ILicenseInformationService2 : ILicenseInformationService
 {
-    public Task<List<string>> FetchLicenseInformationFromAPI(List<string> listOfComponentsForApi);
+    public Task<List<string>> FetchLicenseInformationFromAPI(List<string> listOfComponentsForApi, int timeoutInSeconds);
 }
--- a/src/Microsoft.Sbom.Api/Executors/LicenseInformationFetcher.cs
+++ b/src/Microsoft.Sbom.Api/Executors/LicenseInformationFetcher.cs
@@ -14,8 +14,9 @@ using Serilog;
 
 namespace Microsoft.Sbom.Api.Executors;
 
-public class LicenseInformationFetcher : ILicenseInformationFetcher
+public class LicenseInformationFetcher : ILicenseInformationFetcher2
 {
+    public const int DefaultTimeoutInSeconds = 30;
     private readonly ILogger log;
     private readonly IRecorder recorder;
     private readonly ILicenseInformationService licenseInformationService;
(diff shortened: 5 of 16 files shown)


kidcline1:

src/Microsoft.Sbom.Api/Config/Args/GenerationArgs.cs
```diff
@@ -116,6 +116,15 @@ public class GenerationArgs : GenerationAndValidationCommonArgs
...
     public bool? FetchLicenseInformation { get; set; }
 
+    /// <summary>
+    /// Specifies the timeout in seconds for fetching the license information. Defaults to 30 seconds. Has no effect if
+    /// FetchLicenseInformation (li) argument is false or not provided.
+    /// </summary>
+    [ArgShortcut("lto")]
+    [ArgDescription("Specifies the timeout in seconds for fetching the license information. Defaults to 30 seconds. " +
+        "Has no effect if the FetchLicenseInformation (li) argument is false or not provided. ")]
+    public int? LicenseInformationTimeout { get; set; }
```
DaveTryon, line 126: We should specify units in the name to avoid confusion. Maybe 
```suggestion
    public int? LicenseInformationTimeoutInSeconds { get; set; }
```

src/Microsoft.Sbom.Api/Executors/ILicenseInformationFetcher.cs
```diff
@@ -17,12 +17,16 @@ public interface ILicenseInformationFetcher
...
+    Task<List<string>> FetchLicenseInformationAsync(List<string> listOfComponentsForApi);
+
     /// <summary>
     /// Calls the ClearlyDefined API to get the license information for the list of components.
     /// </summary>
     /// <param name="listOfComponentsForApi"> A list of strings formatted into a list of strings that can be used to call the batch ClearlyDefined API.</param>
+    /// <param name="timeout">Timeout in seconds to use when making web requests</param>
     /// <returns></returns>
-    Task<List<string>> FetchLicenseInformationAsync(List<string> listOfComponentsForApi);
+    Task<List<string>> FetchLicenseInformationAsync(List<string> listOfComponentsForApi, int timeout);
```
DaveTryon, line 29: Another place were we should specify the units in the name
```suggestion
    Task<List<string>> FetchLicenseInformationAsync(List<string> listOfComponentsForApi, int timeoutInSeconds);
```

src/Microsoft.Sbom.Api/Executors/LicenseInformationService.cs
```diff
@@ -30,6 +29,11 @@ public LicenseInformationService(ILogger log, IRecorder recorder, HttpClient htt
     }
 
     public async Task<List<string>> FetchLicenseInformationFromAPI(List<string> listOfComponentsForApi)
+    {
+        return await FetchLicenseInformationFromAPI(listOfComponentsForApi, 30);
```
DaveTryon, line 33: Rather than a magic number, let's use a constant. Something like `DefaultClientTimeoutSeconds`, which could replace the previous `ClientTimeoutSeconds`

src/Microsoft.Sbom.Common/Config/Configuration.cs
```diff
@@ -309,6 +310,14 @@ public ConfigurationSetting<bool> FetchLicenseInformation
         set => fetchLicenseInformation.Value = value;
     }
 
+    /// <inheritdoc cref="IConfiguration.LicenseInformationTimeout" />
+    [DefaultValue(30)]
+    public ConfigurationSetting<int> LicenseInformationTimeout
```
DaveTryon, line 315:
```suggestion
    public ConfigurationSetting<int> LicenseInformationTimeoutInSeconds
```

src/Microsoft.Sbom.Api/Executors/LicenseInformationService.cs
```diff
@@ -30,6 +29,11 @@ public LicenseInformationService(ILogger log, IRecorder recorder, HttpClient htt
...
+    }
+
+    public async Task<List<string>> FetchLicenseInformationFromAPI(List<string> listOfComponentsForApi, int timeout)
```
DaveTryon, line 36:
```suggestion
    public async Task<List<string>> FetchLicenseInformationFromAPI(List<string> listOfComponentsForApi, int timeoutInSeconds)
```

src/Microsoft.Sbom.Common/Config/IConfiguration.cs
```diff
@@ -194,6 +194,12 @@ public interface IConfiguration
     /// </summary>
     ConfigurationSetting<bool> FetchLicenseInformation { get; set; }
 
+    /// <summary>
+    /// Specifies the timeout in seconds for fetching the license information. Defaults to 30 seconds. Has no effect if
+    /// FetchLicenseInformation (li) argument is false or not provided.
+    /// </summary>
+    ConfigurationSetting<int> LicenseInformationTimeout { get; set; }
```
DaveTryon, line 201:
```suggestion
    ConfigurationSetting<int> LicenseInformationTimeoutInSeconds { get; set; }
```

src/Microsoft.Sbom.Common/Config/InputConfiguration.cs
```diff
@@ -141,6 +141,10 @@ public class InputConfiguration : IConfiguration
     [DefaultValue(false)]
     public ConfigurationSetting<bool> FetchLicenseInformation { get; set; }
 
+    /// <inheritdoc cref="IConfiguration.LicenseInformationTimeout" />
+    [DefaultValue(30)]
+    public ConfigurationSetting<int> LicenseInformationTimeout { get; set; }
```
DaveTryon, line 146:
```suggestion
    public ConfigurationSetting<int> LicenseInformationTimeoutInSeconds { get; set; }
```

src/Microsoft.Sbom.Api/Config/Args/GenerationArgs.cs
```diff
@@ -116,6 +116,15 @@ public class GenerationArgs : GenerationAndValidationCommonArgs
     [ArgDescription("If set to true, we will attempt to fetch license information of packages detected in the SBOM from the ClearlyDefinedApi.")]
     public bool? FetchLicenseInformation { get; set; }
 
+    /// <summary>
+    /// Specifies the timeout in seconds for fetching the license information. Defaults to 30 seconds. Has no effect if
+    /// FetchLicenseInformation (li) argument is false or not provided.
+    /// </summary>
+    [ArgShortcut("lto")]
+    [ArgDescription("Specifies the timeout in seconds for fetching the license information. Defaults to 30 seconds. " +
```
DaveTryon, line 124: Are any checks needed here?
- min value?
- max value?
- what if it's set when `FetchLicenseInformation` is false?

The parameter should be documented in both `docs\sbom-tool-arguments.md` and `src\Microsoft.Sbom.Targets\README.md`

src/Microsoft.Sbom.Api/Executors/ILicenseInformationFetcher.cs
```diff
@@ -17,12 +17,16 @@ public interface ILicenseInformationFetcher
     /// <returns></returns>
     List<string> ConvertComponentsToListForApi(IEnumerable<ScannedComponent> scannedComponents);
 
+    /// <inheritdoc cref="FetchLicenseInformationAsync(List{string}, int)"/>>
+    Task<List<string>> FetchLicenseInformationAsync(List<string> listOfComponentsForApi);
```
DaveTryon, line 21: Good choice keeping the old method intact--is there a way to specify that we're using a default timeout here?

src/Microsoft.Sbom.Api/Config/Args/GenerationArgs.cs
```diff
@@ -116,6 +116,15 @@ public class GenerationArgs : GenerationAndValidationCommonArgs
     [ArgDescription("If set to true, we will attempt to fetch license information of packages detected in the SBOM from the ClearlyDefinedApi.")]
     public bool? FetchLicenseInformation { get; set; }
 
+    /// <summary>
```
DaveTryon, line 119: A new parameter should be documented in both `docs\sbom-tool-arguments.md` and `src\Microsoft.Sbom.Targets\README.md`

DaveTryon:
Strictly speaking, this is a breaking change because it changes a public interface; it would break any consumers of the API who had their own implementation of the `ILicenseInformationFetcher` interface. The non-breaking way to make this change is to leave `ILicenseInformationFetcher` unchanged, and to create a new interface:
```
public interface ILicenseInformationFetcher2 : ILicenseInformationFetcher
{
    /// <summary>
    /// Calls the ClearlyDefined API to get the license information for the list of components.
    /// </summary>
    /// <param name="listOfComponentsForApi"> A list of strings formatted into a list of strings that can be used to call the batch ClearlyDefined API.</param>
    /// <param name="timeoutInSeconds">Timeout in seconds to use when making web requests</param>
    /// <returns></returns>
    Task<List<string>> FetchLicenseInformationAsync(List<string> listOfComponentsForApi, int timeoutinSeconds);
}
```

Then you derive `LicenseInformationFetcher` from `ILicenseInformationFetcher2`. In `ComponentDetectionBaseWalker.cs`, you do something like this:
```
ILicenseInformationFetcher2 informationFetcher2 = licenseInformationFetcher as ILIcenseInformationFetcher2;
if (informationFetcher2 is null || configuration.LicenseInformationTimeout is null)
{
    apiResponses = await licenseInformationFetcher.FetchLicenseInformationAsync(listOfComponentsForApi);
}
else
{
    apiResponses = await informationFetcher2.FetchLicenseInformationAsync(listOfComponentsForApi, configuration.LicenseInformationTimeout.Value);
}
```

src/Microsoft.Sbom.Api/Config/Args/GenerationArgs.cs
kidcline1: I agree, I've updated all the timeout variable names to specify InSeconds

src/Microsoft.Sbom.Api/Executors/LicenseInformationService.cs
kidcline1: I agree, I've updated LicenseInformationService and LicenseInformationFetcher to use a constant instead.

src/Microsoft.Sbom.Api/Config/Args/GenerationArgs.cs
kidcline1: Thanks for the catch. The valid range is from `int.MinValue` to `int.MaxValue`. Negative values will be treated as InfiniteTimespan in the `LicenseInformationService`. When -lto is specified and `FetchLicenseInformation` is false, it simply has no effect. However, I did add a warning message for this case stating this.

I added test methods for sanitization of Min, Max, and null cases. I also tried specifying a value higher than `int.MaxValue` via command line argument and the program specified a warning message that the value exceeds the integer range, so we are good there as well (not sure how to automate that test).

src/Microsoft.Sbom.Api/Executors/ILicenseInformationFetcher.cs
kidcline1: Thanks. I updated the method docs to say that a default timeout is used

src/Microsoft.Sbom.Api/Config/Args/GenerationArgs.cs
kidcline1: Done in the new commit.

kidcline1:
> Strictly speaking, this is a breaking change because it changes a public interface; it would break any consumers of the API who had their own implementation of the `ILicenseInformationFetcher` interface. The non-breaking way to make this change is to leave `ILicenseInformationFetcher` unchanged, and to create a new interface:
> 
> ```
> public interface ILicenseInformationFetcher2 : ILicenseInformationFetcher
> {
>     /// <summary>
>     /// Calls the ClearlyDefined API to get the license information for the list of components.
>     /// </summary>
>     /// <param name="listOfComponentsForApi"> A list of strings formatted into a list of strings that can be used to call the batch ClearlyDefined API.</param>
>     /// <param name="timeoutInSeconds">Timeout in seconds to use when making web requests</param>
>     /// <returns></returns>
>     Task<List<string>> FetchLicenseInformationAsync(List<string> listOfComponentsForApi, int timeoutinSeconds);
> }
> ```
> 
> Then you derive `LicenseInformationFetcher` from `ILicenseInformationFetcher2`. In `ComponentDetectionBaseWalker.cs`, you do something like this to safely call the method:
> 
> ```
> ILicenseInformationFetcher2 licenseInformationFetcher2 = licenseInformationFetcher as ILicenseInformationFetcher2;
> // Maybe spit out a warning message if licenseInformationFetcher2 is null but configuration.LicenseInformationTimeout is not null
> if (licenseInformationFetcher2 is null || configuration.LicenseInformationTimeout is null)
> {
>     apiResponses = await licenseInformationFetcher.FetchLicenseInformationAsync(listOfComponentsForApi);
> }
> else
> {
>     apiResponses = await licenseInformationFetcher2.FetchLicenseInformationAsync(listOfComponentsForApi, configuration.LicenseInformationTimeout.Value);
> }
> ```

I was wondering how best to handle that. Thank you for the suggestion - I have implemented the two new interfaces `ILicenseInformationFetcher2` and `ILicenseInformationService2` as you suggested to prevent breaking changes to the API.

kidcline1:
> Left several comments

Hello @DaveTryon, thank you for the thorough code review. I have made the requested changes in the latest commit.

src/Microsoft.Sbom.Api/Executors/LicenseInformationService.cs
```diff
@@ -30,6 +30,11 @@ public LicenseInformationService(ILogger log, IRecorder recorder, HttpClient htt
     }
 
     public async Task<List<string>> FetchLicenseInformationFromAPI(List<string> listOfComponentsForApi)
+    {
+        return await FetchLicenseInformationFromAPI(listOfComponentsForApi, DefaultTimeoutInSeconds);
```
DaveTryon, line 34: The default is already implemented in the version without a timeout, so there's no need for `DefaultTimeoutInSeconds`
```suggestion
        return await FetchLicenseInformationFromAPI(listOfComponentsForApi);
```

src/Microsoft.Sbom.Api/Executors/LicenseInformationService.cs
```diff
@@ -15,12 +15,12 @@
 
 namespace Microsoft.Sbom.Api.Executors;
 
-public class LicenseInformationService : ILicenseInformationService
+public class LicenseInformationService : ILicenseInformationService2
 {
+    private const int DefaultTimeoutInSeconds = LicenseInformationFetcher.DefaultTimeoutInSeconds;
```
DaveTryon, line 20: This can just go away
```suggestion
```

src/Microsoft.Sbom.Api/Config/Args/GenerationArgs.cs
DaveTryon: An infinite timeout seems unreasonably large. If actually applied, it would be a timeout of over 1600 years, which is probably not what you want in a build pipeline. Let's go with a smaller range--a max of 86400 (24 hours) seems more than adequate. If our supported range is 1-86400, then any other value should probably be a warning that we're ignoring their value and using the default, or an error that their value is out of range.

src/Microsoft.Sbom.Api/Executors/LicenseInformationFetcher.cs
```diff
@@ -14,8 +14,9 @@
 
 namespace Microsoft.Sbom.Api.Executors;
 
-public class LicenseInformationFetcher : ILicenseInformationFetcher
+public class LicenseInformationFetcher : ILicenseInformationFetcher2
 {
+    public const int DefaultTimeoutInSeconds = 30;
```
DaveTryon, line 19: This might be better in the `Microsoft.Sbom.Common.Constants` class. It puts it with other constants and avoids coupling to concrete execution classes

src/Microsoft.Sbom.Api/Executors/LicenseInformationService.cs
```diff
@@ -38,7 +43,14 @@ public async Task<List<string>> FetchLicenseInformationFromAPI(List<string> list
...
         httpClient.DefaultRequestHeaders.Accept.Add(new MediaTypeWithQualityHeaderValue("application/json"));
-        httpClient.Timeout = TimeSpan.FromSeconds(ClientTimeoutSeconds);
+        if (timeoutInSeconds >= 0)
+        {
+            httpClient.Timeout = TimeSpan.FromSeconds(timeoutInSeconds);
+        }
+        else
+        {
+            httpClient.Timeout = System.Threading.Timeout.InfiniteTimeSpan;
+        }
```
DaveTryon, line 53:
```suggestion
        if (timeoutInSeconds > 0)
        {
            httpClient.Timeout = TimeSpan.FromSeconds(timeoutInSeconds);
        }
```

src/Microsoft.Sbom.Common/Config/Configuration.cs
```diff
@@ -309,6 +310,14 @@ public ConfigurationSetting<bool> FetchLicenseInformation
         set => fetchLicenseInformation.Value = value;
     }
 
+    /// <inheritdoc cref="IConfiguration.LicenseInformationTimeoutInSeconds" />
+    [DefaultValue(30)]
```
DaveTryon, line 314: Would be ideal to use a the const value instead of a magic number

src/Microsoft.Sbom.Api/Config/ConfigSanitizer.cs
```diff
@@ -81,6 +82,20 @@ public IConfiguration SanitizeConfig(IConfiguration configuration)
...
+        {
+            configuration.LicenseInformationTimeoutInSeconds = new(LicenseInformationFetcher.DefaultTimeoutInSeconds, SettingSource.Default);
+        }
+
+        // Check if arg -lto is specified but -li is not
+        if (configuration.FetchLicenseInformation?.Value != true && !configuration.LicenseInformationTimeoutInSeconds.IsDefaultSource)
+        {
+            logger.Warning("A license fetching timeout is specified (argument -lto), but this has no effect when FetchLicenseInfo is unspecified or false (argument -li)");
+        }
+
```
DaveTryon, line 98: This where we should enforce the limits that specify our allowed range of 1 to 86400, with unit tests to exercise the code paths

test/Microsoft.Sbom.Api.Tests/Config/ConfigSanitizerTests.cs
```diff
@@ -351,4 +352,30 @@ public void ConfigSantizer_Validate_ReplacesBackslashes_Linux(ManifestToolAction
             Assert.IsTrue(config.TelemetryFilePath.Value.StartsWith($"/{nameof(config.TelemetryFilePath)}/", StringComparison.Ordinal));
         }
     }
+
+    [TestMethod]
```
DaveTryon, line 356: You may need to break this down into separate tests to test the allowed range of values

DaveTryon (changes requested):
Thanks for the changes--it's very close. I think we need to be more explicit/restrictive in the allowed range of timeout values, make sure that the docs and tests cover the range, and we'll be ready to go

src/Microsoft.Sbom.Api/Executors/LicenseInformationService.cs
kidcline1: Hello Dave. I'm sorry, I don't follow on this one. If we remove `DefaultTimeoutInSeconds` from this method call as suggested, it just recursively calls itself until a Stack Overflow occurs. Did you perhaps mean modifying the call in `LicenseInformationFetcher.FetchLicenseInformationAsync(List<string> listOfComponentsForApi)` instead to call `LicenseInformationService.FetchLicenseInformationFromAPI(List<string> listOfComponentsForApi)` instead of `LicenseInformationFetcher.FetchLicenseInformationAsync(List<string> listOfComponentsForApi, int timeoutInSeconds)`?


```
// LicenseInformationFetcher.cs
public async Task<List<string>> FetchLicenseInformationAsync(List<string> listOfComponentsForApi)
{
    return await FetchLicenseInformationAsync(listOfComponentsForApi, DefaultTimeoutInSeconds);
}
```
to this?
```
// LicenseInformationFetcher.cs
public async Task<List<string>> FetchLicenseInformationAsync(List<string> listOfComponentsForApi)
{
    return await licenseInformationService.FetchLicenseInformationFromAPI(listOfComponentsForApi);
}
```

kidcline1:
> Thanks for the changes--it's very close. I think we need to be more explicit/restrictive in the allowed range of timeout values, make sure that the docs and tests cover the range, and we'll be ready to go

Thanks. I've made the changes requested. For the change I commented on, I made the change based on what I think you were getting at since `LicenseInformationFetcher` and `LicenseInformationService` are so similar.

kidcline1:
Hello @DaveTryon, I've made the changes requested. Is there anything else I need to do on my end?

DaveTryon (approved):
Thanks, @kidcline1! I was busy on other tasks and lost track of this. These changes look really good to me. Assuming everything goes smoothly, we should get this merged in the next day or two. Our entire organization is currently operating under a release freeze for the holidays, so we won't be able to release the new version until early January.

src/Microsoft.Sbom.Api/Executors/ILicenseInformationService2.cs
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace Microsoft.Sbom.Api.Executors;
+
+public interface ILicenseInformationService2 : ILicenseInformationService
```
sfoslund, line 9: I don't love the name ILicenseInformationService2  here- can we have something more descriptive? Maybe ILicenseInformationTimeoutableService?

sfoslund: Alternatively- we could just replace the old ILicenseInformationService and bump the API version by a minor version instead of our usual patch increase. I would prefer that approach but open to opinions @DaveTryon

DaveTryon: >I don't love the name ILicenseInformationService2 here- can we have something more descriptive? Maybe ILicenseInformationTimeoutableService?

This is a well-established pattern (in the Windows world, at least). You're right that the description is lacking, but trying to be more descriptive leads to weirdness in later iterations. You end up with ILicenseInformationServiceWithX, then ILicenseInformationServiceWithXandY, then ILicenseInformationServiceWithXandYandZ.

> Alternatively- we could just replace the old ILicenseInformationService and bump the API version by a minor version instead of our usual patch increase.

If someone has their own implementation of `ILicenseInformationService`, then changing the interface in any way would be a breaking change to them. As such, we would need to bump the major version. Extending the interface gives us the functionality without breaking any existing customers. Since this is new functionality, bumping the minor version is probably appropriate.

sfoslund: Gotcha, that's fine with me

sfoslund (approved):

Issue #801, opened 2024-11-15
[Microsoft.Sbom.Targets] Use fully qualified MSBuild task name for Unzip task in targets file


bstadick:
The Unzip task used in [Microsoft.Sbom.Targets.targets:50](https://github.com/microsoft/sbom-tool/blob/4c2fc7780f6d866ddf4575dbeb1ac4f4db1ffc8e/src/Microsoft.Sbom.Targets/Microsoft.Sbom.Targets.targets#L50) can conflict with other tasks  also named Unzip from other MSBuild task packages such as MSBuildCommunityTasks. To avoid such conflicts it would be best to use the fully qualified task name `Microsoft.Build.Tasks.Unzip` for the Unzip task.

Suggestion:
`<Microsoft.Build.Tasks.Unzip DestinationFolder="$(NugetPackageUnzip)" SourceFiles="$(NugetPackage)" OverwriteReadOnlyFiles="true" ContinueOnError="ErrorAndContinue" />`

sfoslund:
Fixed in https://github.com/microsoft/sbom-tool/pull/803

PR #803, opened 2024-11-15
Use the fully qualified name for the Zip and Unzip tasks

bstadick:
Use the fully qualified task names Microsoft.Build.Tasks.Unzip and Microsoft.Build.Tasks.ZipDirectory to avoid name collisions with other msbuild task packages. Changes to other msbuild tasks in use like RemoveDir are less necessary as they are less likely to have a name conflict. The Unzip and ZipDirectory are newer msbuild tasks so packages are more likely to have made tasks with the same name in lieu of them not being originally present in the core msbuild tasks.

Resolution to issue #801.

Tested task target changes locally with use of MSBuildCommunityTasks and successfully generated nuget packages that contained the SBOM file.


baronfel:
This is a very reasonable thing to do when you have very generic Task names you're dealing with - this seems like a good correctness fix to me.

sfoslund (approved):

Issue #802, opened 2024-11-15
PIP Comp Detect Adapter - No License or Supplier information


samuelc7:
The scan manifest file that Component Detection creates shows license and author information of each pip package that was detected. The sbom-tool does not correctly collect that information - it shows "NO ASSERTION" for the license declared and "NO ASSERTION" for the supplier:

ScanManifest-2024....json (Comp Detect file)

```
"component": {
        "name": "torch",
        "version": "2.5.1",
        "author": "PyTorch Team",
        "license": "BSD-3-Clause",
        "type": "Pip",
        "id": "torch 2.5.1 - pip",
        "packageUrl": {
          "Scheme": "pkg",
          "Type": "pypi",
          "Namespace": null,
          "Name": "torch",
          "Version": "2.5.1",
          "Qualifiers": null,
          "Subpath": null
        }
      }
```

manifest.spdx.json (sbom-tool)
```
{
      "name": "torch",
      "SPDXID": "SPDXRef-Package-A40751563D79B440AD6920A1E59C425463CBCF0A7835485457E3BAFBD06B4260",
      "downloadLocation": "NOASSERTION",
      "filesAnalyzed": false,
      "licenseConcluded": "NOASSERTION",
      "licenseDeclared": "NOASSERTION",
      "copyrightText": "NOASSERTION",
      "versionInfo": "2.5.1",
      "externalRefs": [
        {
          "referenceCategory": "PACKAGE-MANAGER",
          "referenceType": "purl",
          "referenceLocator": "pkg:pypi/torch@2.5.1"
        }
      ],
      "supplier": "NOASSERTION"
    },
```

THEJRRR:
We don't have a regulatory or internal use case for this yet, but we would like to get to it when time allows.

WimLeflere:
[The Minimum Elements For a Software Bill of Materials (NTIA)](https://www.ntia.gov/files/ntia/publications/sbom_minimum_elements_report.pdf) specifies the 'Supplier' field as a minimum (required) element.

PR #804, opened 2024-11-15
Modified pip extension - get license info from pipcomponent and added…

samuelc7:
… supplier.

For #802 

Noticed that the `component.LicenseConcluded` never had the correct license info from the Component Detection scan manifest files but the `pipComponent` obj did in the `pipComponent.License` attribute.

Added the Supplier as the pipComponent Author. SPDX 2.2 requires the Supplier to be one of:
`Person: ...`
`Organization: ...`

but I think it's more important to capture the author as the supplier even if Person/Org isn't prepended to it. I will remove the supplier if the community does not agree with this.


sfoslund:

PR #830, opened 2024-12-11
Defining and generating spdx 3.0 json elements

pragnya17:
With the completion of this feature, we will be able to generate spdx 3.0 json elements which make up an SBOM document.
 The goal is to generate spdx 3.0 elements in the format that is specified by the 3.0 spec - [SPDX Specification 3.0.1](https://spdx.github.io/spdx-spec/v3.0.1/) 

Specifically these are the changes that are introduced in this PR:
1. Entities - Define the different spdx 3.0 elements as classes under "Entities". Instances of these classes are used to represent spdx 3.0 json elements such as files, packages, etc.
2. Generator - Generate a json document using the spdx 3.0 elements instead of the spdx 2.0 elements. 
3. Unit testing for the generation of all these spdx 3.0 elements


src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Entities/AnyLicenseInfo.cs
```diff
@@ -0,0 +1,19 @@
...
+namespace Microsoft.Sbom.Parsers.Spdx30SbomParser.Entities;
+
+/// <summary>
+/// Class defintion is based on: https://spdx.github.io/spdx-spec/v3.0.1/model/SimpleLicensing/Classes/AnyLicenseInfo/
+/// </summary>
+public class AnyLicenseInfo : Element
+{
+    public AnyLicenseInfo()
+    {
+        Type = "AnyLicenseInfo";
```
alisonlomaka, line 16: Could use nameof() instead of hard-coding?

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Entities/Enums/RelationshipType.cs
```diff
@@ -0,0 +1,309 @@
...
+    /// </summary>
+    HAS_ASSOCIATED_VULNERABILITY,
+
+    /// <summary>
+    /// The from SoftwareArtifact is concluded by the SPDX data creator to be governed by each to license.
+    /// </summary>
+    HAS_CONCLUDED_LICENSE,
+
+    /// <summary>
+    /// The from Element treats each to Element as a data file. A data file is an artifact that stores data required or optional for the from Element's functionality. A data file can be a database file, an index file, a log file, an AI model file, a calibration data file, a temporary file, a backup file, and more. For AI training dataset, test dataset, test artifact, configuration data, build input data, and build output data, please consider using the more specific relationship types: trainedOn, testedOn, hasTest, configures, hasInput, and hasOutput, respectively. This relationship does not imply dependency.
```
DaveTryon, line 131: Can we split this comment into multiple lines to make it easier to read without have to scroll multiple screens to the right?

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Entities/Spdx30Relationship.cs
```diff
@@ -0,0 +1,54 @@
...
+    public Spdx30Relationship()
+    {
+        Type = "Relationship";
+    }
+
+    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
+    [JsonPropertyName("completeness")]
+    public object Completeness { get; set; }
+
+    /*[JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingDefault)]
```
DaveTryon, line 29: ?

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Entities/Spdx30Relationship.cs
```diff
@@ -0,0 +1,54 @@
...
+    /// </summary>
+    [JsonRequired]
+    [JsonPropertyName("from")]
+    public string From { get; set; }
+
+    [JsonRequired]
+    [JsonPropertyName("relationshipType")]
+    public RelationshipType RelationshipType { get; set; }
+
+    /*[JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingDefault)]
```
DaveTryon, line 44: Seems like `startTime` should be enabled or removed--not sure which

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Utils/ElementSerializer.cs
```diff
@@ -0,0 +1,70 @@
...
+    public override void Write(Utf8JsonWriter writer, List<Element> elements, JsonSerializerOptions options)
+    {
+        writer.WriteStartArray();
+
+        foreach (var element in elements)
+        {
+            switch (element.Type)
+            {
+                case "AnyLicenseInfo":
+                    JsonSerializer.Serialize(writer, (AnyLicenseInfo)element, options);
```
DaveTryon, line 27: Serialize has a templated version that might be slightly cleaner--note that this pattern would apply to all of these:
```suggestion
                    JsonSerializer.Serialize<AnyLicenseInfo>(writer, element at AnyLicenseInfo, options);
```

Part of me wants to see us use a visitor pattern here instead of a switch statement, but that may not be worth the cost

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Utils/InternalMetadataProviderIdentityExtensions.cs
```diff
@@ -0,0 +1,190 @@
...
+    {
+        if (internalMetadataProvider is null)
+        {
+            throw new ArgumentNullException(nameof(internalMetadataProvider));
+        }
+
+        // First check if the user provided a package version.
+        if (internalMetadataProvider.TryGetMetadata(MetadataKey.PackageSupplier, out object packageSupplier))
+        {
+            return packageSupplier as string;
```
DaveTryon, line 136: What's the type of `packageSupplier`? If it's always a string, then we can strongly type it in the call to `TryGetMetadata`. If it's not a string, then the cast will fail and we'll return `null`. Is that ok?

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Utils/InternalMetadataProviderIdentityExtensions.cs
```diff
@@ -0,0 +1,190 @@
...
+    public static string GetGenerationTimestamp(this IInternalMetadataProvider internalMetadataProvider)
+    {
+        if (internalMetadataProvider is null)
+        {
+            throw new ArgumentNullException(nameof(internalMetadataProvider));
+        }
+
+        if (internalMetadataProvider.TryGetMetadata(MetadataKey.GenerationTimestamp, out object generationTimestamp))
+        {
+            return generationTimestamp as string;
```
DaveTryon, line 185: What's the type of `generationTimestamp`? If it's always a string, then we can strongly type it in the call to `TryGetMetadata`. If it's not a string, then the cast will fail and we'll return `null`. Is that ok?

test/Microsoft.Sbom.Parsers.Spdx30SbomParser.Tests/Parser/GeneratorTests.cs
```diff
@@ -0,0 +1,233 @@
...
+
+[TestClass]
+public class GeneratorTests
+{
+    private readonly Generator generator = new Generator();
+    private readonly Mock<IRecorder> recorderMock = new Mock<IRecorder>();
+    private readonly Mock<ILogger> mockLogger = new Mock<ILogger>();
+    private readonly Mock<IFileSystemUtils> fileSystemMock = new Mock<IFileSystemUtils>();
+    private readonly Mock<IManifestConfigHandler> mockConfigHandler = new Mock<IManifestConfigHandler>();
+    //private readonly Mock<IMetadataProvider> mockMetadataProvider = new Mock<IMetadataProvider>();
```
DaveTryon, line 33: Should the commented mocks go away?

test/Microsoft.Sbom.Parsers.Spdx30SbomParser.Tests/Parser/GeneratorTests.cs
```diff
@@ -0,0 +1,233 @@
...
+
+        // Replace placeholders with appropriate regex patterns
+        pattern = pattern.Replace(@"\.\*", ".*");
+
+        return pattern;
+    }
+
+    private List<Checksum> GetSampleChecksums()
+    {
+        var checksumSha1 = new Checksum
```
DaveTryon, line 175: VS might suggest that you inline this:
```
return new List<Checksum>
{
  new Checksum
  {
    Algorithm = AlgorithmName.SHA1,
    ChecksumValue = "sha1Value"
  },  new Checksum
  {
    Algorithm = AlgorithmName.SHA256,
    ChecksumValue = "sha256Value"
  },
};

File: test/Microsoft.Sbom.Parsers.Spdx30SbomParser.Tests/Parser/GeneratorTests.cs [hunk 10]
```diff
@@ -0,0 +1,233 @@
...
+        var config = new Configuration
+        {
+            PackageName = new ConfigurationSetting<string>("the-package-name"),
+            PackageVersion = new ConfigurationSetting<string>("the-package-version"),
+            NamespaceUriUniquePart = new ConfigurationSetting<string>("some-custom-value-here"),
+            NamespaceUriBase = new ConfigurationSetting<string>("http://sbom.microsoft"),
+            PackageSupplier = new ConfigurationSetting<string>("the-package-supplier"),
+        };
+
+        var logger = new Mock<ILogger>();
```
DaveTryon, line 208: I have a lot of opinions on unit tests--let me capture some ideas:
1. Test fixtures--including the mocks--are good. They make each specific test shorter and focus on just what is happening. Your `[TestInitialize]` method builds placeholder fixtures, then each test sets just what it cares about
2. Creating mocks the `MockBehavior.Strong` is good. It ensures that your tests describe what methods should be called on the mock
3. Using the mock's `VerifyAll` method in a `[TestCleanup]` function provides a clean way to ensure that every method with a `Mock.Setup` call was actually called

test/Microsoft.Sbom.Parsers.Spdx30SbomParser.Tests/Parser/JsonStrings/SbomDocCreationJsonStrings.cs
```diff
@@ -0,0 +1,60 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Sbom.Parser.JsonStrings;
+
+public static class SbomDocCreationJsonStrings
+{
+    [System.Diagnostics.CodeAnalysis.SuppressMessage("Style", "JSON002:Probable JSON string detected", Justification = "Need to use JSON string")]
+    public const string DocCreationJsonString =
+    @"[
```
DaveTryon, line 10: Pretty printing of these files is consistent. Some have different indents or no indents at al. If these are read-only test resources, it might be simpler to save them as JSON files that the tests read, or let the compiler embed them as resources, then access them via a helper method.

test/Microsoft.Sbom.Parsers.Spdx30SbomParser.Tests/Utils/SPDXExtensionsTest.cs
```diff
@@ -0,0 +1,160 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+/*
```
DaveTryon, line 4: Code in this file is all commented out... not sure what that means

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Constants.cs
```diff
@@ -0,0 +1,54 @@
...
+using System.Collections.Generic;
+using Microsoft.Sbom.Extensions.Entities;
+
+namespace Microsoft.Sbom.Parsers.Spdx30SbomParser;
+
+internal static class Constants
+{
+    internal const string SPDXName = "SPDX";
+    internal const string SPDXVersion = "3.0";
+    internal const string DataLicenceValue = "CC0-1.0";
```
DaveTryon, line 13: Not used?

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Utils/ElementSerializer.cs
```diff
@@ -0,0 +1,70 @@
...
+        throw new NotImplementedException("Deserialization of Elements into specific subtypes is not implemented yet.");
+    }
+
+    public override void Write(Utf8JsonWriter writer, List<Element> elements, JsonSerializerOptions options)
+    {
+        writer.WriteStartArray();
+
+        foreach (var element in elements)
+        {
+            switch (element.Type)
```
alisonlomaka, line 24: Haven't tried this, but the internet tells me that it is possible to switch on type - pseudocode:

switch(element)
{
case AnyLicenseInfo:
break;
case CreationInfo:
break;
}

It would make this code less fragile since the case labels could then be compiler verified.

alisonlomaka: Or, should Element expose a Serialize method to be overridden by inheriting classes? This code smells a bit like a problem for polymorphism.

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Utils/InternalMetadataProviderIdentityExtensions.cs
```diff
@@ -0,0 +1,190 @@
+// Copyright (c) Microsoft. All rights reserved.
```
alisonlomaka, line 1: Is this identical to the Spdx22SbomParser copy of the file? Can we move it to a common location and pull from there instead of duplicating?

alisonlomaka: Same question for SPDXExtensions.cs.

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Utils/InternalMetadataProviderIdentityExtensions.cs
pragnya17: Yes both of these calls you commented on can be strongly typed since they should always be strings. Other calls in the old 2.2 generator left it as an object type, which is why I retained it here. But I agree strong typing makes more sense here.

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Entities/Spdx30Relationship.cs
pragnya17: Both start and end time should be enabled, I accidentally had it commented out.

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Generator.cs
```diff
@@ -0,0 +1,615 @@
...
+{
+    private static readonly Dictionary<AlgorithmName, HashAlgorithm> AlgorithmMap = new()
+    {
+        { AlgorithmName.SHA1, HashAlgorithm.sha1 },
+        { AlgorithmName.SHA256, HashAlgorithm.sha256 },
+        { AlgorithmName.SHA512, HashAlgorithm.sha512 },
+        { AlgorithmName.MD5, HashAlgorithm.md5 }
+    };
+
+    public AlgorithmName[] RequiredHashAlgorithms => new[] { AlgorithmName.SHA256, AlgorithmName.SHA1 };
```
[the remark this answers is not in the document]
pragnya17: In 2.2, we for some reason required sha1 and sha256. Last I talked to John Wade about it he mentioned that we might just want to keep the same behavior for 3.0 (at least for now). I'll double check with him again though.

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Generator.cs
```diff
@@ -0,0 +1,615 @@
...
+    // any of the required hashes
+    private void EnsureRequiredHashesPresent(Checksum[] fileHashes)
+    {
+        foreach (var hashAlgorithmName in from hashAlgorithmName in RequiredHashAlgorithms
+                                          where !fileHashes.Select(fh => fh.Algorithm).Contains(hashAlgorithmName)
+                                          select hashAlgorithmName)
+        {
+            throw new MissingHashValueException($"The hash value for algorithm {hashAlgorithmName} is missing from {nameof(fileHashes)}");
+        }
+    }
```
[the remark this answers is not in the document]
pragnya17: I moved these methods to a new class. I can't move the required hashes array since it needs to be implemented as part of the IManifestGenerator in the generator files

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Generator.cs
```diff
@@ -0,0 +1,615 @@
...
+            Name = toolName,
+        };
+
+        spdxOrganization.AddSpdxId();
+        spdxTool.AddSpdxId();
+
+        var spdxCreationInfo = new CreationInfo
+        {
+            Id = "_:creationinfo",
+            SpecVersion = Constants.SPDXVersion,
```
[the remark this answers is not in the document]
pragnya17: In the 2.2 Generator it is also defined as just major.minor - https://github.com/microsoft/sbom-tool/blob/main/src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Constants.cs. So I think this should be fine here as well

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Utils/InternalMetadataProviderIdentityExtensions.cs
pragnya17: I moved all but one of these methods (which is not shared by the generators) into shared file. For spdxExtensions all the methods are different so I kept them as is

test/Microsoft.Sbom.Parsers.Spdx30SbomParser.Tests/Parser/JsonStrings/SbomDocCreationJsonStrings.cs
pragnya17: The reason I chose this method was because we have to convert the json to a string anyways in order to Assert on the string equality + this way I can use regular expressions for things that might differ from run to run like creation timestamp. This is also how the unit tests are written for the Spdx22Sbom methods. 

I can format the json strings better for consistenct. But considering this, do we still want to save them as Json files instead of keeping them as strings?

test/Microsoft.Sbom.Parsers.Spdx30SbomParser.Tests/Parser/GeneratorTests.cs
pragnya17: I don't think 1. applies to the unit tests but 2 and 3 are great points - thanks for pointing them out and I've included those changes now

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Constants.cs
```diff
@@ -0,0 +1,38 @@
...
```
DaveTryon, line 13: It looks like the following are defined but not used--not sure what that means, though...
- `DataLicenceValue`
- `SPDXDocumentIdValue`
- `RootPackageIdValue`
- `DataLicenseHeaderName`
- `SPDXDocumentNameFormatString`

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Entities/Element.cs
```diff
@@ -0,0 +1,70 @@
...
+using System.Collections.Generic;
+using System.Text.Json.Serialization;
+
+namespace Microsoft.Sbom.Parsers.Spdx30SbomParser.Entities;
+
+/// <summary>
+/// Base domain class from which all other SPDX-3.0 domain classes derive.
+/// https://spdx.github.io/spdx-spec/v3.0.1/model/Core/Classes/Element/
+/// </summary>
+public class Element
```
DaveTryon, line 14: Should this be abstract? I'm not sure if we ever have just an `Element` that isn't a derived object.

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Entities/AnyLicenseInfo.cs
```diff
@@ -0,0 +1,19 @@
...
+namespace Microsoft.Sbom.Parsers.Spdx30SbomParser.Entities;
+
+/// <summary>
+/// Class defintion is based on: https://spdx.github.io/spdx-spec/v3.0.1/model/SimpleLicensing/Classes/AnyLicenseInfo/
+/// </summary>
+public class AnyLicenseInfo : Element
+{
+    public AnyLicenseInfo()
+    {
+        Type = nameof(AnyLicenseInfo);
```
DaveTryon, line 16: This is way better than hardcoded strings, but we might be able to do even better. Since all of the types derive from `Element`, you could centralize the code that sets the `Type` property in the constructor of the `Element` class:

`Type = GetType().Name;`

For derived classes, this will provide the name of the derived class, not the base class.

Looking at the code, I think the `NoneElement` is the only class that sets its type to a value that doesn't match the class. In that case. you can continue to set it in the constructor for `NoneElement`, since the base class constructor is called before the derived class constructor and the property will just get overwritten

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Utils/ElementSerializer.cs
DaveTryon: +1 for polymorphism here if we can make it work without too much trouble. The current version will break if a class gets renamed and the compiler doesn't know that the class name and the switch statement are somehow linked. The safe non-polymorphism route is to move all of the Type values into a constants class, then use those constants in both the class constructors and in the switch. That's a bit ugly :(

alisonlomaka:
[heart]         Alison Lomaka reacted to your message:
________________________________

alisonlomaka:
[like]  Alison Lomaka reacted to your message:
________________________________

alisonlomaka:
[like]  Alison Lomaka reacted to your message:
________________________________

alisonlomaka:
[like]  Alison Lomaka reacted to your message:
________________________________

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Utils/ElementSerializer.cs
alisonlomaka: No objection to using the as operator instead of casting, but it doesn't really address my primary concern, which is the maintainability of hard-coding the class names as strings.

I notice that JsonSerializer has overloads that allow passing the Type in; would that allow us to write code like this instead?

JsonSerializer.Serialize(writer, element, element.GetType(), options);

That would be a nice elegant solution. Otherwise, but we could implement a virtual method on Element and override it in each derived type; it isn't pretty but with expression-bodied members we can keep it concise and it would eliminate hard-coding as well as keeping type-specific behavior within the type.

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Entities/Element.cs
pragnya17: Yes it can be made abstract, there will be no objects other than derived ones.

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Entities/AnyLicenseInfo.cs
pragnya17: Makes sense, I've made that change except for NoneElement, NoAssertionElement, File, Package, and Spdx30Relationship which follow a different pattern.

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Constants.cs
pragnya17: Everything except DataLicenseHeaderName is used in generator.cs. I will remove DataLicenseHeaderName from Constants.cs

test/Microsoft.Sbom.Parsers.Spdx30SbomParser.Tests/Parser/JsonStrings/SbomDocCreationJsonStrings.cs
DaveTryon: It's hard to argue with existing precedent.

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Utils/ElementSerializer.cs
pragnya17: I ended up using Alison's suggestion - JsonSerializer.Serialize(writer, element, element.GetType(), options); which is much cleaner. Thank you! I'll leave this thread open in case there are any more comments regarding this issue but I think this should solve the previous concerns.

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Utils/ElementSerializer.cs
pragnya17: This is no longer relevant with the new changes.

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Entities/ContentIdentifier.cs
```diff
@@ -17,7 +17,6 @@ public class ContentIdentifier : Software
 
     public ContentIdentifier()
     {
-        Type = nameof(ContentIdentifier);
```
alisonlomaka: Nit; can remove empty constructor?

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Entities/Element.cs
```diff
@@ -11,11 +11,12 @@ namespace Microsoft.Sbom.Parsers.Spdx30SbomParser.Entities;
 /// Base domain class from which all other SPDX-3.0 domain classes derive.
 /// https://spdx.github.io/spdx-spec/v3.0.1/model/Core/Classes/Element/
 /// </summary>
-public class Element
+public abstract class Element
 {
     protected Element()
     {
         CreationInfoDetails = "_:creationinfo";
+        Type = GetType().Name;
```
alisonlomaka, line 19: Love it, nice change. :)

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Utils/ElementSerializer.cs
```diff
@@ -0,0 +1,29 @@
...
+        throw new NotImplementedException("Deserialization of Elements into specific subtypes is not implemented yet.");
+    }
+
+    public override void Write(Utf8JsonWriter writer, List<Element> elements, JsonSerializerOptions options)
+    {
+        writer.WriteStartArray();
+
+        foreach (var element in elements)
+        {
+            JsonSerializer.Serialize(writer, element, element.GetType(), options);
```
DaveTryon, line 24: This is nice!

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Utils/ElementSerializer.cs
```diff
@@ -21,48 +21,7 @@ public override void Write(Utf8JsonWriter writer, List<Element> elements, JsonSe
...
-                    JsonSerializer.Serialize(writer, element as NamespaceMap, options);
-                    break;
-                case "NoAssertionElement":
-                    JsonSerializer.Serialize(writer, element as NoAssertionElement, options);
-                    break;
-                default:
-                    JsonSerializer.Serialize(writer, element, options);
-                    break;
-            }
+            JsonSerializer.Serialize(writer, element, element.GetType(), options);
```
alisonlomaka, line 24: Nice, new code is very concise and expressive!

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Generator.cs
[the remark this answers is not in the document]
pragnya17: Talked to @rob gallo and he confirmed that it was required. Will check with Adrian in the new year to get more context and document why it is required.

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Generator.cs
```diff
@@ -0,0 +1,615 @@
...
+        }
+
+        if (externalDocumentReferenceInfo.Checksum is null)
+        {
+            throw new ArgumentNullException(nameof(externalDocumentReferenceInfo.Checksum));
+        }
+
+        var sha1Hash = externalDocumentReferenceInfo.Checksum.FirstOrDefault(h => h.Algorithm == AlgorithmName.SHA1) ??
+                       throw new MissingHashValueException(
+                           $"The hash value for algorithm {AlgorithmName.SHA1} is missing from {nameof(externalDocumentReferenceInfo)}");
```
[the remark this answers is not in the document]
pragnya17: Talked to Rob and he confirmed that SHA1 was required. Will check with Adrian in the new year to get more context and document why it is required.

alisonlomaka:
[like]  Alison Lomaka reacted to your message:
________________________________

pragnya17:
> @pragnya17 please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

@microsoft-github-policy-service agree [company="{Microsoft}"]

pragnya17:
@microsoft-github-policy-service agree company="Microsoft"

alisonlomaka:
[heart]         Alison Lomaka reacted to your message:
________________________________

PR #848, opened 2024-12-31
Update MSTest.Test* packages from 3.6.4 to 3.7.0, follow best practices

DaveTryon:
Dependabot tried to update `MSTest.TestFramework` from 3.6.4 to 3.7.0 in #839, and `MSTest.TestAdapter` in #841. These PR's both failed, as the 2 pieces need to move together for the tests to work correctly. In addition, the newer version of `MSTest.TestFramework` adds several new warnings, and since our builds treat warnings as errors, the build breaks.

This PR touches a lot of files, but since these packages are all test-specific, so are the code changes. It was done as a series of commits, and reviewing each commit in sequence will probably simplify the review process. Commits are as follows:

Commit | Description/Coment
--- | ---
[1](https://github.com/microsoft/sbom-tool/commit/5512a2f23685422003ae188c73d15f8698e0ea46) | Unchanged from #839 
[2](https://github.com/microsoft/sbom-tool/commit/03b9a4ac629c71c400214be30cf45ee02de5c522) | Suppress new warnings
[3](https://github.com/microsoft/sbom-tool/commit/353b0db5a98e1d3549d8f0698d2ada42c9be81e5) | Unchanged from #841 -- this allowed the tests to run
[4](https://github.com/microsoft/sbom-tool/commit/1ea8e90a4305e53a9f011b4b85b2dfc4e7d2297d) | Added a suppression that was flagged once both packages were updated
[5](https://github.com/microsoft/sbom-tool/commit/b8fcf1f66315877814413285deb3480243e39abf) | [MSTEST0005](https://learn.microsoft.com/en-us/dotnet/core/testing/mstest-analyzers/mstest0005) We were using TestContext improperly, fix it
[6](https://github.com/microsoft/sbom-tool/commit/7275b724652228305f268c5a5e910dd0ce45b9f7) | [MSTEST0006](https://learn.microsoft.com/en-us/dotnet/core/testing/mstest-analyzers/mstest0006) `ExpectedException` is no longer considered a best practice, so I changed them to `ThrowsException` or `ThrowsExceptionAsync`. Most of these changes were tool-driven, but some were manual and exposed some dead code in the unit tests
[7](https://github.com/microsoft/sbom-tool/commit/22517b288625ecc2864bee1241c0c5a221d85703) | [MSTEST0017](https://learn.microsoft.com/en-us/dotnet/core/testing/mstest-analyzers/mstest0017) `Assert.AreEqual` wants the expected parameter first--there were many places in the code where we had the test parameter first. These changes were largely tool-driven
[8](https://github.com/microsoft/sbom-tool/commit/f794cc42fdfa1e1fe92857a18eaa17835bcaace8) | [MSTEST0024](https://learn.microsoft.com/en-us/dotnet/core/testing/mstest-analyzers/mstest0024) was fixed by [5](https://github.com/microsoft/sbom-tool/commit/b8fcf1f66315877814413285deb3480243e39abf), so the suppression can go away
[9](https://github.com/microsoft/sbom-tool/commit/99f1757a67b5995c5748629137281fa89723ee26) | [MSTEST0034](https://learn.microsoft.com/en-us/dotnet/core/testing/mstest-analyzers/mstest0034) This just adds a parameter to an attribute decoration
[10](https://github.com/microsoft/sbom-tool/commit/764c30a0d27ce4b95d49cb1fc759f8f51d2f9f2d) | [MSTEST0037](https://learn.microsoft.com/en-us/dotnet/core/testing/mstest-analyzers/mstest0037) `Assert.AreEqual`, `Assert.IsNull`, and `Assert.IsNotNull` are all preferred over `Assert.IsTrue` becuase they are more descriptive. These changes were largely tool-driven.
[11](https://github.com/microsoft/sbom-tool/commit/ff3f4b0d8d178ac323750e794059f5f2edda5f04) | I noticed a few more places where the parameters of `Assert.AreEqual` were in the wrong order, but the analyzer failed to catch the error. These changes were manual.
[12](https://github.com/microsoft/sbom-tool/commit/8c9248b6bb81fa7e289f0ce0c58a9fb1bf6e16bc) | [MSTEST0026](https://learn.microsoft.com/en-us/dotnet/core/testing/mstest-analyzers/mstest0026) Null propagation in asserts is discouraged
[13](https://github.com/microsoft/sbom-tool/commit/e8b2539e1cb4dea649900c0cc2254ce840e5e1d9) | [MSTEST0032](https://learn.microsoft.com/en-us/dotnet/core/testing/mstest-analyzers/mstest0032) Non-nullable types don't need to be asserted as non-null. Note that this undoes some of the changes from [12](https://github.com/microsoft/sbom-tool/commit/8c9248b6bb81fa7e289f0ce0c58a9fb1bf6e16bc)
[14](https://github.com/microsoft/sbom-tool/commit/ded35baee9f4da3957836ada6fe19754ecfd849b) | [JSON002](https://learn.microsoft.com/en-us/visualstudio/ide/reference/json002?view=vs-2022) is no longer showing up as an error, so I removed the suppression
[15](https://github.com/microsoft/sbom-tool/commit/ad24014ac372ae2d06af2aa6934857778f4af61b) | Remove a couple of using blocks that were accidentally added along the way
[16](https://github.com/microsoft/sbom-tool/commit/dca3b38fd953672fe537af341929742ff1b6c389) | Restore a check was dropped in [6](https://github.com/microsoft/sbom-tool/commit/7275b724652228305f268c5a5e910dd0ce45b9f7)
[17](https://github.com/microsoft/sbom-tool/pull/848/commits/2d93c2d6bea31d14f17773c5d41be63c5b6a1b60) | Fix broken tests on macOS, add local JSON002 (only seems to impact Visual Studio, not command line)

Note: When running from inside Visual Studio, the net472 projects get skipped. They run correctly from the command line, so any potential regressions should get caught in the pipelines. The net80 versions of the tests work as expected in both Visual Studio and from the command line. I opened https://github.com/microsoft/testfx/issues/4487 to report that issue. It was resolved as a dupe of https://github.com/microsoft/testfx/issues/4426

### DIFF
--- a/test/Microsoft.Sbom.Api.Tests/Config/ConfigurationBuilderTestsForGeneration.cs
+++ b/test/Microsoft.Sbom.Api.Tests/Config/ConfigurationBuilderTestsForGeneration.cs
@@ -82,8 +82,7 @@ public class ConfigurationBuilderTestsForGeneration : ConfigurationBuilderTestsB
     }
 
     [TestMethod]
-    [ExpectedException(typeof(ValidationArgException))]
-    public async Task ConfigurationBuilderTest_Generation_BuildDropPathDoNotExist_Throws()
+    public void ConfigurationBuilderTest_Generation_BuildDropPathDoNotExist_Throws()
     {
         var configFileParser = new ConfigFileParser(fileSystemUtilsMock.Object);
         var cb = new ConfigurationBuilder<GenerationArgs>(mapper, configFileParser);
@@ -97,12 +96,11 @@ public class ConfigurationBuilderTestsForGeneration : ConfigurationBuilderTestsB
             PackageSupplier = "Contoso"
         };
 
-        var configuration = await cb.GetConfiguration(args);
+        Assert.ThrowsExceptionAsync<ValidationArgException>(() => cb.GetConfiguration(args));
     }
 
     [TestMethod]
-    [ExpectedException(typeof(AccessDeniedValidationArgException))]
-    public async Task ConfigurationBuilderTest_Generation_BuildDropPathNotWriteAccess_Throws()
+    public void ConfigurationBuilderTest_Generation_BuildDropPathNotWriteAccess_Throws()
     {
         var configFileParser = new ConfigFileParser(fileSystemUtilsMock.Object);
         var cb = new ConfigurationBuilder<GenerationArgs>(mapper, configFileParser);
@@ -118,7 +116,7 @@ public class ConfigurationBuilderTestsForGeneration : ConfigurationBuilderTestsB
             PackageSupplier = "Contoso"
         };
 
-        var configuration = await cb.GetConfiguration(args);
+        Assert.ThrowsExceptionAsync<AccessDeniedValidationArgException>(() => cb.GetConfiguration(args));
     }
 
     [TestMethod]
@@ -277,8 +275,7 @@ public class ConfigurationBuilderTestsForGeneration : ConfigurationBuilderTestsB
     [DataRow("https://")]
     [DataRow("ww.com")]
     [DataRow("https//test.com")]
-    [ExpectedException(typeof(ValidationArgException), "The value of NamespaceUriBase must be a valid URI.")]
-    public async Task ConfigurationBuilderTest_Generation_BadNSBaseUri_Fails(string badNsUri)
+    public void ConfigurationBuilderTest_Generation_BadNSBaseUri_Fails(string badNsUri)
     {
         var configFileParser = new ConfigFileParser(fileSystemUtilsMock.Object);
         var cb = new ConfigurationBuilder<GenerationArgs>(mapper, configFileParser);
@@ -296,6 +293,6 @@ public class ConfigurationBuilderTestsForGeneration : ConfigurationBuilderTestsB
             PackageSupplier = "Contoso"
         };
 
-        var config = await cb.GetConfiguration(args);
+        Assert.ThrowsExceptionAsync<ValidationArgException>(() => cb.GetConfiguration(args));
     }
 }
--- a/test/Microsoft.Sbom.Api.Tests/Config/ConfigurationBuilderTestsForGeneration.cs
+++ b/test/Microsoft.Sbom.Api.Tests/Config/ConfigurationBuilderTestsForGeneration.cs
@@ -82,7 +82,7 @@ public class ConfigurationBuilderTestsForGeneration : ConfigurationBuilderTestsB
     }
 
     [TestMethod]
-    public void ConfigurationBuilderTest_Generation_BuildDropPathDoNotExist_Throws()
+    public async Task ConfigurationBuilderTest_Generation_BuildDropPathDoNotExist_Throws()
     {
         var configFileParser = new ConfigFileParser(fileSystemUtilsMock.Object);
         var cb = new ConfigurationBuilder<GenerationArgs>(mapper, configFileParser);
@@ -96,11 +96,11 @@ public class ConfigurationBuilderTestsForGeneration : ConfigurationBuilderTestsB
             PackageSupplier = "Contoso"
         };
 
-        Assert.ThrowsExceptionAsync<ValidationArgException>(() => cb.GetConfiguration(args));
+        await Assert.ThrowsExceptionAsync<ValidationArgException>(() => cb.GetConfiguration(args));
     }
 
     [TestMethod]
-    public void ConfigurationBuilderTest_Generation_BuildDropPathNotWriteAccess_Throws()
+    public async Task ConfigurationBuilderTest_Generation_BuildDropPathNotWriteAccess_Throws()
     {
         var configFileParser = new ConfigFileParser(fileSystemUtilsMock.Object);
         var cb = new ConfigurationBuilder<GenerationArgs>(mapper, configFileParser);
@@ -116,7 +116,7 @@ public class ConfigurationBuilderTestsForGeneration : ConfigurationBuilderTestsB
             PackageSupplier = "Contoso"
         };
 
-        Assert.ThrowsExceptionAsync<AccessDeniedValidationArgException>(() => cb.GetConfiguration(args));
+        await Assert.ThrowsExceptionAsync<AccessDeniedValidationArgException>(() => cb.GetConfiguration(args));
     }
 
     [TestMethod]
@@ -275,7 +275,7 @@ public class ConfigurationBuilderTestsForGeneration : ConfigurationBuilderTestsB
     [DataRow("https://")]
     [DataRow("ww.com")]
     [DataRow("https//test.com")]
-    public void ConfigurationBuilderTest_Generation_BadNSBaseUri_Fails(string badNsUri)
+    public async Task ConfigurationBuilderTest_Generation_BadNSBaseUri_Fails(string badNsUri)
     {
         var configFileParser = new ConfigFileParser(fileSystemUtilsMock.Object);
         var cb = new ConfigurationBuilder<GenerationArgs>(mapper, configFileParser);
@@ -293,6 +293,6 @@ public class ConfigurationBuilderTestsForGeneration : ConfigurationBuilderTestsB
             PackageSupplier = "Contoso"
         };
 
-        Assert.ThrowsExceptionAsync<ValidationArgException>(() => cb.GetConfiguration(args), "The value of NamespaceUriBase must be a valid URI.");
+        await Assert.ThrowsExceptionAsync<ValidationArgException>(() => cb.GetConfiguration(args), "The value of NamespaceUriBase must be a valid URI.");
     }
 }
--- a/test/Microsoft.Sbom.Api.Tests/Config/ConfigurationBuilderTestsForGeneration.cs
+++ b/test/Microsoft.Sbom.Api.Tests/Config/ConfigurationBuilderTestsForGeneration.cs
@@ -293,6 +293,6 @@ public class ConfigurationBuilderTestsForGeneration : ConfigurationBuilderTestsB
             PackageSupplier = "Contoso"
         };
 
-        Assert.ThrowsExceptionAsync<ValidationArgException>(() => cb.GetConfiguration(args));
+        Assert.ThrowsExceptionAsync<ValidationArgException>(() => cb.GetConfiguration(args), "The value of NamespaceUriBase must be a valid URI.");
     }
 }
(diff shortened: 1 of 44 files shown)


test/Microsoft.Sbom.Api.Tests/Executors/ExternalDocumentReferenceWriterTest.cs
```diff
@@ -78,8 +78,6 @@ public async Task PassExternalDocumentReferenceInfosChannel_ReturnsJsonDocWithSe
         {
             var root = result.Document.RootElement;
 
-            Assert.IsNotNull(root);
```
alisonlomaka, line 81: Why remove this assert?

test/Microsoft.Sbom.Api.Tests/SignValidator/SignValidationProviderTests.cs
```diff
@@ -31,15 +31,12 @@ public void SignValidationProvider_AddsValidator_Succeeds()
...
-    [ExpectedException(typeof(ArgumentNullException))]
     public void SignValidationProvider_NullValidators_Throws()
     {
         var mockOSUtils = new Mock<IOSUtils>();
         mockOSUtils.Setup(o => o.GetCurrentOSPlatform()).Returns(OSPlatform.Windows);
 
-        var signValidator = new SignValidationProvider(null, mockOSUtils.Object);
-        signValidator.Init();
-        signValidator.Get();
+        Assert.ThrowsException<ArgumentNullException>(() => new SignValidationProvider(null, mockOSUtils.Object));
```
alisonlomaka, line 39: Is this code equivalent since it omits the Init and Get calls that the previous code made? Not sure where we were expecting the exception to occur previously.

test/Microsoft.Sbom.Api.Tests/Utils/ComponentDetectionCliArgumentBuilderTests.cs
```diff
@@ -146,27 +146,21 @@ public void Build_AddDetectorArgsViaAddArgCombineWithOtherDetectorArgs()
...
     public void Build_WithNullValue()
     {
-        var builder = new ComponentDetectionCliArgumentBuilder()
+        Assert.ThrowsException<ArgumentNullException>(() => new ComponentDetectionCliArgumentBuilder()
             .SourceDirectory("X:/")
             .AddArg("ManifestFile", null)
-            .AddArg("--DirectoryExclusionList", "X:/hello");
-
-        builder.Build();
+            .AddArg("--DirectoryExclusionList", "X:/hello"));
```
alisonlomaka, line 154: For both these tests, original code makes it look like the method under test was Build(), but it is no longer called in the new code. Is that expected?

alisonlomaka (commented):
I saw in your comments that you used tooling to assist some of these refactors - might make an interesting short and sweet tech talk/demo for the team.

test/Microsoft.Sbom.Api.Tests/Config/ConfigurationBuilderTestsForGeneration.cs
```diff
@@ -97,12 +96,11 @@ public async Task ConfigurationBuilderTest_Generation_BuildDropPathDoNotExist_Th
             PackageSupplier = "Contoso"
         };
 
-        var configuration = await cb.GetConfiguration(args);
+        Assert.ThrowsExceptionAsync<ValidationArgException>(() => cb.GetConfiguration(args));
```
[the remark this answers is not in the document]
DaveTryon: The task returns the specific expected Exception that was thrown. We only need to await if we want to capture the exception for further validation. The function itself includes an await on the action, so we don't need to add an extra wrapper if we just want to ensure that the exception matched our expectations.. Here's the actual code being run (this is from the "Go to definition" of the method call, and I've added an ALL CAPS comment where we wait):
```
    public static async Task<T> ThrowsExceptionAsync<T>(Func<Task> action) where T : Exception
    {
        return await ThrowsExceptionAsync<T>(action, string.Empty, null).ConfigureAwait(continueOnCapturedContext: false);
    }

    public static async Task<T> ThrowsExceptionAsync<T>(Func<Task> action, string message) where T : Exception
    {
        return await ThrowsExceptionAsync<T>(action, message, null).ConfigureAwait(continueOnCapturedContext: false);
    }

    public static async Task<T> ThrowsExceptionAsync<T>(Func<Task> action, string message, params object?[]? parameters) where T : Exception
    {
        Guard.NotNull(action, "action");
        Guard.NotNull(message, "message");
        string arg;
        string message2;
        try
        {
            await action().ConfigureAwait(continueOnCapturedContext: false);  // <=== THIS IS WHERE IT WAITS
        }
        catch (Exception ex)
        {
            if (!typeof(T).Equals(ex.GetType()))
            {
                arg = BuildUserMessage(message, parameters);
                message2 = string.Format(CultureInfo.CurrentCulture, FrameworkMessages.WrongExceptionThrown, arg, typeof(T), ex.GetType());
                ThrowAssertFailed("Assert.ThrowsException", message2);
            }

            return (T)ex;
        }

        arg = BuildUserMessage(message, parameters);
        message2 = string.Format(CultureInfo.CurrentCulture, FrameworkMessages.NoExceptionThrown, arg, typeof(T));
        ThrowAssertFailed("Assert.ThrowsException", message2);
        return null;
    }
```

test/Microsoft.Sbom.Api.Tests/Executors/ExternalDocumentReferenceWriterTest.cs
DaveTryon: Another case where .NET doesn't believe that `root` can be null, due to the fact that .NET 8 assumes that reference types are non-nullable unless they are declared nullable. In a case where it returns null, the test will throw a `NullReferenceException` and still fail. Not exactly what we would hope, but still OK

test/Microsoft.Sbom.Api.Tests/SignValidator/SignValidationProviderTests.cs
DaveTryon: The exception is occurring in the constructor. It doesn't feel like that was the original intent so perhaps the original test was not actually testing what it was supposed to test

test/Microsoft.Sbom.Api.Tests/Utils/ComponentDetectionCliArgumentBuilderTests.cs
DaveTryon: The exception is occurring in the constructor. It doesn't feel like that was the original intent so perhaps the original test was not actually testing what it was supposed to test

test/Microsoft.Sbom.Api.Tests/Config/ConfigurationBuilderTestsForGeneration.cs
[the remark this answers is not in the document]
DaveTryon: I see your point. I've added `await` before each instance of `Assert.ThrowsExceptionAsync`

test/Microsoft.Sbom.Api.Tests/Entities/FileValidationResultTest.cs
```diff
@@ -27,18 +27,18 @@ public void FileValidationResultErrorTypeMapping(ErrorType input, EntityErrorTyp
         var fileValidationResult = new FileValidationResult() { ErrorType = input, Path = "random" };
         var entityError = fileValidationResult.ToEntityError();
 
-        Assert.AreEqual(entityError.ErrorType, output);
-        Assert.AreEqual(entityError.Details, null);
+        Assert.AreEqual(output, entityError.ErrorType);
```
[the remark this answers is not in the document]
DaveTryon: Fixed

test/Microsoft.Sbom.Api.Tests/SignValidator/SignValidationProviderTests.cs
DaveTryon: Follow-up: The behavior being checked is indeed in the constructor. The calls to `Init` and `Get` were probably just an old copy/paste error

test/Microsoft.Sbom.Tool.Tests/IntegrationTests.cs
```diff
@@ -28,7 +28,7 @@ public static void Setup(TestContext context)
         testRunDirectory = context.TestRunDirectory;
     }
 
-    [ClassCleanup]
+    [ClassCleanup(ClassCleanupBehavior.EndOfClass)]
```
Evangelink, line 31: FYI this can also be configured at assembly level using `[assembly: ClassCleanupExecution(ClassCleanupBehavior.EndOfClass)]` so that all class cleanup are called at the end of the class and not at the end of the assembly.

test/Microsoft.Sbom.Api.Tests/Utils/ComponentDetectionCliArgumentBuilderTests.cs
DaveTryon: The Exceptions were actually coming from the `AddArg` method. I've renamed these tests to reflect what is actually being called and scoped them down to just the code that is needed

Evangelink (approved):
I would also strongly encourage you to enable test parallelization either from runsettings or using the folowing assembly level attribute `[assembly: Parallelize(Scope = ExecutionScope.MethodLevel, Workers = 0)]`.

Finally, it'd be good to also consider moving to MSTest runner, which is faster, more ligtweight and the recommended way to run your tests.

alisonlomaka (approved):

test/Microsoft.Sbom.Api.Tests/Executors/ComponentToPackageInfoConverterTests.cs
```diff
@@ -257,7 +258,8 @@ public async Task ConvertWorksWithBuildComponentPathNull()
 
         CollectionAssert.AreEquivalent(expectedPackageNames, output.Select(c => c.PackageName).ToList());
 
-        Assert.IsFalse(errors?.Any());
+        Assert.IsNotNull(errors);
```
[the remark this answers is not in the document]
Evangelink: Happy to see the rule is being liked!

DaveTryon:
> I would also strongly encourage you to enable test parallelization either from runsettings or using the folowing assembly level attribute `[assembly: Parallelize(Scope = ExecutionScope.MethodLevel, Workers = 0)]`.
> 
> Finally, it'd be good to also consider moving to MSTest runner, which is faster, more ligtweight and the recommended way to run your tests.

@Evangelink, this entire PR was unexpected and frankly work that we hadn't scheduled. It's bloomed to 43 files, which is way too big for a single PR. Trying to add parallelization to this PR seems like it would only lead to pain and more churn. Please feel free to open an issue suggesting the parallelization feature and/or moving to MSTest runner, but I'll honestly say that it's unlikely to find much traction unless there's a compelling reason--our tests are fast enough that the inner dev loop won't be significantly improved by shaving a few seconds from the run time.

JoseRenan (approved):

PR #874, opened 2025-01-15
Add ADO PR build

sfoslund:
We have discussed moving away from GH actions and using ADO pipelines for our PR CI instead. This PR is the first step in that migration. Once this is merged and I am able to get the PR pipeline running, I will follow up with another PR to remove the GH action.

### DIFF
--- a/pipelines/sbom-tool-main-build.yaml
+++ b/pipelines/sbom-tool-main-build.yaml
@@ -49,28 +49,7 @@ extends:
             name: sbom-windows-build-pool
             os: windows
           steps:
-          - task: UseDotNet@2
-            displayName: 'Use .NET Core'
-            inputs:
-              useGlobalJson: true
-
-          - task: DotNetCoreCLI@2
-            displayName: 'Restore solution'
-            inputs:
-              command: restore
-              feedsToUse: config
-              nugetConfigPath: nuget.config
-              verbosityRestore: Normal
-
-          - task: DotNetCoreCLI@2
-            displayName: Build
-            inputs:
-              arguments: '-c $(BuildConfiguration)'
-
-          - task: DotNetCoreCLI@2
-            displayName: Run tests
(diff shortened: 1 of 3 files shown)


pipelines/sbom-tool-pr-build.yaml
```diff
@@ -0,0 +1,141 @@
...
+        os: windows
+    stages:
+      - stage: stage1
+        jobs:
+        - job: Job_1
+          displayName: Build (Windows)
+          pool:
+            name: sbom-windows-build-pool
+            os: windows
+          steps:
```
DaveTryon, line 30: It seems like lines 31-52 get repeated for each OS job. To reduce potential drift, can we put them into a separate template file and reference that template file in each OS job? A separate PR might even be able to leverage the template as part of sbom-tool-main-build.yaml

sfoslund: Good idea! I've made that change, including in the main build, and kicked off a [test run of the main build](https://dev.azure.com/mseng/AzureDevOps/_build/results?buildId=29509504&view=results).

pipelines/sbom-tool-main-build.yaml
```diff
@@ -232,23 +211,7 @@ extends:
...
-              command: restore
-              feedsToUse: config
-              nugetConfigPath: nuget.config
-              verbosityRestore: Normal
-
-          - task: DotNetCoreCLI@2
-            displayName: Build
-            inputs:
-              arguments: '-c $(BuildConfiguration)'
+          - template: pipelines/build-test-tool-template.yaml@self
```
DaveTryon, line 214: Odd that we were originally only running the tests on the Windows build. I like seeing them run on all platforms 😃

sfoslund: Agreed!

PR #881, opened 2025-01-18
Update MSTest to metapackage and MTP

Youssef1313:
FYI @Evangelink

PR #895, opened 2025-01-23
Collect code coverage

DaveTryon:
We currently produce no code coverage in our build pipelines. Clicking on the "Code Coverage" tab (sample [link](https://dev.azure.com/mseng/AzureDevOps/_build/results?buildId=29532738&view=codecoverage-tab) produces a blank page. This PR collects code coverage data during tests and publishes it to the "Code Coverage" tab.

Some observations:
1. The Code Coverage tab organizes data by path. The paths are different on Windows, Linux, and macOS, and that gets confusing. To keep the tab simple, only the Windows results are published there
2. Full coverage results are published in HTML format as a build artifact
3. So far, coverage data is only available at the file level--ideally we'd be able to get method and line level information. Perhaps that's a future improvement.
4. Collecting and organizing the code coverage data makes the PR builds slower by a minute or so

### DIFF
--- a/pipelines/build-test-tool-template.yaml
+++ b/pipelines/build-test-tool-template.yaml
@@ -24,6 +24,7 @@ steps:
       arguments: '--collect:"XPlat Code Coverage;Format=Cobertura"'
 
   - task: PublishCodeCoverageResults@2
-    displayName: "Publish code coverage"
+    displayName: "Publish code coverage (Windows only)"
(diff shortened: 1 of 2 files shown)


alisonlomaka (approved):

DaveTryon:
Closing this in favor of #881, which will collect code coverage via the new test runner

PR #218, opened 2023-03-01
LicenseInfoFromFiles ignored to fit SPDX2.2 specs.

sebasgomez238:
Addresses #216 by ignoring LicenseInfoFromFiles when FilesAnalyzed is false.

PR #917, opened 2025-02-03
Remove unnecessary parser errors which disallow syft SBOMs

sfoslund:
This PR removes 2 SPDX 2.2 parser exceptions which are currently enforcing rules which are not explicitly required for SPDX 2.2 SBOMs. These errors have not been a problem thus far because the SBOMs generated by this tool (currently the only type of SBOM we parse with the parser package) do not violate these assumptions. However, when we start parsing SBOMs generated by syft, we need to allow for more variance. These changes include:

- Do not error when a file entry is missing a SHA256 hash. SHA256 is not required for 2.2 SBOMs. 
- Do not error when a package entry's LicenseInfoFromFiles property is empty, which is also not required for 2.2 SBOMs. It appears that [we previously made a change to allow this property to be empty in our own generation logic](https://github.com/microsoft/sbom-tool/pull/218) but never made the corresponding change to accept SBOMs without this property.

### DIFF
--- a/src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Utils/SPDXToSbomFormatConverterExtensions.cs
+++ b/src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Utils/SPDXToSbomFormatConverterExtensions.cs
@@ -26,10 +26,6 @@ public static class SPDXToSbomFormatConverterExtensions
     public static SbomFile ToSbomFile(this SPDXFile spdxFile)
     {
         var checksums = spdxFile.FileChecksums?.Select(c => c.ToSbomChecksum());
-        if (!checksums.Any() || checksums.All(c => c.Algorithm != AlgorithmName.SHA256))
-        {
-            throw new ParserException("File hash is missing a SHA256 value");
-        }
 
         return new SbomFile
         {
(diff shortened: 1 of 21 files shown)


src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Utils/SPDXToSbomFormatConverterExtensions.cs
```diff
@@ -26,10 +26,6 @@ public static class SPDXToSbomFormatConverterExtensions
     public static SbomFile ToSbomFile(this SPDXFile spdxFile)
     {
         var checksums = spdxFile.FileChecksums?.Select(c => c.ToSbomChecksum());
-        if (!checksums.Any() || checksums.All(c => c.Algorithm != AlgorithmName.SHA256))
-        {
-            throw new ParserException("File hash is missing a SHA256 value");
```
DaveTryon, line 31: Before this change, code coverage shows that we're covering the throw but never getting past it. Based on that, should there be at least 1 unit test change?

src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Utils/SPDXToSbomFormatConverterExtensions.cs
```diff
@@ -49,12 +45,6 @@ public static SbomFile ToSbomFile(this SPDXFile spdxFile)
     /// <returns></returns>
     public static SbomPackage ToSbomPackage(this SPDXPackage spdxPackage)
     {
-        if (spdxPackage.FilesAnalyzed &&
-            (spdxPackage.LicenseInfoFromFiles == null || !spdxPackage.LicenseInfoFromFiles.Any()))
-        {
-            throw new ParserException("Package license list was empty.");
```
DaveTryon, line 55: Before this change, code coverage shows that we're covering this throw. Does that mean that we need to update a unit test?

sfoslund: Your review was too quick for me! :) Just added the test changes

sfoslund:
Ignore the last commit, just pushed an empty commit to retrigger the CLI check that was stuck

DaveTryon (approved):

PR #860, opened 2025-01-10
Spdx 3.0 Parser for SBOM files

pragnya17:
Parsing spdx 3.0 with Unit tests and support for parsing with different compliance standards. More details here: https://dev.azure.com/mseng/1ES/_workitems/edit/2233308

PR #918, opened 2025-02-03
Update metadata contract to be backcompatible with SPDX 2.2 parser

pragnya17:
The metadata class name was updated in an earlier PR (https://github.com/microsoft/sbom-tool/pull/860/files) which was a breaking change for consumers of the SPDX 2.2 parser. We are introducing this change to allow for backwards compatibility with the SPDX 2.2 parser and compatibility with the SPDX 3.0 parser.

### DIFF
--- a/src/Microsoft.Sbom.Contracts/Contracts/SpdxMetadata.cs
+++ b/src/Microsoft.Sbom.Contracts/Contracts/SpdxMetadata.cs
@@ -9,7 +9,7 @@ namespace Microsoft.Sbom.Contracts;
 /// <summary>
 /// The object representation of all the metadata in an SPDX document.
 /// </summary>
-public abstract class SpdxMetadata
+public class SpdxMetadata
 {
     /// <summary>
     /// The version of the SPDX specification used in this document.
--- a/src/Microsoft.Sbom.Contracts/Contracts/SpdxMetadata.cs
+++ b/src/Microsoft.Sbom.Contracts/Contracts/SpdxMetadata.cs
@@ -9,7 +9,7 @@ namespace Microsoft.Sbom.Contracts;
 /// <summary>
 /// The object representation of all the metadata in an SPDX document.
 /// </summary>
-public class SpdxMetadata
+public abstract class SpdxMetadata
 {
     /// <summary>
     /// The version of the SPDX specification used in this document.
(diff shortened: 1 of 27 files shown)


sfoslund (changes requested):
The one other place we will need to change is `SbomParser`, which references `SpdxMetadata` right now in the `GetMetadata` method, making it not backwards compatible. Depending on how permanent we want these changes to be, we could do this by making ISbomParser generic depending on metadata type, or we can just temporarily revert it to only use the 2.2 metadata version

src/Microsoft.Sbom.Contracts/Contracts/SpdxMetadata.cs
```diff
@@ -9,7 +9,7 @@ namespace Microsoft.Sbom.Contracts;
 /// <summary>
 /// The object representation of all the metadata in an SPDX document.
 /// </summary>
-public class SpdxMetadata
+public abstract class SpdxMetadata
```
DaveTryon, line 12: Will this be abstract once the breaking changes are merged? If not, should it just be non-abstract now?

pragnya17: No it won't so we can just make it non-abstract

DaveTryon (approved):
Left 1 question to consider.

src/Microsoft.Sbom.Contracts/Contracts/SpdxMetadata.cs
sfoslund: Yeah when I originally made the changes that this PR are based on I was under the impression that we did not want to make any breaking changes at all, so I was experimenting with how to do that. But if this is just a temporary change and we plan on making breaking changes in a future release, then I agree that it makes sense to leave this as non-abstract and we can probably also delete `Spdx30Metadata`

pragnya17:
> The one other place we will need to change is `SbomParser`, which references `SpdxMetadata` right now in the `GetMetadata` method, making it not backwards compatible. Depending on how permanent we want these changes to be, we could do this by making ISbomParser generic depending on metadata type, or we can just temporarily revert it to only use the 2.2 metadata version

Would you mind linking which file you're referring to? I looked through all the references of GetMetadata and it seems right to me

sfoslund:
Yeah, this is where I think we need an update: https://github.com/microsoft/sbom-tool/blob/main/src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Parser/SPDXParser.cs#L175. DM'ed you to discuss this break in more detail :)

sfoslund (approved):

PR #922, opened 2025-02-05
Adding note to sbom-validate about the output path

Ninja4Code:
Added a note to the sbom-tool validate command to warn users not to output their validation file in the same folder as the drop path.  The output path needs to be outside of your drop path.  Otherwise on subsequent validations you'll receive a FAILURE even though nothing changed in your project dependencies.  This is because the validate file gets added to your drop path and it's messed up the hash comparison because you added a file into drop path.  Hopefully this makes sense.

### DIFF
--- a/docs/sbom-tool-cli-reference.md
+++ b/docs/sbom-tool-cli-reference.md
@@ -132,6 +132,8 @@ This sample command provides the minimum mandatory arguments required to validat
      `-o` is the output path where the tool will write the validation results. This path can be any file path on the system. In this case the tool will look for the validationOutputPath directory, create a file named output.json, and write the validation output.
      `-mi` is the ManifestInfo, which provides the user's desired name and version of the manifest format.
 
+NOTE: The output path should not be in the same drop path as specified in the generate command, otherwise the validation file will be added to the files to validate. The first validation will report SUCCESS but then other validate commands after it will fail.  This is because the validation file has been added in the drop path location and it will no longer validate because it detects that your validate output file has been injected into your drop path.
+
 Currently only SPDX2.2 is supported.
 
 ## Common scenarios where users can provide additional parameters


docs/sbom-tool-cli-reference.md
```diff
@@ -132,6 +132,8 @@ This sample command provides the minimum mandatory arguments required to validat
      `-o` is the output path where the tool will write the validation results. This path can be any file path on the system. In this case the tool will look for the validationOutputPath directory, create a file named output.json, and write the validation output.
      `-mi` is the ManifestInfo, which provides the user's desired name and version of the manifest format.
 
+NOTE: The output path should not be in the same drop path as specified in the generate command, otherwise the validation file will be added to the files to validate. The first validation will report SUCCESS but then other validate commands after it will fail.  This is because the validation file has been added in the drop path location and it will no longer validate because it detects that your validate output file has been injected into your drop path.
```
DaveTryon, line 135: Great callout, @Ninja4Code! This feels like something that we should treat as an error condition in a future breaking change. Here's a suggested rewording that also follows our "one space after a period" standard that we try to maintain:
```suggestion
NOTE: Do not specify an output path that is within the drop path, because the resulting output file will _change the set of files in the drop path_. The first validation will succeed, but subsequent validations will fail because the newly added file is not reflected in the SBOM. This may be treated as an error condition in future versions of the tool.
```

If you would also update the wording for the tool itself, that would be great. Instead of `The path where the output json should be written. ex: Path/output.json`, let's go with `The path where the output json should be written. Do not specify a path within your BuildDropPath. ex: Path/output.json.` in the following locations:
- https://github.com/microsoft/sbom-tool/blob/main/src/Microsoft.Sbom.Api/Config/Args/ValidationArgs.cs#L36-L37
- https://github.com/microsoft/sbom-tool/blob/main/docs/sbom-tool-arguments.md?plain=1#L23

Ninja4Code:
Dave,

I'll take your suggestions and work on the areas you listed.

Thanks,
Charles
________________________________

pragnya17:
@Ninja4Code do you have an update on the change requested for this PR?

Issue #712, opened 2024-09-19
SBOM Task Outputs Directly to Console Instead of Using MSBuild Logging APIs


JonDouglas:
When using the SBOM task in an MSBuild project, the task outputs messages directly to the console rather than utilizing MSBuild's Logging APIs. This behavior results in cluttered and unstructured console output, which is not consistent with standard MSBuild practices. Proper logging through MSBuild's APIs would allow for better categorization and filtering of messages.

Steps to Reproduce:

1. Create a .NET project and include the SBOM task.
2. Run the command `dotnet pack -bl` to build the project with a binary log.
3. Observe the console output during the build process.

Expected behavior:

Sbom task should output messages through MSBuild logging APIs & categorized appropriately.

We can utilize [`TaskLoggingHelper`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.build.utilities.tasklogginghelper?view=msbuild-17-netcore):

```
public class SbomTask : Microsoft.Build.Utilities.Task
{
    public override bool Execute()
    {
        // Example of logging an informational message
        Log.LogMessage(MessageImportance.High, "Finding components...");
        // Example of logging a warning
        Log.LogWarning("No instructions received to scan docker images.");
        // Example of logging an error
        Log.LogError("An error occurred during SBOM generation.");

        // Rest of the task implementation

        return !Log.HasLoggedErrors;
    }
}

```

sfoslund:
@baronfel @KalleOlaviNiemitalo I assumed that you get MSBuild logging for free when using ToolTasks, is that not the case? I have fixed the issue that was blocking the .NET framework case and in my testing its logging output visually appears to be coming via the MSBuild APIs but I'm not entirely sure how to tell. 

As for the non-.NET framework case, it seems like some additional investigation is required here to figure out how best to implement this. My first thought was to add an option to `AddSbomTool` (which is called in the task set up) to allow users to request a MSBuild logger instead of our normal serilog logger, but I haven't had time to look into this further and determine if that would actually work. If you think making a bridge between serilog and the MSbuild logger would work/ be simpler I'd love to hear any more information you have

baronfel:
@sfoslund ToolTask does do the logging - the case I was talking about was the modern .NET case. I think your proposed path should be ok - you know the logging infra for your tool better than I do :) - and I don't have any specific requirements there aside from "don't write directly to stdout".

sfoslund:
[...]

Issue #616, opened 2024-07-17
SBOM API will show `information` logs when verbosity is set to be more restrictive


gustavoaca1997:
Even though we are setting the verbosity to values like Warning, Fatal or Error, we still see the `information` logs:

```powershell
[...]

PR #933, opened 2025-02-16
Use MSBuild Logging APIs for Task output instead of StdOut

baronfel:
Fixes https://github.com/microsoft/sbom-tool/issues/712
Fixes https://github.com/microsoft/sbom-tool/issues/616

This is a refresh of https://github.com/microsoft/sbom-tool/pull/730 that I've gotten the tests to green up on. Along the way I think I've also addressed https://github.com/microsoft/sbom-tool/issues/616, which was a consequence of way the dependencies of the task were initialized in the .NET task.

Since the majority of the tool uses Serilog, the strategy here is to integrate MSBuild's logging APIs into Serilog's logging configuration where possible. This means, broadly, two things:

* creating and using an MSBuild Logger that converts Serilog's messages to calls to the `TaskLoggingHelper`'s logging methods
* ensuring that the Verbosity selected in the Task is used when configuring Serilog

Along the way, I've made it easier to test the post-conditions of the Task on both net472 and .NET.


baronfel:
One big note - I'm using .NET SDK 9.0.200, where some of the Roslyn analyzers have been tweaked/fixed and so are reporting things that I had to fix in order to get the repo to compile.

src/Microsoft.Sbom.Extensions.DependencyInjection/MSBuildLogger.cs
```diff
@@ -0,0 +1,45 @@
...
+            case LogEventLevel.Warning:
+                loggingHelper.LogWarning(logEvent.RenderMessage());
+                break;
+            case LogEventLevel.Error:
+            case LogEventLevel.Fatal:
+                loggingHelper.LogError(logEvent.RenderMessage());
+                break;
+            default:
+                break;
+        }
```
baronfel, line 43: This is the core of the Serilog<->MSBuild bridge

src/Microsoft.Sbom.Extensions.DependencyInjection/ServiceCollectionExtensions.cs
```diff
@@ -200,26 +201,33 @@ public static IServiceCollection ConfigureLoggingProviders(this IServiceCollecti
...
-                .CreateLogger());
+                            // Don't write the detection times table from DetectorProcessingService to the console, only the log file
+                            .Filter.ByExcluding(Matching.WithProperty<string>("DetectionTimeLine", x => !string.IsNullOrEmpty(x)))),
+                        1) // sinkMapCountLimit
+                    .CreateLogger());
+        }
+        else
+        {
+            return new MSBuildLogger(taskLoggingHelper);
+        }
```
baronfel, line 231: And here is where we pick between the MSBuild path and the Serilog-controlled-output path

src/Microsoft.Sbom.Targets/GenerateSbomTask.cs
```diff
@@ -20,56 +21,62 @@ namespace Microsoft.Sbom.Targets;
...
+               .AddSingleton<ISourcesProvider, CGScannedPackagesProvider>()
+               .AddSingleton<IAlgorithmNames, AlgorithmNames>()
+               .AddSingleton<IManifestGenerator, Generator>()
+               .AddSingleton<IMetadataProvider, LocalMetadataProvider>()
+               .AddSingleton<IMetadataProvider, SBOMApiMetadataProvider>()
+               .AddSingleton<IManifestInterface, Validator>()
+               .AddSingleton<IManifestConfigHandler, SPDX22ManifestConfigHandler>())
+               .Build();
+
+            var generator = taskHost.Services.GetRequiredService<ISBOMGenerator>();
```
baronfel, line 78: Big change here. Instead of putting the DI container in the constructor, I moved it into `Execute` so that we could call the Verbosity-defaulting assignment, so we could flow that defaulted verbosity into the Serilog configuration calls.

src/Microsoft.Sbom.Targets/SbomInputValidator.cs
```diff
@@ -85,16 +85,22 @@ public EventLevel ValidateAndAssignVerbosity()
         switch (this.Verbosity.ToLower().Trim())
         {
             case "verbose":
+                this.Verbosity = "Verbose";
```
baronfel, line 88: Adding these meant that the `Verbosity` member on the `GenerateSbom` Task could be checked in tests to verify postconditions more easily than scraping logs.

src/Microsoft.Sbom.Api/Executors/ComponentDetectionBaseWalker.cs
```diff
@@ -108,6 +108,8 @@ async Task Scan(string path)
             var cmdLineParams = configuration.ToComponentDetectorCommandLineParams(cliArgumentBuilder);
 
             var scanSettings = cliArgumentBuilder.BuildScanSettingsFromParsedArgs(cmdLineParams);
+            scanSettings.NoSummary = true;
+            scanSettings.DetectorCategories = ["NuGet"];
```
baronfel, line 112: I could use some specific help here. I think we should be able to set the first of this _always_ for the MSBuild Targets - we never want anything printing to stdout, and without this flag, the component-detection library writes a report to stdout directly instead of the given ILogger. This is not ideal, I would consider a bug against the library for corrupting stdout.

The second line I _know_ this is not the right location for, but I think limiting the scope of scanned artifacts to NuGet packages in a .NET project is a defensible default, and it saves an entire 1.5s off of the ~6.75s it takes to run this target on an almost empty class library project.

I think we need some way to thread these requests down from the MSBuild Task, but it's abstracted away enough that it's not clear to me how to do this.

baronfel:
I've attached a zipped binlog of the tool working with these changes. The detailed logs are in the binlog as expected.

[sbom-pack.zip](https://github.com/user-attachments/files/18817297/sbom-pack.zip)

To the end user, it looks like this:

```terminal
> dotnet pack -bl
Restore complete (0.3s)
  sbom-tests succeeded with 1 warning(s) (5.6s) → bin\Release\net9.0\sbom-tests.dll
    E:\code\nuget-packages\microsoft.sbom.targets\3.1.1-preview.0.6\build\Microsoft.Sbom.Targets.targets(57,5): warning : Deleting pre-existing folder E:\Code\Scratch\sbom-tests\bin\Release\sbom-tests.1.0.0.4b5fcd2f.temp\_manifest as DeleteManifestDirIfPresent is 'true'.

Build succeeded with 1 warning(s) in 6.2s
```

I'd like to dig in a _bit_ more about the spurious warning, I expect that's a mismatched warning regex or something on MSbuild's side.

src/Microsoft.Sbom.Api/Hashing/IHashAlgorithmProvider.cs
```diff
@@ -7,5 +7,5 @@ namespace Microsoft.Sbom.Api.Hashing;
 
 public interface IHashAlgorithmProvider
 {
-    AlgorithmName Get(string algorithmName);
+    public AlgorithmName Get(string algorithmName);
```
sfoslund, line 10: Are all of these 'public' changes necessary? I would be fine with these changes but I would prefer that they be split off into a separate PR if possible

src/Microsoft.Sbom.Tool/Microsoft.Sbom.Tool.csproj
```diff
@@ -24,5 +24,6 @@
   <ItemGroup>
     <PackageReference Include="AutoMapper.Extensions.Microsoft.DependencyInjection" />
     <PackageReference Include="Microsoft.Extensions.Hosting" />
+    <PackageReference Include="Microsoft.Build.Utilities.Core" />
```
DaveTryon, line 27: It shouldn't be necessary to add `Microsoft.Build.Utilities.Core` to the standalone tool
```suggestion
```

src/Microsoft.Sbom.Extensions.DependencyInjection/Microsoft.Sbom.Extensions.DependencyInjection.csproj
```diff
@@ -18,6 +18,7 @@
 
   <ItemGroup>
     <PackageReference Include="AutoMapper.Extensions.Microsoft.DependencyInjection" />
+    <PackageReference Include="Microsoft.Build.Utilities.Core" />
```
DaveTryon, line 21: I really don't like the idea of adding the `Microsoft.Build.*` packages to the standalone tool. They were included exclusively for the functionality provided by the `Microsoft.Sbom.Targets` project, and I'd love to keep things that way. Is there a way to move the logging mapping out of the dependency injection assembly? One thought that comes to mind is to replace the `TaskLoggingHelper` parameter with some sort of delegate, and placing the creation of the `MSBuildLogger` within that delegate. If no delegate is provided, then the existing code gets run

src/Microsoft.Sbom.Extensions.DependencyInjection/MSBuildLogger.cs
DaveTryon: As called out elsewhere, I'd love to see this live in the `Microsoft.Sbom.Targets` project, and use some sort of delegate to pass it into the DependencyInjection project. The idea is to keep the MDBuild dependencies tightly scoped, since their functionality is exclusive to the NuGet targeting

src/Microsoft.Sbom.Api/Hashing/IHashAlgorithmProvider.cs
baronfel: I can pull them out - heads-up that any 9.0.200 usage is going to bring them back though

DaveTryon:
> https://github.com/microsoft/sbom-tool/blob/3d1a59f0947cb868708eaccff243c83410f1f57b/src/Microsoft.Sbom.Api/Workflows/SBOMGenerationWorkflow.cs#L240-L242
> 
> . Is this _really_ a warning? A user asked to delete a manifest if it was already there, and we're doing what they asked? This feels more like a debug/diagnostic to me. The big problem here is twofold:
> * many users run MSBuild with TreatWarningsAsErrors enabled
> * this particular warning, because it's coming from deep in the Serilog stack and not from direct use of MSBuild Logging APIs, doesn't have a _Code_ associated with it, so users can't use MSbuild's knobs to disable/hide this warning when it _does_ appear

Please feel free to change this `log.Warning` to `log.Info`

src/Microsoft.Sbom.Api/Executors/ComponentDetectionBaseWalker.cs
DaveTryon: Let's focus on the scope of _this_ PR, which is to deal with the stdout logging. Perf improvements and/or potentially refactoring the MSBuild task should really be kept for subsequent PR's.

src/Microsoft.Sbom.Extensions.DependencyInjection/ServiceCollectionExtensions.cs
```diff
@@ -55,20 +55,29 @@ public static IServiceCollection AddSbomConfiguration(this IServiceCollection se
...
+                {
+                    return logger;
+                }
+                else
+                {
+                    var level = x.GetService<InputConfiguration>()?.Verbosity?.Value ?? LogEventLevel.Information;
+                    var defaultLogger = CreateDefaultLogger(level);
+                    Log.Logger = defaultLogger;
+                    return defaultLogger;
+                }
```
baronfel, line 78: This change was made because when I tried to pass in an optional ILogger, the separate transient ILogger resolution that was side-effecting `Log.Logger` seemed to cause an infinite loop. Since there's only one ILogger, making these a) singleton, and b) doing the Log.Logger overriding only for cases where the logger wasn't explicitly known beforehand seems to solve the problem.  The tool-driven use case shouldn't be effected here.

src/Microsoft.Sbom.Targets/GenerateSbomTask.cs
```diff
@@ -24,52 +27,62 @@ namespace Microsoft.Sbom.Targets;
 /// <summary>
 /// MSBuild task for generating SBOMs from build output.
 /// </summary>
-public partial class GenerateSbom : Task
+public partial class GenerateSbom : Task, ICancelableTask
```
baronfel, line 30: Added `CancelableTask` support to help clean up - though many of the task-using calls lower in the stack don't support CancellationTokens. The MSBuild side is now ready to handle them if/when they get support - right now you get a pretty nasty build hang so eventually it would be nice to fix that.

src/Microsoft.Sbom.Targets/GenerateSbomTask.cs
```diff
@@ -84,25 +97,36 @@ public override bool Execute()
...
                 runtimeConfiguration: runtimeConfiguration,
                 specifications: ValidateAndAssignSpecifications(),
-                externalDocumentReferenceListFile: this.ExternalDocumentListFile)).GetAwaiter().GetResult();
+                externalDocumentReferenceListFile: this.ExternalDocumentListFile),
+                this.cancellationTokenSource.Token).GetAwaiter().GetResult();
 #pragma warning restore VSTHRD002 // Avoid problematic synchronous waits
 
+            if (!result.IsSuccessful)
+            {
+                Log.LogError("SBOM generation failed. Check the build log for details.");
```
baronfel, line 116: This is a user-facing string that tell the user that something went wrong - since I'm not sure what a reasonable summary of the kinds of errors might be, I instead chose to emit details to the MSBuild log at a more verbose log level. Users that need details can check that out.

src/Microsoft.Sbom.Targets/MSBuildLogger.cs
```diff
@@ -0,0 +1,59 @@
...
+using Microsoft.Build.Utilities;
+using Serilog.Events;
+using ILogger = Serilog.ILogger;
+
+/// <summary>
+/// A class to remap Errors logged from ComponentDetection assemblies to Warnings or Errors in MSBuild from Serilog to MSBuild's native logging facilities. This class interprets a few kinds of properties from the
+/// logged Serilog message:
+/// <list type="bullet">
+/// <item><description><c>msbuild.code</c> becomes the <c>code</c> for the <see cref="TaskLoggingHelper.LogWarning(string, string, string, string, string, int, int, int, int, string, object[])" /> or <see cref="TaskLoggingHelper.LogError(string, string, string, string, string, int, int, int, int, string, object[])"/> methods for Serilog <see cref="LogEventLevel.Warning"/>, <see cref="LogEventLevel.Error"/>, and <see cref="LogEventLevel.Fatal"/> events.</description></item>
+/// <item><description><c>msbuild.importance</c> becomes the <c>importance</c> for the <see cref="TaskLoggingHelper.LogMessage(MessageImportance, string, object[])" /> method for Serilog <see cref="LogEventLevel.Debug"/> and <see cref="LogEventLevel.Information"/> events.</description></item>
```
baronfel, line 17: These aren't _currently_ used in this repo, but if you wanted to make certain kinds of errors user-suppressible you'd do this with _code_s - and this plumbing means it should be easy to add them as state to the serilog messages.

baronfel:
I've taken another pass at this, addressing the feedback (thanks folks!) and I'm much happier with how this integration looks now. There's only one actually _functional_ gap, and I need a bit of direction on where I should go with that.

src/Microsoft.Sbom.Api/Workflows/SBOMGenerationWorkflow.cs
```diff
@@ -237,7 +237,7 @@ private void RemoveExistingManifestDirectory()
                         $"overwrite this folder.");
                 }
 
-                log.Warning(
+                log.Information(
```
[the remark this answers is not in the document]
baronfel: If we leave this as a warning, then it need to be assigned a Code so that MSBuild users can ignore it or otherwise manage it. Without a Code, users with `TreatWarningsAsErrors` will fail and not have any mechanism to change that, which is a _huge_ area of push back from users.

src/Microsoft.Sbom.Api/Executors/ComponentDetectionBaseWalker.cs
```diff
@@ -108,7 +108,8 @@ async Task Scan(string path)
             var cmdLineParams = configuration.ToComponentDetectorCommandLineParams(cliArgumentBuilder);
 
             var scanSettings = cliArgumentBuilder.BuildScanSettingsFromParsedArgs(cmdLineParams);
-
+            // TODO: the MSBuild pathway needs a way to disable this setting, because the CG tools write directly to stdout (which is not kind).
+            scanSettings.NoSummary = true;
```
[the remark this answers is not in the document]
baronfel: I agree - I just don't know what the best way to do that would be because we're like 5 layers deep in DI indirection here. Got any ideas for if there's some kind of knob that's already available that I can twiddle? The main pathway seems to be something like:

`IConfiguration -> string[] -> parsing routing -> ScanSettings` - should I add a flag to IConfiguration and have the MSbuild-DI set up an IConfiguration that sets that flag?

src/Microsoft.Sbom.Api/Workflows/SBOMGenerationWorkflow.cs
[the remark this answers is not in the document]
sfoslund: Gotcha. Information is okay with me then, but tagging @DaveTryon in case he has an opinion

src/Microsoft.Sbom.Targets/GenerateSbomTask.cs
```diff
@@ -24,52 +27,62 @@ namespace Microsoft.Sbom.Targets;
...
     }
 
+    public void Cancel() => cancellationTokenSource.Cancel();
+
     /// <inheritdoc/>
     public override bool Execute()
     {
         try
         {
+            var taskLoggingHelper = new TaskLoggingHelper(this);
```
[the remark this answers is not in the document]
baronfel: This was just a miss on my part - I forgot that Task.Log was a TaskLoggingHelper already.

src/Microsoft.Sbom.Api/Executors/ComponentDetectionBaseWalker.cs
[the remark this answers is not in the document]
sfoslund: It's ugly, but yeah that sounds reasonable to me :( again tagging @DaveTryon who might have thoughts

src/Microsoft.Sbom.Targets/GenerateSbomTask.cs
```diff
@@ -84,25 +97,36 @@ public override bool Execute()
...
 #pragma warning restore VSTHRD002 // Avoid problematic synchronous waits
 
+            if (!result.IsSuccessful)
+            {
+                Log.LogError("SBOM generation failed. Check the build log for details.");
+            }
+
+            foreach (var error in result.Errors)
+            {
+                Log.LogMessage(MessageImportance.Normal, "{0}({1}) - {2} - {3}", error.Entity.Id, error.Entity.EntityType, error.ErrorType, error.Details);
```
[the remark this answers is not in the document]
baronfel: It's worth a try :)

src/Microsoft.Sbom.Api/Executors/ComponentDetectionBaseWalker.cs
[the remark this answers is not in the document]
DaveTryon: I agree with @sfoslund. `IConfiguration` changes fall into the category of "these are breaking changes if someone implements a decorator in their code`, so we need to use them carefully. Given that we're about to release a breaking API change, this is probably as good a time as any to change the interface

baronfel:
I've just rebased against main - the major conflict was that there's now SPDX3.0 configurations in the DI registration.

I've addressed all outstanding feedback I think - the last thing is to check out some hangs that @sfoslund reported to me today in the CLI tool itself.

src/Microsoft.Sbom.Api/Executors/ComponentDetectionBaseWalker.cs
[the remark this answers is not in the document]
baronfel: How about the `RuntimeConfiguration` change I did instead? @sfoslund prototyped that and it seems doable?

baronfel:
@sfoslund the loop is coming from a circular dependency in the DI configuration. `ILogger` requires knowing the verbosity (from `InputConfiguration`) in order to initialize, but reading the `InputConfiguration` requires `IConfigurationBuilder<T>`, which requires `ConfigFileParser`, which requires `IFileSystemUtils`, all implementations of which require `ILogger`.

baronfel:
Alright, fixed the loop by using the strategy from `main` - the FileSystemUtils get a default logger with an expected verbosity.
Also fixed more warnings coming from component detection, but in a way that remaps all of them cleanly to MSBuild Information without impacting the tool experience.

I am now _completely_ satisfied with this PR. *whew*!

src/Microsoft.Sbom.Extensions.DependencyInjection/ServiceCollectionExtensions.cs
```diff
@@ -46,29 +45,11 @@ namespace Microsoft.Sbom.Extensions.DependencyInjection;
...
-        return services;
-    }
-
-    public static IServiceCollection AddSbomTool(this IServiceCollection services, LogEventLevel logLevel = LogEventLevel.Information)
+    public static IServiceCollection AddSbomTool(this IServiceCollection services)
     {
         services
             .AddSingleton<IConfiguration, Configuration>()
-            .AddTransient(_ => FileSystemUtilsProvider.CreateInstance(CreateLogger(logLevel)))
-            .AddTransient(x =>
```
sfoslund, line 67: My understanding is that this will create a breaking change for API users, as they will now have to inject a logger similarly to how we are now doing it in Program.cs, is that right? This is okay with me as our next release will be a breaking change anyways, but tagging @DaveTryon for visibility

sfoslund:
Looks like we're seeing some CI unit test failures with this change: https://dev.azure.com/mseng/1ES/_build/results?buildId=29662326&view=logs&j=aeb7b77f-2f5c-5c5e-fb92-f326324a83fb&t=947a9a91-a289-5b68-4a98-262ba6071be8

src/Microsoft.Sbom.Extensions.DependencyInjection/ServiceCollectionExtensions.cs
baronfel: Oof, good call - I didn't realize that the dependency injection project was public API surface area. Let me noodle on how that could be fixed.

sfoslund:
Also @DaveTryon do you know why the 'Check for API changes' commentor check is failing?

src/Microsoft.Sbom.Targets/GenerateSbomTask.cs
```diff
@@ -74,6 +45,51 @@ public override bool Execute()
...
+            var runtimeConfiguration = new RuntimeConfiguration
+            {
+                NamespaceUriBase = this.NamespaceBaseUri,
+                NamespaceUriUniquePart = this.NamespaceUriUniquePart,
+                DeleteManifestDirectoryIfPresent = this.DeleteManifestDirIfPresent,
+                Verbosity = logVerbosity,
+                NoComponentGovernanceSummary = true
+            };
+
+            var taskHost = Host.CreateDefaultBuilder()
```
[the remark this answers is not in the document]
baronfel: Great call - I've done so now.

src/Microsoft.Sbom.Extensions.DependencyInjection/ServiceCollectionExtensions.cs
baronfel: I've pushed a commit that minimizes the API surface area changes, and adds a ton of xmldoc that instructs users on the new requirement of registering their own `ILogger`. That won't stop someone that just updates and then runs tests from failing, but maybe that will help mitigate somewhat.

baronfel:
The last failures I think are real - I've seen them when packing my own projects using the newly-built nuget package manually. I'll test with a package from `main` to compare. The failures are a double-access to the manifest.spdx.json - I've only seen them happen for the spdx_3.0 pathway, but it's possible the 2.2 pathway could also be impacted? The stack looks like this:

```
System.IO.IOException: The process cannot access the file 'E:\Code\sbom-tool\test\Microsoft.Sbom.Tool.Tests\bin\Debug\net8.0\TestResults\Deploy_chusk 20250301T213843_15624\E2E_GenerateManifest_GeneratesManifest_ReturnsZeroExitCode\_manifest\spdx_3.0\manifest.spdx.json' because it is being used by another process.
   at Microsoft.Win32.SafeHandles.SafeFileHandle.CreateFile(String fullPath, FileMode mode, FileAccess access, FileShare share, FileOptions options)
   at Microsoft.Win32.SafeHandles.SafeFileHandle.Open(String fullPath, FileMode mode, FileAccess access, FileShare share, FileOptions options, Int64 preallocationSize, Nullable`1 unixCreateMode)
   at System.IO.Strategies.OSFileStreamStrategy..ctor(String path, FileMode mode, FileAccess access, FileShare share, FileOptions options, Int64 preallocationSize, Nullable`1 unixCreateMode)
   at System.IO.Strategies.FileStreamHelpers.ChooseStrategyCore(String path, FileMode mode, FileAccess access, FileShare share, FileOptions options, Int64 preallocationSize, Nullable`1 unixCreateMode)
   at System.IO.FileInfo.OpenRead()
   at Microsoft.ComponentDetection.Common.LazyComponentStream.SafeOpenFile()
```

sfoslund:
You're seeing the above error only with spdx 3.0? @pragnya17 is currently working on SPDX 3.0 and is not yet finished so I expect this issue will be addressed by her future work. Regardless, AFAIK we do not yet have SPDX 3.0 test coverage for the .NET integration in this repo so that wouldn't explain the test failures we're seeing. It looks like it's something related to DI added in this PR based on this line: 
[error : SBOM generation failed: Method not found: 'Microsoft.Extensions.DependencyInjection.IServiceCollection Microsoft.Extensions.DependencyInjection.OptionsServiceCollectionExtensions.Configure(Microsoft.Extensions.DependencyInjection.IServiceCollection, System.Action`1<!!0>)'.](https://dev.azure.com/mseng/1ES/_build/results?buildId=29671974&view=logs&j=e811f8ac-a658-51ed-df8d-54b37820cc23&t=fded921d-1ce1-56c0-5fd8-861d61c15562&l=1406)

sfoslund:
It looks like that error is still showing up in our CI runs on this PR: https://dev.azure.com/mseng/1ES/_build/results?buildId=29671974&view=results

pipelines/build-test-tool-template.yaml
```diff
@@ -1,45 +1,76 @@
 steps:
   - task: UseDotNet@2
-    displayName: 'Use .NET Core'
+    displayName: Use .NET Core
```
sfoslund, line 3: Please revert changes in this file

src/Microsoft.Sbom.Api/Microsoft.Sbom.Api.csproj
```diff
@@ -24,17 +24,11 @@
         <PackageReference Include="Serilog.Extensions.Hosting" />
         <PackageReference Include="Serilog.Sinks.Console" />
         <PackageReference Include="Spectre.Console.Cli" />
-        <PackageReference Include="System.IO.FileSystem.AccessControl" />
```
sfoslund, line 27: What is the reason for the changes in this file?

src/Microsoft.Sbom.Common/Microsoft.Sbom.Common.csproj
```diff
@@ -10,7 +10,6 @@
 
   <ItemGroup>
     <PackageReference Include="Serilog.Sinks.Console" />
-    <PackageReference Include="System.IO.FileSystem.AccessControl" />
```
sfoslund, line 13: What is the reason for the change in this file?

src/Microsoft.Sbom.Extensions.DependencyInjection/ServiceCollectionExtensions.cs
sfoslund: @Forgind is there a way we can avoid making a breaking change here?

test/Microsoft.Sbom.Targets.E2E.Tests/GenerateSbomE2ETests.cs
```diff
@@ -146,80 +149,6 @@ private void InspectPackageIsWellFormed(bool isManifestPathGenerated = true)
         }
     }
 
-    [TestMethod]
-    public void SbomGenerationSucceedsForDefaultProperties()
```
sfoslund, line 150: @Forgind can you please provide your reasoning for deleting this tests here for visibility for the whole team?

src/Microsoft.Sbom.Api/Microsoft.Sbom.Api.csproj
Forgind: When I built locally, I got an error that some package references might not be needed, and it pointed here. I deleted it, and it still built, so it seems to have been right.

src/Microsoft.Sbom.Common/Microsoft.Sbom.Common.csproj
Forgind: Same as previous

pipelines/build-test-tool-template.yaml
Forgind: Why do you dislike these changes? Is it just because they're unrelated to the PR? They eliminated a bunch of warnings in the CI pipeline, which seems like a plus to me.

src/Microsoft.Sbom.Extensions.DependencyInjection/ServiceCollectionExtensions.cs
Forgind: baronfel didn't come up with a way, and he knows this part much better than I do. I can try to look around, but I have little hope of success.

test/Microsoft.Sbom.Targets.E2E.Tests/GenerateSbomE2ETests.cs
Forgind: Some of the test code multitargets to net8.0 and net472. There's a target there that only executes for net472. It builds another project that only targets net8.0—and it passed net8.0 as the targetframework with which to build that project—then copies some of its assets over and tries to use them. That isn't a supported scenario and never has been. End-to-end tests should only exist if they're testing supported scenarios. It now fails because it copies in baronfel's new thing (that uses APIs that don't exist in net472). I don't think it really needs to execute them (might be wrong on that) just pack properly, but that involves loading the assembly, which causes a missing method exception.

If we wanted to avoid deleting the tests, what I'd do is either:
 - Make the relevant project multitarget to net8.0.
   - It also depends on a lot of other projects (and packages) that target net8.0, so we'd either need to make those conditional or have them also target net472.
   - This would be reasonably robust to additional changes.
 - Choose specific files to copy or not copy to the net472 project's output directory.
   - Figuring out which files would be...tricky...and it would still be an unsupported scenario.
   - This would be very fragile. A small change could break it and lead to just as much confusion as I faced.

I'm not terribly fond of either option. The second would just be kicking the can down the road, as far as I'm concerned, and the first would likely require major (infrastructure-y) changes.

pipelines/build-test-tool-template.yaml
sfoslund: I'm seeing a lot of commented code and unnecessary changes (like quotes and dotnet --info) which I don't like so it's hard to tell which of these changes are actually functional. Can you revert the unnecessary changes/ commented code at a minimum and I'll consider whatever's left over?

Forgind: I think it's about as small a change as I can make it. I considered combining restore/build or even restore/build/test, but there isn't much harm to keeping them separate.

pragnya17:
Waiting on .NET team to get back to this PR

Forgind:
> Waiting on .NET team to get back to this PR

Last I checked (in April), this was just waiting on approval.

Forgind:
And I just resolved the merge conflict.

sfoslund:
> Last I checked (in April), this was just waiting on approval.

I don't think that's accurate, some of my previous comments have not been resolved (like reverting changes to [build-test-tool-template.yaml](https://github.com/microsoft/sbom-tool/pull/933/files/264b28932de0bb1f972e9dc7362eee8b229ab300#diff-bc1c8f1d5b68852082e8c884e08c3509526208917af8e442e0ab0f177f5ce9a0)) and I believe this PR also contains a breaking change, which would need to be addressed

Forgind:
The breaking change part was discussed already in [this thread](https://github.com/microsoft/sbom-tool/pull/933#discussion_r2049342923). It sounded like you wanted to avoid the breaking change, but baronfel did try to mitigate as much as possible, so I don't think there's much more to do there unless you'd rather just close this PR.

As to the build-test-tool-template.yaml changes, I minimized those, as you can see from my last two commits. It's been a couple months, but my vague recollection is that the previous version chose the wrong version of the .NET SDK, and this PR revealed that. I'd rather not un-fix a bug.

sfoslund:

PR #936, opened 2025-02-18
Fix NugetPackagesPath

hsebs:
`globalPackagesFolder` can be redefined by `nuget.config`.

`NullSettings` does not use `nuget.config`.

We should use `DefaultSettings` instead of `NullSettings`.


sfoslund:

PR #924, opened 2025-02-07
API changes to support SPDX 3.0

pragnya17:
1. Update the validation/parsing and generation workflows to reflect the different SPDX 3.0 defined entities. 
2. Add a cmd line parameter for different compliance standards. For example, `parse SbomFileExample.txt -complianceStandard NTIA`
3. Make sure there are no duplicate spdx elements
4. Unit testing command line API changes, workflows, utility methods

Issue #685, opened 2024-08-23
SBOMxxxx vs Sbomxxxx inconsistencies


vichor:
I am following the [API reference documentation](https://github.com/microsoft/sbom-tool/blob/main/docs/sbom-tool-api-reference.md), specifically the section [Self-provided data-based SBOM generator API](https://github.com/microsoft/sbom-tool/blob/main/docs/sbom-tool-api-reference.md#self-provided-data-based-sbom-generator-api). There the snippet states to create a new instance of the class SBOMFile, only that this type does not exist.

Looking for SBOMFile in the code finds an SBOMFile.cs file which defines the class SbomFile.

I have checked other elements and the same situation appears, but not for all of them. It seems, thus, there is some inconsistency that would be good to fix:
- The documentation refer to SBOMxxxx class
- The filesystem defines an SBOMxxxx.cs file
- The class is called Sbomxxxx

xxxx may be File, Package, Specification, GenerationResult, 
Elements that do not present this issue are Reference, Metadata, Relationship, ValidationResult

DaveTryon:
Hi, @vichor! Admittedly this isn't great. It looks like the mixed-case "Sbom" started to appear later in the development process and was never fully implemented. Unfortunately, our team is focusing on other things right now and just don't have the time to commit to "nice to have" items. You could open a PR, if you want, and we'll consider merging it. Probably the preferred option would be to follow the convention of using SBOM for consistency with the file names and the documentation. That's straightforward with Visual Studio's rename ability, which will update all other references at the same time. The tricky bit is that we need to maintain backward compatibility with the old names. So if you were updating the SbomFoo class, you'd do something like this:

```
#pragma warning disable SA1402 // File may only contain a single type
public class SBOMFoo
#pragma warning restore SA1402 // File may only contain a single type
{
  // stuff
}

/// <summary>
/// Provided for backwards compatibility. Please make all changes in the SBOMFoo class
/// </summary>
public class SbomFoo : SBOMFoo
{
  // Empty except for non-default constructors
}
```

You'd also want to have at least 1 unit test for each of the legacy names, probably something along the lines of

```
[TestMethod]
public void LegacyTypeIsSupported()
{
  var legacy = new SbomFoo();
  Assert.IsInstanceOfType(legacy, typeof(SBOMFoo));
}

PR #947, opened 2025-02-24
Normalize "SBOM" to "Sbom" through most of the code base

DaveTryon:
As called out in #685, our code is very inconsistent in its usage of "SBOM" and "Sbom". Older files seemed to start with "SBOM", then migrated somewhat toward "Sbom", but we have a bit of a mix. We punted on #685 because of compatibility concerns.

As we've discussed separately, SPDX 3.0 is a breaking change, so we have a window of opportunity to standardize our API. This PR is an attempt to name that happen, following the following rules:
1. The new standard is to follow PascalCasing
2. File names and class names are case-consistent
3. Let Visual Studio automate as many of the changes as possible
4. Make the commits as tightly scoped as possible
5. Try to keep the docs in sync with the code changes
6. To keep things clean, use "Sbom" in all code constructs, with 2 exceptions to come later:
  1. Files that include "SPDX" in the name (minimize merge conflicts with #924 
  2. Leave the test files alone where possible. I think I renamed 1 test file by mistake, but I can back that out if we want
7. Since I'm working on Windows, which ignores case differences, the file renames were a 2-step process:
  1. Rename `SBOMFoo` to `SBOMFoo_`, allowing Visual Studio to update all usages
  2. Rename `SBOMFoo_` to `SbomFoo`, allowing Visual Studio to update all usages
8. Build and run tests frequently along the way to make sure that we don't perpetuate breakage
9. Search for `SBOM\w\w` using regular expressions as a way to find cases that may have been missed

The PR itself is broad (92 files), but most of the review should be trivial.

Proposed roadmap:
1. [x] Merge #924
2. [x] Update this PR to reflect changes from #924
3. [ ] Merge this PR
4. [ ] Decide if we want to normalize "SPDX" to "Spdx". If so, create a similar PR
5. [ ] Create a **migration document** from API 3.x to 4.0 (we need do this even if we choose not to normalize the casing)
6. [ ] Normalize casing in test names (doesn't impact the API, could happen after we release the 4.0 API)


DaveTryon:
Merge state seems to be messed up. Closing and replacing with #950

PR #953, opened 2025-02-28
Check for invalid manifestInfo's during the redaction workflow

pragnya17:
Redaction for SBOM's is only supported for SPDX 2.2 manifests. Any other manifestInfo should not be supported when the customer uses the -redact CLI command on the SBOM tool.


pragnya17:
This PR https://github.com/microsoft/sbom-tool/pull/961 has the changes instead

Issue #930, opened 2025-02-12
Problems running unit tests locally


DaveTryon:
On some machines, running `dotnet test` has recently begun to fail on E2E tests. The log output from these tests suggests that the component scanning is getting thrown off by the fact that the test logs are being written within the scope of the folders that are being scanned, so we end up with locked files. (Kudos to @alisonlomaka for figuring this out).

Our build pipelines specify a different results directory, so they're just fine. If you hit this error running tests locally, the workaround is to specify a different restuls directory, along the lines of
```
dotnet test -- --results-directory %temp%\sbom-tool-test-results
```

We're not sure why this suddenly changed--possibly a recent change to the test runner. We might be able to change the default behavior by using a `.runsettings` file as documented at https://learn.microsoft.com/en-us/visualstudio/test/configure-unit-tests-by-using-a-dot-runsettings-file?view=vs-2022, but for now this simply documents the workaround.

pragnya17:
The specific error that occurred was:
Encountered an error while generating hash for file` \sbom-tool\test\Microsoft.Sbom.Targets.Tests\bin\Debug\net8.0\TestResults\Microsoft.Sbom.Targets.Tests_net8.0_x64.log: The process cannot access the file 'C:\Users\ppandrate\Source\Repos\sbom-tool\test\Microsoft.Sbom.Targets.Tests\bin\Debug\net8.0\TestResults\Microsoft.Sbom.Targets.Tests_net8.0_x64.log' because it is being used by another process.`

By creating a temporary directory we can avoid this conflict when writing to the log file.

PR #957, opened 2025-03-04
Fix problems running E2E tests locally

DaveTryon:
As called out in #930, we have recently begun to have local E2E test failures. Upon investigation, the problem was determined that the test logs are being written to the folder we are using as the `BuildDropPath`, and that causes the SBOM validation to fail. This seems to be a recent change (bug?) in Visual Studio--we have been able to specify the `--results-directory` parameter when running from the command line, but attempts to change it when running from inside Visual Studio have proven unsuccessful. 

This PR works around the problem by copying the existing folder to a temporary folder before running the tests, then cleaning up the temporary folder after the tests complete. Since the `Microsoft.Sbom.Targets.Tests` project has both `net472` and `net80` targets, we need to ensure that each target has its own directory.

With these changes in place, I can successfully run all tests from inside Visual Studio, or by the command prompt via `dotnet test`, without needing to specify the `--results-directory` parameter.

The `AbstractGenerateSbomTaskInputTests.cs` and `AbstractGenerateSbomTaskTests.cs` tests were refactored to allow the names of the concrete classes to be passed into the abstract base classes. The old code used `CurrentDirectory`, which no longer seemed correct, so I renamed it to `TestBuildDropPath`. The SPDX 3.0 E2E tests will probably use these same abstract classes.

I'm not thrilled with having 3 copies of the `Xcopy` function. At one point I had it down to 2 copies (one per project), but once it was written, the extra overhead was worse than having the extra copy. Happy to entertain options here if people have strong feelings.


Microsoft.Sbom.sln
```diff
@@ -54,6 +54,9 @@ EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Sbom.Targets.Tests", "test\Microsoft.Sbom.Targets.Tests\Microsoft.Sbom.Targets.Tests.csproj", "{E31B914C-F24B-4DC8-ACC7-CAEA952563B8}"
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Sbom.Tool.Tests", "test\Microsoft.Sbom.Tool.Tests\Microsoft.Sbom.Tool.Tests.csproj", "{FC5A9799-7C44-4BFA-BA22-55DCAF1A1B9F}"
+	ProjectSection(ProjectDependencies) = postProject
```
DaveTryon, line 57: This section makes the E2E tests run _after_ the DotNetTool project is fully built. Without it, we have a race condition where the E2E tests and the DotNetTool project build can compete over a the NuGet package that the DotNetTool package produces. Changes were made through Visual Studio's project dependencies UI

JoseRenan (approved):

PR #963, opened 2025-03-07
Parsing SPDX 3.0 packages and validating with NTIA

pragnya17:
There was a small bug in the parsing and NTIA validation of SPDX 3.0 packages, causing validation to fail for SBOMs with packages when using a complianceStandard of NTIA. This PR fixes that bug and unit tests the package parsing and validation with NTIA.

### DIFF
--- a/src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Parser/SPDX30Parser.cs
+++ b/src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Parser/SPDX30Parser.cs
@@ -339,7 +339,7 @@ public class SPDX30Parser : ISbomParser
             var packageHasSha256Hash = packageElement.VerifiedUsing?.
                 Any(packageVerificationCode => packageVerificationCode.Algorithm == HashAlgorithm.sha256);
 
-            if (packageHasSha256Hash is null || packageHasSha256Hash is false)
+            if (packageHasSha256Hash is null || packageHasSha256Hash == false)
             {
                 throw new ParserException($"SBOM document is not NTIA compliant because package with SPDX ID {packageSpdxId} does not have a SHA256 hash.");
             }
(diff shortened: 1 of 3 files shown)


src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Parser/SPDX30Parser.cs
```diff
@@ -337,10 +336,10 @@ private void ValidateSbomPackagesForNTIA(List<Element> elementsList)
         {
             var packageSpdxId = packageElement.SpdxId;
 
-            var packageHasSha256Hash = packageElement.VerifiedUsing.
+            var packageHasSha256Hash = packageElement.VerifiedUsing?.
                 Any(packageVerificationCode => packageVerificationCode.Algorithm == HashAlgorithm.sha256);
 
-            if (!packageHasSha256Hash)
+            if (packageHasSha256Hash is null || packageHasSha256Hash is false)
```
DaveTryon, line 342: `is false` is a pattern that we haven't use elsewhere
```suggestion
            if (packageHasSha256Hash is null || !packageHasSha256Hash)
```

test/Microsoft.Sbom.Parsers.Spdx30SbomParser.Tests/Parser/JsonStrings/SbomFullDocWithPackagesStrings.cs
```diff
@@ -0,0 +1,716 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Sbom.Parser.JsonStrings;
+
+public static class SbomFullDocWithPackagesStrings
+{
+    [System.Diagnostics.CodeAnalysis.SuppressMessage("Style", "JSON002:Probable JSON string detected", Justification = "Need to use JSON string")]
+    public const string SbomPackageWithMissingVerificationJsonString =
+    @"
```
DaveTryon, line 10: During my DRI week, I may try to convert all of our inline json files to actual on-disk json files, both to make the tests smaller and to get rid of the double-quotes

DaveTryon (approved):

PR #964, opened 2025-03-07
Add runsettings file to avoid local UT failures

pragnya17:
This issue https://github.com/microsoft/sbom-tool/issues/930 documents local unit test failures caused by locked files. By adding a runsettings file we create a temp directory for local unit test results which avoids locked file issues.

### DIFF
--- a/.runsettings
+++ b/.runsettings
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <RunSettings>
     <RunConfiguration>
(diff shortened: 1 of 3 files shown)


.runsettings
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<RunSettings>
+    <RunConfiguration>
+        <ResultsDirectory>C:\Users\ppandrate\source\repos\sbom-tool\test\.\%temp%\sbom-tool-test-results</ResultsDirectory>
```
DaveTryon, line 4: This can't contain a hardcoded path

PR #973, opened 2025-03-13
Manifest info name should be case insensitive

pragnya17:
Existing code has an overriden `Equals()` method that returns true even if the manifestInfo has differing cases. This PR updates the manifestInfoValidator error message to indicate that and also adds unit tests to validate the behavior.

### DIFF
--- a/test/Microsoft.Sbom.Api.Tests/Config/Validators/ManifestInfoValidatorTests.cs
+++ b/test/Microsoft.Sbom.Api.Tests/Config/Validators/ManifestInfoValidatorTests.cs
@@ -19,7 +19,6 @@ public class ManifestInfoValidatorTests
     [DataRow("randomName", "2.2")]
     [DataRow("randomName", "3.0")]
     [DataRow("SPDX", "randomVersion")]
-    [DataRow("spdx", "randomVersion")]
     [DataRow("asodijf", "randomVersion")]
     [TestMethod]
     public void InvalidManifestInfoThrows(string name, string version)
@@ -36,14 +35,16 @@ public class ManifestInfoValidatorTests
         Assert.ThrowsException<ValidationArgException>(() => validator.ValidateInternal("property", listOfManifestInfos, null));
     }
 
-    [DataRow("2.2")]
-    [DataRow("3.0")]
+    [DataRow("SPDX", "2.2")]
+    [DataRow("SPDX", "3.0")]
+    [DataRow("spdx", "2.2")]
+    [DataRow("spdx", "3.0")]
     [TestMethod]
-    public void ValidManifestInfoPasses(string spdxVersion)
+    public void ValidManifestInfoPasses(string name, string spdxVersion)
     {
         var validManifestInfo = new ManifestInfo
         {
-            Name = "SPDX",
+            Name = name,
             Version = spdxVersion
         };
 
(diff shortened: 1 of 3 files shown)


test/Microsoft.Sbom.Api.Tests/Config/Validators/ManifestInfoValidatorTests.cs
```diff
@@ -19,7 +19,6 @@ public class ManifestInfoValidatorTests
     [DataRow("randomName", "2.2")]
     [DataRow("randomName", "3.0")]
     [DataRow("SPDX", "randomVersion")]
-    [DataRow("spdx", "randomVersion")]
```
jlperkins, line 22: Why remove this? Just curious

pragnya17: Lowercase spdx should not throw, it should be case-insensitive. So I've removed it from this unit test and added the case to the unit test below instead.

jlperkins: Ah - I think this test was looking at spdx + 'randomversion', which I believe should still throw. Like the test case above, with SPDX + 'randomversion', it should throw because of the version, not the name. 

Just for completeness, I'd keep it, but I guess I don't feel that strongly. But we should be clear on what this unit test is trying to do

jlperkins:
Do you mind linking to the Equals() call you reference in the description? Just for ease of review :)

test/Microsoft.Sbom.Api.Tests/Config/Validators/ManifestInfoValidatorTests.cs
DaveTryon: Positive tests should cover all happy paths--negative tests can often have just 1 case for each path and don't need to be exhaustive. Put another way, when the next SPDX format comes along, we would need to update the positive tests but not the negative tests.

It can also be a good idea to include a comment that explains the error should the test fail. This comment gets added to the `DataRow`, then used as the `message` parameter of the `Assert` method in question. If you were to combine these ideas for this test, you'd end up with the something like this:

```
    [DataRow("randomName", "2.2", "Unsupported Name should throw")]
    [DataRow("SPDX", "randomVersion", "Unsupported Version should throw")]
    [TestMethod]
    public void InvalidManifestInfoThrows(string name, string version, string message)
    {
        var invalidManifestInfo = new ManifestInfo
        {
            Name = name,
            Version = version
        };

        IList<ManifestInfo> listOfManifestInfos = new List<ManifestInfo> { invalidManifestInfo };

        var validator = new ManifestInfoValidator(mockAssemblyConfig.Object);
        Assert.ThrowsException<ValidationArgException>(() => validator.ValidateInternal("property", listOfManifestInfos, null), message);
    }
```

jlperkins: Makes sense, TIL negative tests don't need to be exhaustive. Thanks for the thorough answer!

pragnya17:
> Do you mind linking to the Equals() call you reference in the description? Just for ease of review :) Edit: nvm, [found it](https://github.com/microsoft/sbom-tool/blob/26bb1b00ca4b3e08a86336b65483a96c92d10487/src/Microsoft.Sbom.Contracts/Contracts/SbomSpecification.cs#L90)

Yup that's the one I was referring to

PR #977, opened 2025-03-17
Throw validation error if customer attempts to redact SPDX 3.0 SBOM

pragnya17:
Currently the SBOM tool only supports redaction for SPDX 2.2. This change throws a more descriptive error if the customer attempts to redact an SPDX 3.0 SBOM. Previously, the tool just threw a generic deserialization error. 

This PR also adds E2E tests to test the behavior and refactors some of the helper methods used for testing.

### DIFF
--- a/src/Microsoft.Sbom.Api/Workflows/SbomRedactionWorkflow.cs
+++ b/src/Microsoft.Sbom.Api/Workflows/SbomRedactionWorkflow.cs
@@ -57,7 +57,7 @@ public class SbomRedactionWorkflow : IWorkflow<SbomRedactionWorkflow>
                 var validationDetails = await validatedSbom.GetValidationResults();
                 if (validationDetails.Status != FormatValidationStatus.Valid)
                 {
-                    throw new InvalidDataException($"Failed to validate {sbomPath}:\n{string.Join('\n', validationDetails.Errors)}");
+                    throw new InvalidDataException($"Failed to validate {sbomPath}:\n{string.Join('\n', validationDetails.Errors)} Redaction is only supported for SPDX 2.2 currently, please provide a valid SPDX 2.2 SBOM.");
                 }
                 else
                 {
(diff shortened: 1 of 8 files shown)


test/Microsoft.Sbom.Tool.Tests/IntegrationTests.cs
```diff
@@ -229,9 +257,9 @@ private string CreateTestFolder()
         return testFolderPath;
     }
 
-    private static string AppendFullManifestFolderPath(string manifestDir)
+    private static string AppendFullManifestFolderPath(string manifestDir, string spdxVersion = "2.2")
```
DaveTryon, line 260: Since the caller's default for `spdxVersion` is null and you're checking it below, you can just use null here and only have 1 line that determines the default version.
```suggestion
    private static string AppendFullManifestFolderPath(string manifestDir, string spdxVersion = null)
```

src/Microsoft.Sbom.Api/Workflows/SbomRedactionWorkflow.cs
```diff
@@ -57,7 +57,7 @@ public virtual async Task<bool> RunAsync()
                 var validationDetails = await validatedSbom.GetValidationResults();
                 if (validationDetails.Status != FormatValidationStatus.Valid)
                 {
-                    throw new InvalidDataException($"Failed to validate {sbomPath}:\n{string.Join('\n', validationDetails.Errors)}");
+                    throw new InvalidDataException($"Failed to validate {sbomPath}:\n{string.Join('\n', validationDetails.Errors)} Redaction is only supported for SPDX 2.2 currently, please provide a valid SPDX 2.2 SBOM.");
```
DaveTryon, line 60: Grammar--let's use a period instead of a comma here
```suggestion
                    throw new InvalidDataException($"Failed to validate {sbomPath}:\n{string.Join('\n', validationDetails.Errors)} Redaction is only supported for SPDX 2.2 currently. Please provide a valid SPDX 2.2 SBOM.");
```

DaveTryon (approved):

PR #992, opened 2025-03-26
Validate compliance standard for SPDX 3.0

pragnya17:
Previous PR's added compliance standard validation for SPDX 3.0, however there were a few bugs and error handling issues that this PR targets.

The main outcomes from this PR are:
1. Compliance standard input is validated to make sure the only valid value is NTIA and it can only be used with >= SPDX:3.0
2. Errors are handled correctly and displayed in stdout
3. Better unit testing + E2E testing


src/Microsoft.Sbom.Extensions/ISbomParser.cs
```diff
@@ -42,5 +43,5 @@ public interface ISbomParser
     /// Returns the compliance standard set by the user, if it is valid.
     /// </summary>
     /// <param name="spdxVersion"></param>
-    public void SetComplianceStandard(string? complianceStandardFromCli);
```
DaveTryon, line 45: `public` keyword need to be there with latest Visual Studio

src/Microsoft.Sbom.Extensions/ISbomParser.cs
```diff
@@ -42,5 +43,5 @@ public interface ISbomParser
...
+    void SetComplianceStandard(ComplianceStandardType complianceStandard);
```
DaveTryon, line 46: `public` keyword is required for latest standards
```suggestion
    publci void SetComplianceStandard(ComplianceStandardType complianceStandard);
```

src/Microsoft.Sbom.Common/Config/IConfiguration.cs
```diff
@@ -211,7 +211,7 @@ public interface IConfiguration
     /// <summary>
     /// The compliance standard to validate against.
     /// </summary>
-    public ConfigurationSetting<string> ComplianceStandard { get; set; }
+    ConfigurationSetting<ComplianceStandardType> ComplianceStandard { get; set; }
```
DaveTryon, line 214: `public` keyword is required for latest Visual Studio
```suggestion
    public ConfigurationSetting<ComplianceStandardType> ComplianceStandard { get; set; }
```

src/Microsoft.Sbom.Contracts/Contracts/Enums/ComplianceStandardType.cs
```diff
@@ -0,0 +1,66 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Security.Cryptography;
```
DaveTryon, line 5: Not needed?
```suggestion
```

src/Microsoft.Sbom.Api/Utils/Constants.cs
```diff
@@ -47,6 +47,11 @@ public static class Constants
         SPDX30ManifestInfo
     };
 
+    public static Collection<ComplianceStandardType> SupportedComplianceStandards = new()
+    {
+        ComplianceStandardType.NTIA,
```
DaveTryon, line 52: Does this need to include `ComplianceStandardType.None` to accept "-cs None" on teh command line?

src/Microsoft.Sbom.Api/Workflows/SbomParserBasedValidationWorkflow.cs
```diff
@@ -22,6 +22,7 @@
 using Microsoft.Sbom.Extensions;
 using Microsoft.Sbom.JsonAsynchronousNodeKit;
 using Microsoft.Sbom.Parser;
+using Microsoft.Sbom.Parsers.Spdx30SbomParser.ComplianceStandard;
```
DaveTryon, line 25: The workflow should be version-agnostic. `ElementResult` and `InvalidElementInfo` should either be moved to a more central assembly or moved behind an abstraction that is more central.

src/Microsoft.Sbom.Api/Config/ConfigSanitizer.cs
```diff
@@ -225,6 +228,27 @@ private ConfigurationSetting<AlgorithmName> GetHashAlgorithmName(IConfiguration
...
+    private ConfigurationSetting<ComplianceStandardType> GetComplianceStandard(IConfiguration configuration)
+    {
+        // Convert to ComplianceStandard enum value.
+        var oldValue = configuration.ComplianceStandard;
+        var newValue = ComplianceStandardType.FromString(oldValue?.Value?.ToString());
+
+        // Compliance standard is only supported for ManifestInfo value of SPDX 3.0 and above.
+        if (!newValue.Equals(ComplianceStandardType.None) && !configuration.ManifestInfo.Value.Any(mi => mi.Equals(Constants.SPDX30ManifestInfo)))
+        {
+            throw new ValidationArgException($"Please use SPDX >=3.0 or remove the ComplianceStandard parameter.");
```
DaveTryon, line 240: This level of messaging will get hard to maintain going forward. Would it be enough to give them a message like this? $"Compliance standard {newValue} is not supported with manifest format {configuration.ManifestInfo.Value}. Please use a supported combination."

src/Microsoft.Sbom.Contracts/Contracts/Enums/ErrorType.cs
```diff
@@ -46,9 +46,15 @@ public enum ErrorType
...
+    /// </summary>
+    [EnumMember(Value = "Compliance standard validation error")]
+    ComplianceStandardError = 6,
+
     /// <summary>
     /// An unknown error occured.
     /// </summary>
     [EnumMember(Value = "Other")]
-    Other = 6,
+    Other = 7,
```
DaveTryon, line 59: Strictly speaking, the return value can be treated as part of our interface, so changing the "Other" error isn't something that we want to do frequently. Enum's don't need to be sequential, so let's bump this to 255, which is the largest return value supported by batch scripting.

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/ComplianceStandard/InvalidElementInfo.cs
```diff
@@ -0,0 +1,63 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.Sbom.Parsers.Spdx30SbomParser.ComplianceStandard.Enums;
+using Microsoft.Sbom.Parsers.Spdx30SbomParser.ComplianceStandard.Interfaces;
+
+namespace Microsoft.Sbom.Parsers.Spdx30SbomParser.ComplianceStandard;
+
+public class InvalidElementInfo
```
DaveTryon, line 9: As mentioned elsewhere, this probably should not be version-specific

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Parser/ElementsResult.cs
```diff
@@ -12,11 +13,39 @@ public record ElementsResult : ParserStateResult
     public ElementsResult(ParserStateResult result)
         : base(result.FieldName, result.Result, result.ExplicitField, result.YieldReturn)
     {
+        this.Elements = new List<Element>();
```
DaveTryon, line 16: As mentioned elsewhere, this probably shouldn't be specific to SPDX 3.0

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Parser/SPDX30Parser.cs
```diff
@@ -147,28 +153,18 @@ public SpdxMetadata GetMetadata()
...
-        else
-        {
-            throw new ParserException($"Unable to use the given compliance standard {complianceStandardFromCli} to parse the SBOM.");
+            case "NTIA":
+                RequiredComplianceStandard = complianceStandard;
+                entititesRequiringDeserializationWithNTIA = EntitiesWithDifferentNTIARequirements;
+                break;
+            case "None":
+                break;
+            default:
```
DaveTryon, line 166: Best practice: Even though we don't expect any other values to get here, some future change might add that possibility. Throwing from the default case provides a level of future-proofing

src/Microsoft.Sbom.Api/Config/ConfigSanitizer.cs
pragnya17: Yes I think that should be enough info.

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Parser/SPDX30Parser.cs
```diff
@@ -406,23 +403,23 @@ private SpdxMetadata SetMetadata(ElementsResult result)
...
-    /// <param name="elementsList"></param>
-    private void ValidateElementsBasedOnComplianceStandard(ComplianceStandard? complianceStandardAsEnum, List<Element> elementsList)
+    private void ValidateElementsBasedOnComplianceStandard(ComplianceStandardType complianceStandardAsEnum, ElementsResult elementsResult)
     {
-        switch (complianceStandardAsEnum)
+        switch (complianceStandardAsEnum.Name)
         {
-            case ComplianceStandard.NTIA:
-                ValidateNTIARequirements(elementsList);
+            case "NTIA":
```
DaveTryon, line 412: Code smell: When you have multiple places in the code that switch on the same thing, there's probably an opportunity to use a strategy pattern. For example, suppose we had an interface like this:
```
public interface IComplianceEnforcement
{
  public AddFailureElements(List<FileValidationResult> fileValidationFailures, HashSet<InvalidElementInfo> invalidElements);
  public string? AdjustEntityType(string? entityType);
  public void Validate(ElementsResult elementsResult);
};
```

You then switch once on the `ComplianceStandardType` to decide if you build either a `NoComplianceEnforcement` or an `NTIAComplianceEnforcement`, and you can remove all of the switches on compliance standard. As a bonus, you get small classes that are small, easy to test, and easy to reuse in multiple SPDX formats. If/when another standard is needed, you create the new enforcement class, add it to the one place that decides what type to generate, and it just works.

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Parser/SPDX30Parser.cs
pragnya17: Handled https://github.com/microsoft/sbom-tool/pull/993

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Parser/ElementsResult.cs
pragnya17: Handled https://github.com/microsoft/sbom-tool/pull/993

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/ComplianceStandard/InvalidElementInfo.cs
pragnya17: Handled in https://github.com/microsoft/sbom-tool/pull/993

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/ComplianceStandard/Enums/NTIAErrorType.cs
```diff
@@ -0,0 +1,50 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using Microsoft.Sbom.Parsers.Spdx30SbomParser.ComplianceStandard.Interfaces;
+
+namespace Microsoft.Sbom.Parsers.Spdx30SbomParser.ComplianceStandard.Enums;
+
+public class NTIAErrorType : IComplianceStandardErrorType, IEquatable<NTIAErrorType>
```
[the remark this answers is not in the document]
pragnya17: Handled in https://github.com/microsoft/sbom-tool/pull/993

src/Microsoft.Sbom.Api/Workflows/SbomParserBasedValidationWorkflow.cs
pragnya17: Handled in https://github.com/microsoft/sbom-tool/pull/993

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Parser/SPDX30Parser.cs
pragnya17: Handled in https://github.com/microsoft/sbom-tool/pull/993

src/Microsoft.Sbom.Api/Workflows/SbomParserBasedValidationWorkflow.cs
```diff
@@ -293,4 +309,47 @@ private void ThrowOnInvalidInputFiles(List<FileValidationResult> fileValidationF
...
+            case "None":
+                break;
+            default:
+                break;
+        }
+    }
+
+    private void AddInvalidNTIAElementsToFailures(List<FileValidationResult> fileValidationFailures, HashSet<InvalidElementInfo> invalidElements)
+    {
+        foreach (var invalidElementInfo in invalidElements)
```
DaveTryon, line 334: Suggestion: put a guard in this method, then just call the method from line 320. The guard would be something like:

```
if (configuration.ComplianceStandard?.Value?.Name != "NTIA")
{
  return;
}
```

src/Microsoft.Sbom.Api/Workflows/SbomParserBasedValidationWorkflow.cs
```diff
@@ -293,4 +309,47 @@ private void ThrowOnInvalidInputFiles(List<FileValidationResult> fileValidationF
...
+            {
+                Path = invalidElementInfo.ToString(),
+                ErrorType = ErrorType.ComplianceStandardError,
+            });
+        }
+    }
+
+    private bool NoOpComplianceStandard(ConfigurationSetting<Contracts.Enums.ComplianceStandardType> complianceStandard)
+    {
+        if (complianceStandard?.Value?.Name == "None")
```
DaveTryon, line 346:
```suggestion
        return complianceStandard?.Value?.Name == "None";
```

src/Microsoft.Sbom.Common/Constants.cs
```diff
@@ -20,4 +20,7 @@ public static class Constants
 
     public const string DefaultManifestInfoName = "SPDX";
     public const string DefaultManifestInfoVersion = "2.2";
+
```
DaveTryon, line 23: I'm guessing that this extra newline is unintended?
```suggestion
```

test/Microsoft.Sbom.Api.Tests/Workflows/SbomParserBasedValidationWorkflowTests.cs
```diff
@@ -390,4 +392,147 @@ public async Task SbomParserBasedValidationWorkflowTests_ReturnsSuccessAndValida
         fileSystemMock.VerifyAll();
         osUtilsMock.VerifyAll();
     }
+
+    [TestMethod]
+    public async Task SbomParserBasedValidationWorkflowTests_ReturnsNTIAValidationFailures_Succeeds()
+    {
+        var manifestInfo = Constants.SPDX30ManifestInfo;
```
DaveTryon, line 399: Not in this PR, but we _desperately_ need to rethink these tests. No test should take this much setup.

DaveTryon (approved):
Made a couple of minor suggestions--use or ignore as you see fit

test/Microsoft.Sbom.Api.Tests/Workflows/SbomParserBasedValidationWorkflowTests.cs
pragnya17: Agreed I'll create an issue/work item for it

src/Microsoft.Sbom.Common/Constants.cs
pragnya17: I added it to separate the sections, following the format already in place

src/Microsoft.Sbom.Api/Workflows/SbomParserBasedValidationWorkflow.cs
pragnya17: I'm hesitant to do this because I want to keep the switch case for future changes that might include other compliance standards. That way it is very easy to see all the places where you just need to another case for the new compliance standard. This switch case pattern is also utilized in other places in the code, so I'd rather keep it consistent by using it here too.

DaveTryon (approved):

test/Microsoft.Sbom.Tool.Tests/IntegrationTests.cs
```diff
@@ -270,12 +355,13 @@ private void GenerateManifestAndValidateSuccess(string testFolderPath, string ma
...
     }
 
-    private (string arguments, string outputFile) GetValidateManifestArguments(string testFolderPath, string manifestInfoValue = "SPDX:2.2")
+    private (string arguments, string outputFile) GetValidateManifestArguments(string testFolderPath, string manifestInfoValue = "2.2", string complianceStandardValue = "")
     {
         var outputFile = Path.Combine(TestContext.TestRunDirectory, TestContext.TestName, "validation.json");
         var manifestRootFolderName = Path.Combine(testFolderPath, ManifestRootFolderName);
         var arguments = $"validate -m \"{manifestRootFolderName}\" -b \"{testDropDirectory}\" -o \"{outputFile}\"";
-        arguments += $" -mi {manifestInfoValue}";
+        arguments += string.IsNullOrEmpty(manifestInfoValue) ? string.Empty : $" -mi SPDX:{manifestInfoValue}";
```
DaveTryon, line 363: Rather than mixing interpolation and concatenation, can we use all interpolation? Something like this should work:
```
        var manifestInfoArg = string.IsNullOrEmpty(manifestInfoValue) ? string.Empty : $"-mi SPDX:{manifestInfoValue}";
        var complianceArg = string.IsNullOrEmpty(complianceStandardValue) ? string.Empty : $"-cs {complianceStandardValue}";
        var arguments = $"validate -m \"{manifestRootFolderName}\" -b \"{testDropDirectory}\" -o \"{outputFile}\" {manifestInfoArg} {complianceArg}";
```

DaveTryon (approved):

Issue #962, opened 2025-03-07
SBOM generation for SPDX 3.0 is missing files


pragnya17:
There are certain files missing from SPDX 3.0 SBOMs. Need to investigate what types of files are missing. The validation for SPDX 3.0 SBOMs works, so the validator can be used to find which types of files are missing.

pragnya17:
https://github.com/microsoft/sbom-tool/pull/997 fixes this issue and explains the root cause.

PR #997, opened 2025-04-02
Fix SPDX 3.0 manifest missing files bug

pragnya17:
This PR is the fix for issue https://github.com/microsoft/sbom-tool/issues/962. The root cause was that during the generation of SPDX file elements, the SPDX id was not generated using the file hash. This meant that the SPDX id was not unique for each file, which meant that certain files were incorrectly identified as duplicates. Then during deduplication these incorrectly identified "duplicates" were not added to the manifest file. 

This PR fixes the logic for adding SPDX id's to SPDX 3.0 file elements and validates the fix with unit tests and E2E tests.

### DIFF
--- a/src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Utils/SPDXExtensions.cs
+++ b/src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Utils/SPDXExtensions.cs
@@ -31,7 +31,11 @@ public static class SPDXExtensions
     /// </summary>
     public static string GenerateSpdxId(Element element, string id) => $"SPDXRef-{element.Type}-{GetStringHash(id)}";
 
-    public static string GetSpdxId(Element element, string id) => $"SPDXRef{element.Type}-{element.Name}-{id}";
+    public static string GetSpdxId(Element element, string id)
+    {
+        var spdxFileId = $"{Constants.SPDXRefFile}-{element.Name}-{id}";
+        return SpdxIdAllowedCharsRegex.Replace(spdxFileId, "-");
+    }
 
     /// <summary>
     /// Returns the SPDX-compliant external document ID.
--- a/src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Utils/SPDXExtensions.cs
+++ b/src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Utils/SPDXExtensions.cs
@@ -31,6 +31,8 @@ public static class SPDXExtensions
     /// </summary>
     public static string GenerateSpdxId(Element element, string id) => $"SPDXRef-{element.Type}-{GetStringHash(id)}";
 
+    public static string GetSpdxId(Element element, string id) => $"SPDXRef{element.Type}-{element.Name}-{id}";
+
     /// <summary>
     /// Returns the SPDX-compliant external document ID.
     /// </summary>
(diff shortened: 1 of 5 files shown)


test/Microsoft.Sbom.Tool.Tests/IntegrationTests.cs
```diff
@@ -226,18 +248,15 @@ public void E2E_Redact_WithManifestInfo_ReturnsNonZeroExitCode(string manifestIn
...
 
-    private void GenerateManifestAndValidateSuccess(string testFolderPath, string manifestInfoValue = null)
+    private void GenerateManifestAndValidateSuccess(string testFolderPath, string manifestInfoSpdxVersion = null)
     {
         var arguments = $"generate -ps IntegrationTests -pn IntegrationTests -pv 1.2.3 -m \"{testFolderPath}\"  -b \"{testDropDirectory}\" -bc \"{GetSolutionFolderPath()}\"";
-        if (!string.IsNullOrEmpty(manifestInfoValue))
-        {
-            arguments += $" -mi SPDX:{manifestInfoValue}";
-        }
+        arguments += !string.IsNullOrEmpty(manifestInfoSpdxVersion) ? $" -mi SPDX:{manifestInfoSpdxVersion}" : string.Empty;
```
DaveTryon, line 254: This is a funny mix of interpolation and concatenation. How would you feel about making it all concatenation? One option:
```
var manifestInfoArg = string.IsNullOrEmpty(manifestInfoSpdxVersion) ? string.Empty : $"-mi SPDX:{manifestInfoSpdxVersion}" ;
var arguments = $"generate -ps IntegrationTests -pn IntegrationTests -pv 1.2.3 -m \"{testFolderPath}\" -b \"{testDropDirectory}\" -bc \"{GetSolutionFolderPath()}\" {manifestInfoArg}";
```

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Utils/SPDXExtensions.cs
```diff
@@ -31,6 +31,12 @@ public static class SPDXExtensions
     /// </summary>
     public static string GenerateSpdxId(Element element, string id) => $"SPDXRef-{element.Type}-{GetStringHash(id)}";
 
+    public static string GetSpdxId(Element element, string id)
```
DaveTryon, line 34: These public methods with no tests and no parameter validation are a bit concerning. I think we can make this class and all its `public` methods `internal`, but I'm not 100% sure how that will fit with #992

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Utils/SPDXExtensions.cs
```diff
@@ -31,6 +31,12 @@ public static class SPDXExtensions
     /// </summary>
     public static string GenerateSpdxId(Element element, string id) => $"SPDXRef-{element.Type}-{GetStringHash(id)}";
```
DaveTryon, line 32: General comment--only some of the methods in this class are extension methods (i.e., they use the `this` keyword on the first parameter). It might be good to make at least the new method (not necessarily the old ones) actually be extension methods

pragnya17: I will address this in a separate PR

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Utils/SPDXExtensions.cs
pragnya17: I will address this in a separate PR

test/Microsoft.Sbom.Tool.Tests/IntegrationTests.cs
pragnya17: Makes sense to me, I made the change.

DaveTryon (approved):

PR #998, opened 2025-04-02
Move spdx extensions to common utils and refactor SPDX 2.2

pragnya17:
As part of a larger effort to refactor the SPDX extensions in SPDX 2.2 and SPDX 3.0, this PR aims to move the non-extension method to a common utility class. It also removes the duplicates from the SPDX 2.2 project and utilizes the common utility methods. Future PRs will target refactoring for SPDX 3.0.


src/Microsoft.Sbom.Common/Utils/CommonSPDXUtils.cs
```diff
@@ -0,0 +1,56 @@
...
+using System.Security.Cryptography;
+using System.Text;
+using System.Text.RegularExpressions;
+
+namespace Microsoft.Sbom.Common.Utils;
+
+/// <summary>
+/// Provides utility methods for SPDX objects.
+/// </summary>
+public class CommonSPDXUtils
```
DaveTryon, line 14: This class holds no state and could presumably be `static`

src/Microsoft.Sbom.Common/Utils/CommonSPDXUtils.cs
```diff
@@ -0,0 +1,56 @@
...
+    {
+        var spdxExternalDocumentId = $"{Constants.SPDXRefExternalDocument}-{fileName}-{sha1Value}";
+        return SpdxIdAllowedCharsRegex.Replace(spdxExternalDocumentId, "-");
+    }
+
+    /// <summary>
+    /// Compute the SHA256 string representation (omitting dashes) of a given string
+    /// </summary>
+    /// <remarks>
+    /// TODO:  refactor this into Core as similar functionality is duplicated in a few different places in the codebase
```
DaveTryon, line 48: Does this PR complete the TODO ?

pragnya17: This comment was from someone else. I assume that by "Core" they meant Common since we don't have a Core project, so I think this TODO is resolved with this refactor. If you agree as well, I can remove it.

test/Microsoft.Sbom.Common.Tests/CommonSPDXUtilsTests.cs
```diff
@@ -0,0 +1,97 @@
...
+
+    [TestMethod]
+    public void GenerateSpdxFileIdTest()
+    {
+        var spdxId = CommonSPDXUtils.GenerateSpdxFileId("test", "sha1Value");
+        Assert.IsTrue(spdxId.Contains(Constants.SPDXRefFile));
+    }
+
+    [TestMethod]
+    public void GenerateSpdxFileIdTest_SameNameAndSha1Value()
```
ZhengHong-Tan, line 44: would it be possible to use the `[DataRow]` attribute here? I was thinking we could put the different combinations under the same test method while using something like: 

```
[DataRow("test1", "sha1value", "test1", "sha1Value", true)]
...
```

test/Microsoft.Sbom.Common.Tests/CommonSPDXUtilsTests.cs
```diff
@@ -0,0 +1,97 @@
...
+
+namespace Microsoft.Sbom.Parser;
+
+using Microsoft.Sbom.Common;
+using Microsoft.Sbom.Common.Utils;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+[TestClass]
+public class CommonSPDXUtilsTests
+{
```
DaveTryon, line 12: Since the hash values are known, we can improve the strength of the test pinning and make the code more readable by using consts. Something like:
```
const string packageId1 = "package1";
const string packageId2 = "package2";
const string packageId2 = "package3";
const string spdxOfPackage1 = "whatever the string value is";
const string spdxOfPackage2 = "whatever the string value is";
const string spdxOfPackage3 = "whatever the string value is";
```

Then the tests could use the named values, something like
```
[TestMethod]
[DataRow(packageId1, spdxOfPackageId1)]
[DataRow(packageId2, spdxOfPackageId2)]
[DataRow(packageId3, spdxOfPackageId3)]
public void GenerateSpdxPackageId_OutputIsCorrect(string input, string expectedOutput)
{
    Assert.AreEqual(expectedOutput, CommonSPDXUtils.GenerateSpdxPackageId(input));
}
```

You could follow a similar pattern for the `GenerateSpdxFileId` method

test/Microsoft.Sbom.Common.Tests/CommonSPDXUtilsTests.cs
```diff
@@ -0,0 +1,97 @@
...
+    [TestMethod]
+    public void GenerateSpdxPackageIdTest()
+    {
+        var spdxId = CommonSPDXUtils.GenerateSpdxPackageId("test");
+        Assert.IsTrue(spdxId.Contains(Constants.SPDXRefPackage));
```
DaveTryon, line 17: General note about Assert.IsTrue and Assert.IsNotTrue: If your unit test fails, the logs provide no hints to help you know why it failed. Methods like Assert.AreEqual or Assert.AreNotEqual are more informative, since they log the strings in question. If you have a scenario where you really need a boolean check, it's a good idea to use the description parameter to provide more info about the context. In this case, I'd use AreEqual, but let's assume for the sake of argument that the output was different each time--you could still glean more info with something like this:
```
Assert.IsTrue(spdxId.Contains("blah"), $"spdxId: {spdxId}");
```

test/Microsoft.Sbom.Common.Tests/CommonSPDXUtilsTests.cs
DaveTryon: +1 to DataRow for "values should match" tests. The "values should not match" would probably be a separate test method, which could be done with or without DataRow, depending on how you want to structure it

pragnya17: There's only one possible combo for generating the same SPDX ID, however there are multiple possible combos for generating different ones so I can include a data row for those unit tests.

test/Microsoft.Sbom.Common.Tests/CommonSPDXUtilsTests.cs
```diff
@@ -0,0 +1,122 @@
...
+    [DataRow(TestPackage2, TestPackage2SpdxId)]
+    [TestMethod]
+    public void GenerateSpdxPackageIdTest(string package, string expectedSpdxId)
+    {
+        var spdxId = CommonSPDXUtils.GenerateSpdxPackageId(package);
+        Assert.AreEqual(expectedSpdxId, spdxId);
+    }
+
+    [TestMethod]
+    public void GenerateSpdxPackageIdTest_SameStrings()
```
DaveTryon, line 42: Is this method needed? We're already comparing the output to a constant in the previous test

test/Microsoft.Sbom.Common.Tests/CommonSPDXUtilsTests.cs
```diff
@@ -0,0 +1,122 @@
...
+    {
+        var spdxId1 = CommonSPDXUtils.GenerateSpdxPackageId(TestPackage1);
+        var spdxId2 = CommonSPDXUtils.GenerateSpdxPackageId(TestPackage1);
+        Assert.AreEqual(spdxId1, spdxId2);
+    }
+
+    [TestMethod]
+    public void GenerateSpdxPackageIdTest_DiffStrings()
```
DaveTryon, line 50: Again, maybe not needed since we're comparing to const strings? I suppose we could add a test that the constant strings are different, just in case someone ever decided to edit them, but that could be a bit of a stretch

test/Microsoft.Sbom.Common.Tests/CommonSPDXUtilsTests.cs
```diff
@@ -0,0 +1,122 @@
...
+    [DataRow(TestFile2, TestHash2, TestFile2Hash2SpdxId)]
+    [TestMethod]
+    public void GenerateSpdxFileIdTest(string file, string sha1Value, string expectedSpdxId)
+    {
+        var spdxId = CommonSPDXUtils.GenerateSpdxFileId(file, sha1Value);
+        Assert.AreEqual(expectedSpdxId, spdxId);
+    }
+
+    [TestMethod]
+    public void GenerateSpdxFileIdTest_SameNameAndSha1Value()
```
DaveTryon, line 67: Not needed, since we're checking against constant values?

test/Microsoft.Sbom.Common.Tests/CommonSPDXUtilsTests.cs
```diff
@@ -0,0 +1,122 @@
...
+    {
+        var spdxId1 = CommonSPDXUtils.GenerateSpdxFileId(TestFile1, TestHash1);
+        var spdxId2 = CommonSPDXUtils.GenerateSpdxFileId(TestFile1, TestHash1);
+        Assert.AreEqual(spdxId1, spdxId2);
+    }
+
+    [TestMethod]
+    public void GenerateSpdxFileIdTest_DiffName()
```
DaveTryon, line 75: Not needed, since we're checking against constant values?

test/Microsoft.Sbom.Common.Tests/CommonSPDXUtilsTests.cs
```diff
@@ -0,0 +1,122 @@
...
+    [DataRow(TestExternalDoc2, TestHash2, TestExternalDoc2Hash2SpdxId)]
+    [TestMethod]
+    public void GenerateSpdxExternalDocumentIdTest(string externalDoc, string sha1value, string expectedSpdxId)
+    {
+        var spdxId = CommonSPDXUtils.GenerateSpdxExternalDocumentId(externalDoc, sha1value);
+        Assert.AreEqual(expectedSpdxId, spdxId);
+    }
+
+    [TestMethod]
+    public void GenerateSpdxExternalDocumentIdTest_SameNameAndSha1Value()
```
DaveTryon, line 100: Not needed, since we're checking against constant values?

DaveTryon (approved):
Now that we're using constants for the comparison, we might be able to remove some tests

DaveTryon (approved):

PR #999, opened 2025-04-02
Introduce functionality to validate signatures on non-Windows machines

ZhengHong-Tan:
This PR contains the functionality to enable signature validation of the manifest.cat catalog file and manifest json files on non-Windows machines. We use the signtool.exe tool and our own implementation to validate the signatures on Windows machines but let the signtool.exe tool decide the outcome of the validation. We then record the validation results of both implementations on the telemetry file.


DaveTryon:
As we discussed offline, the signing validation code should probably be in the internal repo, since that's where our implementation of `ISignValidator` already lives. We may need to extend the `IRecorder` interface to add validation detail to the telemetry stream. I've sent you a link to the branch we worked on as a starting point.

PR #1001, opened 2025-04-04
Refactor SPDX 3.0 extension methods

pragnya17:
This PR builds off of https://github.com/microsoft/sbom-tool/pull/998 by refactoring the SPDX 3.0 extension methods during generation of a SPDX 3.0 SBOM.


src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Utils/SPDXExtensions.cs
```diff
@@ -19,158 +19,130 @@ namespace Microsoft.Sbom.Parsers.Spdx30SbomParser.Utils;
...
         }
 
-        // Get the SHA1 for this file.
-        var sha1Value = sha1checksums.FirstOrDefault().ChecksumValue;
+        spdxPackage.SpdxId = CommonSPDXUtils.GenerateSpdxPackageId(packageInfo.Id ?? packageIdentity);
+        return spdxPackage.SpdxId;
+    }
 
-        reference.ExternalSpdxId = GenerateSpdxExternalDocumentId(name, sha1Value);
+    public static void AddSpdxId(this CreationInfo creationInfo)
```
DaveTryon, line 98: The reordering of these methods makes it a lot harder to compare the old version to the new version. Is it possible to restore to the original order to simplify the review?

test/Microsoft.Sbom.Parsers.Spdx30SbomParser.Tests/Utils/SPDXExtensionsTest.cs
```diff
@@ -0,0 +1,208 @@
...
+
+    [TestMethod]
+    public void AddSpdxIdToPackage_PackageInfoIsNull_Throws()
+    {
+        Assert.ThrowsException<ArgumentNullException>(() => spdxPackage.AddSpdxId(null));
+    }
+
+    [DataRow(true)]
+    [DataRow(false)]
+    [TestMethod]
```
DaveTryon, line 91: `[DataRow]` here seems like an odd choice. I'd lean toward separate tests

test/Microsoft.Sbom.Parsers.Spdx30SbomParser.Tests/Utils/SPDXExtensionsTest.cs
```diff
@@ -0,0 +1,208 @@
...
+        {
+            spdxPackage.AddSpdxId(packageInfo);
+        }
+        else
+        {
+            packageInfo.PackageVersion = null;
+            spdxPackage.AddSpdxId(packageInfo);
+        }
+
+        Assert.AreEqual(PackageSpdxIdWithPackageId, spdxPackage.SpdxId);
```
DaveTryon, line 104: Wait... shouldn't the SpdxId change if PackageVersion changes?

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Utils/SPDXExtensions.cs
```diff
@@ -19,158 +19,130 @@ namespace Microsoft.Sbom.Parsers.Spdx30SbomParser.Utils;
...
-    /// <param name="spdxPackage"></param>
-    /// <param name="packageInfo"></param>
-    /// <returns>Package ID that encapsulates unique info about a package.</returns>
-    public static string GetSpdxElementId(SbomPackage packageInfo)
-    {
-        if (packageInfo is null)
+        if (string.IsNullOrEmpty(name))
         {
-            throw new ArgumentNullException(nameof(packageInfo));
+            throw new ArgumentException($"'{nameof(name)}' cannot be null or empty.", nameof(name));
```
DaveTryon, line 36: ArgumentException will already include the parameter name. This can be simplified to
```suggestion
            throw new ArgumentException("Parameter cannot be null or empty.", nameof(name));
```

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Utils/SPDXExtensions.cs
pragnya17: I've reordered them, could you take a look and see if that helps?

test/Microsoft.Sbom.Parsers.Spdx30SbomParser.Tests/Utils/SPDXExtensionsTest.cs
pragnya17: No, per my understanding - If the package id is provided, then the package version is not used when creating the SPDX id. So regardless of the version, the SPDX id should be the same if the package id is the same. I also cross checked that SPDX 2.2 has the same behavior.

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Generator.cs
```diff
@@ -230,7 +230,6 @@ public GenerationResult GenerateRootPackage(IInternalMetadataProvider internalMe
         };
 
         spdxSupplier.AddSpdxId();
-        spdxPackage.AddSpdxId();
```
DaveTryon, line 233: Does `spdxPackage` no longer need an SpdxId ?

DaveTryon, line 233: Not sure why this is removed--was it duplicated somewhere else?

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Generator.cs
```diff
@@ -296,7 +295,6 @@ public GenerationResult GenerateJsonDocument(ExternalDocumentReferenceInfo exter
         };
 
         spdxExternalMap.AddExternalSpdxId(externalDocumentReferenceInfo.ExternalDocumentName, externalDocumentReferenceInfo.Checksum);
-        spdxExternalMap.AddSpdxId();
         var externalDocumentReferenceId = spdxExternalMap.ExternalSpdxId;
```
DaveTryon, line 299: Why was this removed--was it redundant?

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Generator.cs
pragnya17: For root packages, we set the SPDX ID differently to indicate it's a root package. I do set the SPDX ID initially using a constant value on line 205. This was an unintentional duplicate that overwrote the constant SPDX ID.

DaveTryon (dismissed):

DaveTryon (approved):

PR #1002, opened 2025-04-04
Introduce new telemetry method to record signature validation results

ZhengHong-Tan:
The changes to the interface allow for recording additional results in the telemetry file in the form of a key-value pair of strings.

### DIFF
--- a/src/Microsoft.Sbom.Api/Output/Telemetry/IRecorder.cs
+++ b/src/Microsoft.Sbom.Api/Output/Telemetry/IRecorder.cs
@@ -90,9 +90,9 @@ public interface IRecorder
     public Task FinalizeAndLogTelemetryAsync();
 
     /// <summary>
-    /// Add an extra key/value pair to the telemetry
+    /// Add an extra result in the form of a key-value pair to the telemetry.
     /// </summary>
-    public void AddProperty(string propertyName, string value);
+    public void AddResult(string propertyName, string value);
 
     public IList<FileValidationResult> Errors { get; }
 }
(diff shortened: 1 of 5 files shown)


DaveTryon (commented):
@ZhengHong-Tan, can you please add some unit tests for the `TelemetryRecorder` class ? These were never added previously, so we don't necessarily need to be exhaustive, but can we at least pin the new behavior?

src/Microsoft.Sbom.Api/Output/Telemetry/TelemetryRecorder.cs
```diff
@@ -102,6 +103,7 @@ public async Task LogException(Exception exception)
                 Switches = this.switches,
                 Parameters = Configuration,
                 Exceptions = exceptionList.ToDictionary(k => k.GetType().ToString(), v => v.Message),
+                AdditionalResults = additionalResults,
```
DaveTryon, line 106: This is the error path. The success path should also be updated. See the `SbomTelemetry` object created on line 310

ZhengHong-Tan: updated the success path, pushing changes in the next commit.

src/Microsoft.Sbom.Api/Output/Telemetry/IRecorder.cs
```diff
@@ -89,5 +89,10 @@ public interface IRecorder
     /// </summary>
     public Task FinalizeAndLogTelemetryAsync();
 
+    /// <summary>
+    /// Add an extra result in the form of a key-value pair to the telemetry.
+    /// </summary>
+    public void AddResult(string propertyName, string value);
```
alisonlomaka, line 95: Since the interface is changing will this be a breaking change requiring a major version bump? If so this might be a good one to bring up at standup for awareness and planning.

alisonlomaka (approved):

PR #1005, opened 2025-04-09
Write E2E tests for validation success and failure (SPDX 2.2 and 3.0)

pragnya17:
This PR adds E2E tests for validation successes and failures for both SPDX 2.2 and 3.0. The failure cases that are tested include `Additional files not in the manifest` and `Files in the manifest missing from the disk`. The other 2 types of validation errors are `Unknown file errors` and `Files with invalid hashes` but these are not easily reproducible so they are omitted from E2E testing.

### DIFF
--- a/test/Microsoft.Sbom.Tool.Tests/IntegrationTests.cs
+++ b/test/Microsoft.Sbom.Tool.Tests/IntegrationTests.cs
@@ -9,6 +9,8 @@ using System.Reflection;
 using System.Runtime.InteropServices;
 using System.Text.Json;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Newtonsoft.Json;
+using JsonSerializer = System.Text.Json.JsonSerializer;
 
 namespace Microsoft.Sbom.Tools.Tests;
 
@@ -308,10 +310,8 @@ public class IntegrationTests
         Assert.IsTrue(stdout.Contains("SPDXRef-Package"));
     }
 
-    [DataRow("2.2")]
-    [DataRow("3.0")]
     [TestMethod]
-    public void E2E_Validate_WithAdditionalFile_StdOutContainsAdditionalFileErrors(string version)
+    public void E2E_Validate_WithAdditionalFile_SPDX22_StdOutContainsAdditionalFileErrors()
     {
         if (!IsWindows)
         {
@@ -319,6 +319,7 @@ public class IntegrationTests
             return;
         }
 
+        var version = "2.2";
         var testFolderPath = CreateTestFolder();
         GenerateManifestAndValidateSuccess(testFolderPath, manifestInfoSpdxVersion: version);
 
@@ -329,6 +330,7 @@ public class IntegrationTests
         var (arguments, outputFile) = GetValidateManifestArguments(testFolderPath, manifestInfoSpdxVersion: version);
 
         var (stdout, stderr, exitCode) = LaunchAndCaptureOutput(arguments);
+        Assert.AreEqual(3, exitCode.Value, $"Unexpected failure: stdout = {stdout}");
         Assert.IsTrue(stdout.Contains("Additional files not in the manifest . . . . . . 1"));
         Assert.IsTrue(stdout.Contains("newFile"));
 
@@ -338,10 +340,8 @@ public class IntegrationTests
         }
     }
 
-    [DataRow("2.2")]
-    [DataRow("3.0")]
     [TestMethod]
-    public void E2E_Validate_WithMissingFile_StdOutContainsMissingFileErrors(string version)
+    public void E2E_Validate_WithAdditionalFile_SPDX30_StdOutContainsAdditionalFileErrors()
     {
         if (!IsWindows)
         {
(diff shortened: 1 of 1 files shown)


test/Microsoft.Sbom.Tool.Tests/IntegrationTests.cs
```diff
@@ -307,6 +308,81 @@ public void E2E_Validate_WithValidComplianceStandard_SupportedManifestInfo_StdOu
...
+                    name = "newFile",
+                    verifiedUsing = new[] { new { algorithm = "sha1", hashValue = "abc123", creationInfo = "_:creationInfo", spdxId = "SPDXRef-PackVerCode", type = "PackageVerificationCode" } },
+                    spdxId = "SPDXRef-File-newFile",
+                    creationInfo = "_:creationInfo",
+                    type = "software_File",
+                };
+                break;
+        }
+
+        AddExtraFileToManifest(testFolderPath, version, newFile);
```
ZhengHong-Tan, line 378: would it make sense to have this method be part of a class? like maybe we can have 2 test classes: `Spdx22TestClass.cs` and `Spdx30TestClass.cs` - I can't think of a better name, but feel free to use a better one! I'm generally thinking that it would be easier in the future if we have to remove 2.2 support, then we could just delete those test classes and any calls to the methods in this ever-growing `IntegrationTests.cs` file. Let me know if you have any questions or if you would like to pair on it!

pragnya17: I definitely agree that these could probably be in separate classes. In that case, I think it would also be worth refactoring all the other SPDX 3.0 E2E tests in this file. This would be a larger change though and I hesitate to include that refactoring in this PR too. I think I could pick this up in a separate PR if you also agree?

test/Microsoft.Sbom.Tool.Tests/IntegrationTests.cs
```diff
@@ -347,19 +423,14 @@ private void GenerateManifestAndValidateSuccess(string testFolderPath, string ma
         Assert.IsTrue(File.Exists(jsonFilePath));
         Assert.IsTrue(File.Exists(shaFilePath));
 
-        // Check that manifestFolderPath is the only folder in the directory
-        var directories = Directory.GetDirectories(Path.Combine(testFolderPath, ManifestRootFolderName));
-        Assert.AreEqual(1, directories.Length, "There should be only one folder in the test directory.");
```
sfoslund, line 352: Why remove this check? I assume now we expect for there to be 2 dirs (one for 2.2 and one for 3.0), is that right? If so, let's assert that.

test/Microsoft.Sbom.Tool.Tests/IntegrationTests.cs
```diff
@@ -307,6 +308,81 @@ public void E2E_Validate_WithValidComplianceStandard_SupportedManifestInfo_StdOu
...
+        var testFolderPath = CreateTestFolder();
+        GenerateManifestAndValidateSuccess(testFolderPath, manifestInfoSpdxVersion: version);
+
+        // Add additional file
+        var newFilePath = Path.Combine(testDropDirectory, "newFile");
+        File.WriteAllText(newFilePath, "This is a test file.");
+
+        var (arguments, outputFile) = GetValidateManifestArguments(testFolderPath, manifestInfoSpdxVersion: version);
+
+        var (stdout, stderr, exitCode) = LaunchAndCaptureOutput(arguments);
```
sfoslund, line 331: Can we assert that the exitCode is what we expect here too, and in the other test?

test/Microsoft.Sbom.Tool.Tests/IntegrationTests.cs
```diff
@@ -398,6 +469,67 @@ private static string GetAppName()
...
+
+        var jsonContent = File.ReadAllText(jsonFilePath);
+        var jsonDocument = JsonDocument.Parse(jsonContent);
+        var rootElement = jsonDocument.RootElement;
+        var filesArray = rootElement.GetProperty(arrayToModify).EnumerateArray().ToList();
+
+        // Add a new file to manifest.spdx.json
+        filesArray.Add(JsonSerializer.SerializeToElement(newFile));
+
+        // Create a new JSON object with the updated "files" array
```
sfoslund, line 498: This logic seems complicated if we're basically just trying to modify a object, is there a way to simplify?

test/Microsoft.Sbom.Tool.Tests/IntegrationTests.cs
pragnya17: Turns out the reason `_manifest` had `spdx_2.2` and `spdx_3.0` was because of the DataRow on this unit test which meant the folder structure was not cleaned up properly. I've separated the tests and added back the assertion that there should only be one folder within `_manifest`.

test/Microsoft.Sbom.Tool.Tests/IntegrationTests.cs
pragnya17: Simplified in latest commit

sfoslund (approved):

PR #1011, opened 2025-04-10
SBOM content diff checker between SPDX 2.2 and SPDX 3.0

pragnya17:
This PR aims to write a diff checker to compare the content equality between an SPDX 2.2 SBOM and SPDX 3.0 SBOM.

### DIFF
--- a/src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Utils/ElementSerializer.cs
+++ b/src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Utils/ElementSerializer.cs
@@ -9,8 +9,64 @@ using Microsoft.Sbom.Common.Spdx30Entities;
 
 public class ElementSerializer : JsonConverter<List<Element>>
 {
-    public override List<Element> Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options) =>
-        throw new NotImplementedException("Deserialization of Elements into specific subtypes is not implemented yet.");
+    public override List<Element> Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
+    {
+        if (reader.TokenType != JsonTokenType.StartArray)
+        {
+            throw new JsonException("Expected a JSON array.");
+        }
+
+        var elements = new List<Element>();
+
+        // Read the start of the array
+        reader.Read();
+
+        while (reader.TokenType != JsonTokenType.EndArray)
+        {
+            // Create a JsonDocument for the current element
+            using var jsonDocument = JsonDocument.ParseValue(ref reader);
+            var jsonObject = jsonDocument.RootElement;
+
+            // Determine the type of the element
+            if (!jsonObject.TryGetProperty("type", out var typeProperty))
+            {
+                throw new JsonException("Missing 'type' property in JSON element.");
+            }
+
+            var typeValue = typeProperty.GetString();
+            Element element;
+
+            // Map the type to the corresponding subclass
+            switch (typeValue)
+            {
+                case "software_File":
+                    element = JsonSerializer.Deserialize<File>(jsonObject.GetRawText(), options);
+                    break;
+
+                case "software_Package":
+                    element = JsonSerializer.Deserialize<Package>(jsonObject.GetRawText(), options);
+                    break;
+
+                case "ExternalMap":
+                    element = JsonSerializer.Deserialize<ExternalMap>(jsonObject.GetRawText(), options);
+                    break;
+
+                case "Relationship":
+                    element = JsonSerializer.Deserialize<Relationship>(jsonObject.GetRawText(), options);
+                    break;
+
+                default:
+                    throw new JsonException($"Unknown element type: {typeValue}");
+            }
+
+            elements.Add(element);
+
+            // Move to the next element in the array
+            reader.Read();
+        }
+
+        return elements;
+    }
 
     public override void Write(Utf8JsonWriter writer, List<Element> elements, JsonSerializerOptions options)
     {
--- a/src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Utils/SPDXToSbomFormatConverterExtensions.cs
+++ b/src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Utils/SPDXToSbomFormatConverterExtensions.cs
@@ -18,6 +18,19 @@ namespace Microsoft.Sbom.Utils;
 /// </summary>
 public static class SPDXToSbomFormatConverterExtensions
 {
+    public static SbomFile ToSbomFile(this File spdxFile)
+    {
+        var sbomFile = new SbomFile
+        {
+            Checksum = spdxFile.VerifiedUsing.ToSbomChecksum(),
+            FileCopyrightText = spdxFile.CopyrightText == "NOASSERTION" ? null : spdxFile.CopyrightText,
+            Path = spdxFile.Name,
+            Id = spdxFile.SpdxId
+        };
+
+        return sbomFile;
+    }
+
     /// <summary>
     /// Converts a <see cref="SPDXFile"/> object to a <see cref="SbomFile"/> object.
     /// </summary>
@@ -25,15 +38,15 @@ public static class SPDXToSbomFormatConverterExtensions
     /// <returns></returns>
     public static SbomFile ToSbomFile(this File spdxFile, List<Element> spdx30Elements, List<Common.Spdx30Entities.Relationship> relationships)
     {
-        // Manually setting the last two properties to default values.
         // Note that the SPDX 3.0 HAS_DECLARED_LICENSE relationship type is equivalent to LicenseInfoInFiles internally.
         var sbomFile = new SbomFile
         {
             Checksum = spdxFile.VerifiedUsing.ToSbomChecksum(),
             FileCopyrightText = spdxFile.CopyrightText == "NOASSERTION" ? null : spdxFile.CopyrightText,
-            LicenseConcluded = spdxFile.GetLicense(RelationshipType.HAS_CONCLUDED_LICENSE, spdx30Elements, relationships).FirstOrDefault(),
-            LicenseInfoInFiles = spdxFile.GetLicense(RelationshipType.HAS_DECLARED_LICENSE, spdx30Elements, relationships),
+            LicenseConcluded = spdxFile.GetSingleLicense(RelationshipType.HAS_CONCLUDED_LICENSE, spdx30Elements, relationships),
+            LicenseInfoInFiles = spdxFile.GetMultipleLicenses(RelationshipType.HAS_DECLARED_LICENSE, spdx30Elements, relationships),
             Path = spdxFile.Name,
+            Id = spdxFile.SpdxId
         };
 
         return sbomFile;
@@ -50,12 +63,13 @@ public static class SPDXToSbomFormatConverterExtensions
             Checksum = spdxPackage.VerifiedUsing.ToSbomChecksum(),
             LicenseInfo = new LicenseInfo
             {
-                Concluded = spdxPackage.GetLicense(RelationshipType.HAS_CONCLUDED_LICENSE, spdx30Elements, relationships).FirstOrDefault(),
-                Declared = spdxPackage.GetLicense(RelationshipType.HAS_DECLARED_LICENSE, spdx30Elements, relationships).FirstOrDefault(),
+                Concluded = spdxPackage.GetSingleLicense(RelationshipType.HAS_CONCLUDED_LICENSE, spdx30Elements, relationships),
+                Declared = spdxPackage.GetSingleLicense(RelationshipType.HAS_DECLARED_LICENSE, spdx30Elements, relationships),
             },
             Supplier = spdxPackage.GetSupplier(spdx30Elements),
             PackageUrl = spdxPackage.GetPackageUrl(spdx30Elements),
-            FilesAnalyzed = false,
+            FilesAnalyzed = true,
+            Id = spdxPackage.SpdxId,
         };
 
         return sbomPackage;
@@ -117,7 +131,7 @@ public static class SPDXToSbomFormatConverterExtensions
         return internalChecksums;
     }
 
-    internal static List<string> GetLicense(this Element element, RelationshipType relationshipType, List<Element> spdx30Elements, List<Common.Spdx30Entities.Relationship> relationships)
+    internal static List<string> GetMultipleLicenses(this Element element, RelationshipType relationshipType, List<Element> spdx30Elements, List<Common.Spdx30Entities.Relationship> relationships)
     {
         var spdxId = element.SpdxId;
         var relationshipsDescribingElement = relationships.Where(relationship => relationship.From == spdxId);
@@ -134,19 +148,56 @@ public static class SPDXToSbomFormatConverterExtensions
         var licenseElements = new List<string>();
         foreach (var toElementSpdxId in toElements)
         {
-            var licenseElement = spdx30Elements.FirstOrDefault(element => element.SpdxId == toElementSpdxId);
-            if (!licenseElement.Name.Equals("NoAssertion"))
+            var licenseElementsWithMatchingSpdxId = spdx30Elements.Where(element => element.SpdxId == toElementSpdxId);
+            if (licenseElementsWithMatchingSpdxId.Count() != 1)
             {
-                licenseElements.Add(licenseElement.Name);
+                return null;
+            }
+            else
+            {
+                licenseElements.Add(licenseElementsWithMatchingSpdxId.First().Name);
             }
         }
 
+        // If there are no license elements, return null.
         return licenseElements.Count == 0 ? null : licenseElements;
+    }
+
+    internal static string GetSingleLicense(this Element element, RelationshipType relationshipType, List<Element> spdx30Elements, List<Common.Spdx30Entities.Relationship> relationships)
+    {
+        var spdxId = element.SpdxId;
+        var relationshipsDescribingElement = relationships.Where(relationship => relationship.From == spdxId);
+
+        // Get all relationships that describe license information for the given package.
+        var toElements = new List<string>();
+        foreach (var relationship in relationshipsDescribingElement)
+        {
+            if (relationship.RelationshipType.Equals(relationshipType))
+            {
+                toElements = relationship.To;
+            }
+        }
+
+        // There should only be 1 relationship element for a given RelationshipType for the given Package.
+        if (toElements.Count != 1)
+        {
+            return null;
+        }
+
+        var licenseElements = spdx30Elements.Where(element => element.SpdxId == toElements.First());
+
+        // There should only be 1 matching license element.
+        return licenseElements.Count() == 1 ? licenseElements.First().Name : null;
     }
 
     internal static string GetSupplier(this Package spdxPackage, List<Element> spdx30Elements)
     {
         var organizationSpdxId = spdxPackage.SuppliedBy;
+        if (organizationSpdxId is null)
+        {
+            return null;
+        }
+
         var organizationElement = spdx30Elements
             .FirstOrDefault(element => element is Organization && element.SpdxId == organizationSpdxId);
         return organizationElement.Name.Equals("NOASSERTION") ? null : organizationElement.Name;
--- a/src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Utils/SPDXToSbomFormatConverterExtensions.cs
+++ b/src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Utils/SPDXToSbomFormatConverterExtensions.cs
@@ -1,10 +1,13 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System.Collections.Generic;
 using System.Linq;
 using Microsoft.Sbom.Common.Spdx30Entities;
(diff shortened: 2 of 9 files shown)


src/Microsoft.Sbom.Api/Utils/Comparer/SbomRelationshipComparer.cs
```diff
@@ -0,0 +1,51 @@
...
+namespace Microsoft.Sbom.Api.Utils.Comparer;
+
+/// <summary>
+/// Compares SBOM relationships based on their properties.
+/// </summary>
+public class SbomRelationshipComparer : IEqualityComparer<SbomRelationship>
+{
+    public bool Equals(SbomRelationship relationship1, SbomRelationship relationship2)
+    {
+        if (relationship1 == null || relationship2 == null)
```
DaveTryon, line 17: I think we're trying to use `is null` and `is not null` instead of `== null` and `!= null` in new code

src/Microsoft.Sbom.Api/Utils/Constants.cs
```diff
@@ -71,6 +71,7 @@ public static class Constants
     public const string CatalogFileName = "manifest.cat";
     public const string BsiFileName = "bsi.json";
     public const string DeleteManifestDirBoolVariableName = "DeleteManifestDirIfPresent";
+    public const string RootPackageIdValue = "SPDXRef-RootPackage";
```
DaveTryon, line 74: This constant is defined in both SPDX2 and SPDX3. That suggests that we're handling things inconsistently. I'm not entirely sure of the best fix. Making the parsers depend on another package doesn't seem great, but neither does having 3 copies of the same data.

src/Microsoft.Sbom.Common/Constants.cs
```diff
@@ -26,4 +26,5 @@ public static class Constants
     public const string SPDXRefFile = "SPDXRef-File";
     public const string SPDXRefPackage = "SPDXRef-Package";
     public const string SPDXRefExternalDocument = "DocumentRef";
+    public const string NoAssertionValue = "NOASSERTION";
```
DaveTryon, line 29: This constant is defined in both SPDX2 and SPDX3. That suggests that we're handling things inconsistently. I'm not entirely sure of the best fix. Making the parsers depend on another package doesn't seem great, but neither does having 3 copies of the same data.

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Generator.cs
```diff
@@ -524,18 +519,22 @@ private List<Element> GetSpdxRelationshipsFromSbomFile(Element spdxFileElement,
...
 
-        // If they exist, convert licenseDeclared to SPDX license elements and add Relationship elements for them
-        if (fileInfo.LicenseInfoInFiles == null || !fileInfo.LicenseInfoInFiles.Any())
-        {
-            return spdxRelationshipAndLicenseElementsToAddToSBOM;
-        }
-
         var toRelationships = new List<string>();
-        foreach (var licenseInfoInOneFile in fileInfo.LicenseInfoInFiles)
+        if (fileInfo.LicenseInfoInFiles == null)
```
DaveTryon, line 523:
```suggestion
        if (fileInfo.LicenseInfoInFiles is null)
```

src/Microsoft.Sbom.Api/Utils/Constants.cs
pragnya17: Moved to Api constants

src/Microsoft.Sbom.Api/Utils/Comparer/SbomFileComparer.cs
```diff
@@ -0,0 +1,45 @@
...
+
+    public bool Equals(SbomFile file1, SbomFile file2)
+    {
+        if (file1 is null || file2 is null)
+        {
+            return false;
+        }
+
+        var licenseInfosEqual = (file1.LicenseInfoInFiles is null && file2.LicenseInfoInFiles is null) ||
+                        file1.LicenseInfoInFiles?.SequenceEqual(file2.LicenseInfoInFiles ?? Enumerable.Empty<string>()) == true;
```
DaveTryon, line 23: Just to be sure, if `file1.LicenseInfoInFiles` is `null`, this ends up returning _false_ because `null != true` ?

src/Microsoft.Sbom.Api/Utils/Comparer/SbomPackageComparer.cs
```diff
@@ -0,0 +1,75 @@
...
+        if (package1 is null || package2 is null)
+        {
+            return false;
+        }
+
+        // Normalize PackageUrl for root packages since the SBOM tool version will vary between SPDX 2.2 and SPDX 3.0.
+        // Also normalize the Checksum since SPDX 2.2 does not support PackageVerificationCode to Checksum conversions for Root packages.
+        if (package1.Id == Constants.RootPackageIdValue && package2.Id == Constants.RootPackageIdValue)
+        {
+            package1.PackageUrl = NormalizePackagePurl(package1);
```
DaveTryon: I wouldn't expect an `Equals` method to modify the underlying data. Can we save the localized version in a local variable, instead?

src/Microsoft.Sbom.Api/Utils/Comparer/SbomFileComparer.cs
DaveTryon, line 23: `LicenseInfoInFiles` is an `IEnumerable<string>` without (as far as I can see) a guaranteed order. If `file1.LicenseInfo` is `["a", "b"]` and `file2.LicenseInfo` is `["b", "a"]`, should they be considered as equal? What about case sensitivity? Is "MIT" the same as "mit" ?

src/Microsoft.Sbom.Api/Utils/Comparer/SbomRelationshipComparer.cs
```diff
@@ -0,0 +1,51 @@
...
+{
+    public bool Equals(SbomRelationship relationship1, SbomRelationship relationship2)
+    {
+        if (relationship1 is null || relationship2 is null)
+        {
+            return false;
+        }
+
+        var equals = relationship1.RelationshipType.ToString().Equals(relationship2.RelationshipType.ToString(), StringComparison.OrdinalIgnoreCase) &&
+                relationship1.SourceElementId == relationship2.SourceElementId &&
```
DaveTryon: Indent here seems unusual

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Utils/ElementSerializer.cs
```diff
@@ -9,8 +9,72 @@
 
 public class ElementSerializer : JsonConverter<List<Element>>
```
DaveTryon, line 10: Side note: It feels weird to deserialize in a class that is called a Serializer. In a separate PR, I'd love to see this renamed to something like `ElementJsonConverter`

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Utils/ElementSerializer.cs
```diff
@@ -9,8 +9,72 @@
...
 {
-    public override List<Element> Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options) =>
-        throw new NotImplementedException("Deserialization of Elements into specific subtypes is not implemented yet.");
+    public override List<Element> Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
+    {
+        if (reader.TokenType != JsonTokenType.StartArray)
```
DaveTryon, line 14: Don't we already have code to read SPDX 3.0 files?

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Utils/SPDXToSbomFormatConverterExtensions.cs
```diff
@@ -185,17 +185,32 @@ private static string GetSingleLicense(this Element element, RelationshipType re
...
         return organizationElement?.Name;
     }
 
+    private static string GetOrganizationSpdxId(this Package spdxPackage, List<Element> spdx30Elements)
+    {
+        // Handle special case for the root package.
+        if (spdxPackage.SpdxId == Parsers.Spdx30SbomParser.Constants.RootPackageIdValue)
+        {
+            var creationInfo = spdx30Elements
+            .First(element => element is CreationInfo) as CreationInfo;
```
DaveTryon: Should this line be indented?
```suggestion
                .First(element => element is CreationInfo) as CreationInfo;
```

test/Microsoft.Sbom.Parsers.Spdx30SbomParser.Tests/Parser/JsonStrings/SbomFileJsonStrings.cs
```diff
@@ -103,7 +103,7 @@ public static class SbomFileJsonStrings
             ""to"": [
               ""SPDXRef-AnyLicenseInfo-3BA3FA6D3D66FE2BA75992BB0850D080F1223256368A76C77BEF8E0F6AC71896""
             ],
-            ""creationInfo"": ""_:creationinfo"",
+            ""creationInfo"": ""_:creationinfo"",
```
DaveTryon: Trim trailing spaces?
```suggestion
            ""creationInfo"": ""_:creationinfo"",
```

test/Microsoft.Sbom.Tool.Tests/IntegrationTests.cs
```diff
@@ -138,6 +139,40 @@ public void E2E_GenerateAndValidateSPDX30Manifest_ValidationSucceeds_ReturnsZero
...
+        GenerateManifestAndValidateSuccess(spdx30TestFolderPath, manifestInfoSpdxVersion: "3.0");
+        var spdx30ManifestPath = Path.Combine(AppendFullManifestFolderPath(spdx30TestFolderPath, "3.0"), ManifestFileName);
+        var spdx30Json = ReadJsonFile(spdx30ManifestPath);
+
+        // Use SbomEqualityComparer to compare the two manifests
+        var comparer = new SbomEqualityComparer(spdx22Json, spdx30Json);
+        var areEqual = comparer.DocumentsEqual();
+
+        // Assert that the manifests are equal
+        Assert.IsTrue(areEqual, "The SPDX 2.2 and SPDX 3.0 manifests should be equivalent.");
```
DaveTryon: Would it be interesting/possible to have e2e tests where we tweaked one aspect at a time to ensure that we pin how they differ?

src/Microsoft.Sbom.Api/Utils/Comparer/SbomEqualityComparer.cs
```diff
@@ -0,0 +1,291 @@
...
+    {
+        this.spdx22Json = spdx22Json;
+        this.spdx30Json = spdx30Json;
+        this.sbomFileComparer = new SbomFileComparer();
+        this.sbomPackageComparer = new SbomPackageComparer();
+        this.sbomRelationshipComparer = new SbomRelationshipComparer();
+        this.sbomReferenceComparer = new SbomReferenceComparer();
+    }
+
+    public bool DocumentsEqual()
```
DaveTryon: Is there a cheap way that we can we capture how the documents are different? Maybe a parameter we could use to capture the differences or a parameter that we populate while performing the comparison?

src/Microsoft.Sbom.Api/Utils/Comparer/SbomEqualityComparer.cs
```diff
@@ -0,0 +1,291 @@
...
+
+        var spdx30Files = elements.OfType<File>().ToList();
+        var spdx30Packages = elements.OfType<Package>().ToList();
+        var spdx30ExternalDocumentRefs = elements.OfType<ExternalMap>().ToList();
+        var spdx30Relationships = elements.OfType<Relationship>().ToList();
+
+        var externalDocRefsEqual = CheckExternalDocRefs(spdx22ExternalDocumentRefs, spdx30ExternalDocumentRefs);
+        if (!externalDocRefsEqual)
+        {
+            return externalDocRefsEqual;
```
DaveTryon, line 63: For cases where there can only be 1 possible value (applies to all of these return statements), it's often cleaner to just return that one possible value
```suggestion
            return false;
```

src/Microsoft.Sbom.Api/Utils/Comparer/SbomEqualityComparer.cs
```diff
@@ -0,0 +1,291 @@
...
+        return spdx22InternalExternalDocRefs.SetEquals(spdx30InternalExternalDocRefs);
+    }
+
+    private HashSet<SbomFile> ConvertToSbomFiles(List<SPDXFile> files)
+    {
+        var sbomFiles = new HashSet<SbomFile>(sbomFileComparer);
+
+        foreach (var file in files)
+        {
+            sbomFiles.Add(file.ToSbomFile());
```
DaveTryon, line 152: What is the expectation on the case-sensitivity of these `HashSet` objects? I know that we have some code that uses `IOSUtils.IsCaseSensitiveOS` to handle casing differently on Windows vs Linux or macos

src/Microsoft.Sbom.Api/Utils/Comparer/SbomFileComparer.cs
pragnya17: The intent is that if both values are null this will return true, and if not we compare the actual values for equality. Does that help clarify things?

pragnya17: LicenseInfoInFiles is always populated as NOASSERTION in SPDX 2.2 files, therefore I hadn't considered the order and case sensitivity here. Given the data, I don't think it's necessary to consider those cases, but for code completeness I'm happy to make the change. Feel free to let me know which direction you'd prefer.

test/Microsoft.Sbom.Tool.Tests/IntegrationTests.cs
pragnya17: There's lots of different combinations of how the docs can differ, are there any specific cases you're thinking of that we should target?

src/Microsoft.Sbom.Api/Utils/Comparer/SbomEqualityComparer.cs
pragnya17: How about logging by default whenever there is a difference? I was thinking it would be sufficient to log which part of the SBOM is different. For example, if the files differ we could log that the files don't match.

I'm hesitant to log the details of which files could be different because that could result in a lot of log output since the SBOMs can get pretty large.

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Utils/ElementSerializer.cs
pragnya17: Yes the SPDX30Parser already does this, but in a slightly different way since we mirrored the SPDX22Parser. The Parser parses each json object in the array and performs deduplication while parsing. 

The SBOM equality comparer simply needs to read all the elements, therefore I chose to implement the Read() on the ElementSerializer in order to be able to do the following:

`var elements = graphArray.Deserialize<List<Element>>(this.serializerOptions);`

pragnya17: We could call the Parser code here, but that would introduce an extra dependency. Ideally, we would refactor all the SPDX parser code for both 2.2 and 3.0 to just use the Read methods from a custom serializer like ElementSerializer but that's a larger change.

src/Microsoft.Sbom.Api/Utils/Comparer/SbomFileComparer.cs
DaveTryon: It's probably OK if this is intended for internal usage. We'd want to fix this if we made this a generally available feature, through

src/Microsoft.Sbom.Api/Utils/Comparer/SbomEqualityComparer.cs
DaveTryon: Did you mean to change this? It doesn't seem to be updated

src/Microsoft.Sbom.Api/Utils/Comparer/SbomEqualityComparer.cs
```diff
@@ -0,0 +1,291 @@
...
+        var externalDocRefsEqual = CheckExternalDocRefs(spdx22ExternalDocumentRefs, spdx30ExternalDocumentRefs);
+        if (!externalDocRefsEqual)
+        {
+            return externalDocRefsEqual;
+        }
+
+        var relationshipsEqual = CheckRelationships(spdx22Relationships, spdx30Relationships);
+        if (!relationshipsEqual)
+        {
+            return relationshipsEqual;
```
DaveTryon, line 69:
```suggestion
            return false;
```

src/Microsoft.Sbom.Api/Utils/Comparer/SbomEqualityComparer.cs
```diff
@@ -0,0 +1,291 @@
...
+        }
+
+        var filesEqual = CheckFiles(spdx22Files, spdx30Files, elements, spdx30Relationships);
+        if (!filesEqual)
+        {
+            return filesEqual;
```
DaveTryon, line 75:
```suggestion
            return false;
```

src/Microsoft.Sbom.Api/Utils/Comparer/SbomEqualityComparer.cs
```diff
@@ -0,0 +1,291 @@
...
+        }
+
+        var packagesEqual = CheckPackages(spdx22Packages, spdx30Packages, elements, spdx30Relationships);
+        if (!packagesEqual)
+        {
+            return packagesEqual;
```
DaveTryon, line 81:
```suggestion
            return false;
```

src/Microsoft.Sbom.Api/Utils/Comparer/SbomEqualityComparer.cs
DaveTryon: A logging statement each time we have an early exit should be enough for our needs. Again, if we were making this production code we'd probably want some option that's easier to adopt programmatically

src/Microsoft.Sbom.Api/Utils/Comparer/SbomFileComparer.cs
DaveTryon: I understand the "both null is true" case, but the "at least one is not null" case is comparing the raw `file1.LicenseInfoInFiles` against an adjusted `file2.LicenseInfoInFiles`. If the assertion is that a null value and an empty list are equivalent, then you can reduce all cases to the following:

```
var file1LicenseInfoInFiles = file1.LicenseInfoInFiles ?? Enumerable.Empty<string>();
var file2LicenseInfoInFiles = file2.LicenseInfoInFiles ?? Enumerable.Empty<string>();
var licenseInfosEqual = file1LicenseInfoInFiles.SequenceEqual(file2LicenseInfoInFiles);
```

src/Microsoft.Sbom.Api/Utils/Comparer/SbomPackageComparer.cs
DaveTryon: This is still changing the inputs

test/Microsoft.Sbom.Tool.Tests/IntegrationTests.cs
DaveTryon: I'd suggest 1 case for each scenario that leads to an early failure. Ideally each test could detect that it failed for the expected reason. If we go with reporting results via logging, then your mock could monitor the logging calls and set a flag when a specific call occurs. Happy to chat on this, as needed

src/Microsoft.Sbom.Api/Utils/Comparer/SbomPackageComparer.cs
pragnya17: This is outdated, does the new implementation match what you expected?

src/Microsoft.Sbom.Api/Utils/Comparer/SbomFileComparer.cs
pragnya17: Technically speaking, a null value and an empty list shouldn't evaluate to the same thing. A null value indicates "NOASSERTION" while an empty string means there are no licenses to be found. SPDX defines both those things separately.

test/Microsoft.Sbom.Api.Tests/Utils/SbomEqualityComparerTests.cs
```diff
@@ -196,15 +196,15 @@ public void SetUp()
     public void CheckFiles_MatchingFiles_ReturnsTrue()
     {
         var result = comparer.CheckFiles(spdx22Files, spdx30Files, spdx30Elements, relationships);
-        Assert.IsTrue(result, "File contents should be equal");
+        Assert.AreEqual(SbomEqualityComparisonResult.Equal, result, "File contents should be equal");
```
DaveTryon: Now that you're using `Assert.AreEqual` and `Assert.AreNotEqual` instead of `Assert.IsTrue` and `Assert.IsFalse`, the comments in the asserts are a lot less valuable but might even be considered a maintenance liability

DaveTryon (approved):
Approved with a couple of optional suggestions

DaveTryon (approved):

PR #1012, opened 2025-04-10
Add SPDX 3.0 extensions to convert to internal SBOM components

pragnya17:
This PR adds extensions for SPDX 3.0 elements so that they can be converted into internal SBOM components. This conversion enables use to get the content of an SPDX 3.0 document in a uniform format that can be used internally.

### DIFF
--- a/src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Utils/SPDXToSbomFormatConverterExtensions.cs
+++ b/src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Utils/SPDXToSbomFormatConverterExtensions.cs
@@ -205,7 +205,7 @@ public static class SPDXToSbomFormatConverterExtensions
 
     internal static string GetPackageUrl(this Package spdxPackage, List<Element> spdx30Elements)
     {
-        var externalIdentifierSpdxId = spdxPackage.ExternalIdentifier.FirstOrDefault();
+        var externalIdentifierSpdxId = spdxPackage.ExternalIdentifier?.FirstOrDefault();
         if (externalIdentifierSpdxId is null)
         {
             return null;
--- a/src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Utils/SPDXToSbomFormatConverterExtensions.cs
+++ b/src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Utils/SPDXToSbomFormatConverterExtensions.cs
@@ -23,15 +23,11 @@ public static class SPDXToSbomFormatConverterExtensions
         return new SbomFile
         {
             Checksum = spdxFile.VerifiedUsing?.ToSbomChecksum(),
-            FileCopyrightText = spdxFile.CopyrightText == "NOASSERTION" ? null : spdxFile.CopyrightText,
+            FileCopyrightText = spdxFile.CopyrightText,
             Path = spdxFile.Name,
             Id = spdxFile.SpdxId,
-            LicenseConcluded = spdx30Elements != null && relationships != null
-            ? spdxFile.GetSingleLicense(RelationshipType.HAS_CONCLUDED_LICENSE, spdx30Elements, relationships)
-            : null,
-            LicenseInfoInFiles = spdx30Elements != null && relationships != null
-            ? spdxFile.GetMultipleLicenses(RelationshipType.HAS_DECLARED_LICENSE, spdx30Elements, relationships)
-            : null
+            LicenseConcluded = spdxFile.GetSingleLicense(RelationshipType.HAS_CONCLUDED_LICENSE, spdx30Elements, relationships),
+            LicenseInfoInFiles = spdxFile.GetMultipleLicenses(RelationshipType.HAS_DECLARED_LICENSE, spdx30Elements, relationships)
         };
     }
 
@@ -41,8 +37,8 @@ public static class SPDXToSbomFormatConverterExtensions
         {
             PackageName = spdxPackage.Name,
             PackageVersion = spdxPackage.PackageVersion,
-            PackageSource = spdxPackage.DownloadLocation == "NOASSERTION" ? null : spdxPackage.DownloadLocation,
-            CopyrightText = spdxPackage.CopyrightText == "NOASSERTION" ? null : spdxPackage.CopyrightText,
+            PackageSource = spdxPackage.DownloadLocation,
+            CopyrightText = spdxPackage.CopyrightText,
             Checksum = spdxPackage.VerifiedUsing?.ToSbomChecksum(),
             LicenseInfo = new LicenseInfo
             {
@@ -118,6 +114,12 @@ public static class SPDXToSbomFormatConverterExtensions
 
     private static List<string> GetMultipleLicenses(this Element element, RelationshipType relationshipType, List<Element> spdx30Elements, List<Relationship> relationships)
     {
+        // Return null if spdx30Elements or relationships are null
+        if (spdx30Elements == null || relationships == null)
+        {
+            return null;
+        }
+
         var spdxId = element.SpdxId;
         var relationshipsDescribingElement = relationships.Where(relationship => relationship.From == spdxId);
 
@@ -150,6 +152,12 @@ public static class SPDXToSbomFormatConverterExtensions
 
     private static string GetSingleLicense(this Element element, RelationshipType relationshipType, List<Element> spdx30Elements, List<Relationship> relationships)
     {
+        // Return null if spdx30Elements or relationships are null
+        if (spdx30Elements == null || relationships == null)
+        {
+            return null;
+        }
+
         var spdxId = element.SpdxId;
         var relationshipsDescribingElement = relationships.Where(relationship => relationship.From == spdxId);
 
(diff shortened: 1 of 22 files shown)


src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Utils/SPDXToSbomFormatConverterExtensions.cs
```diff
@@ -15,36 +18,202 @@ namespace Microsoft.Sbom.Utils;
...
+        return sbomFile;
+    }
+
     /// <summary>
     /// Converts a <see cref="SPDXFile"/> object to a <see cref="SbomFile"/> object.
     /// </summary>
     /// <param name="spdxFile"></param>
     /// <returns></returns>
-    public static SbomFile ToSbomFile(this File spdxFile)
+    public static SbomFile ToSbomFile(this File spdxFile, List<Element> spdx30Elements, List<Common.Spdx30Entities.Relationship> relationships)
```
sfoslund, line 39: It looks like there's a lot of overlap between the 2 `ToSbomFile` methods, can we combine them by making the additional args optional?

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Utils/SPDXToSbomFormatConverterExtensions.cs
```diff
@@ -15,36 +18,202 @@ namespace Microsoft.Sbom.Utils;
...
+            Supplier = spdxPackage.GetSupplier(spdx30Elements),
+            PackageUrl = spdxPackage.GetPackageUrl(spdx30Elements),
+            FilesAnalyzed = true,
+            Id = spdxPackage.SpdxId,
+        };
+
+        return sbomPackage;
+    }
+
+    public static List<SbomRelationship> ToSbomRelationship(this Common.Spdx30Entities.Relationship spdxRelationship)
```
sfoslund, line 78: What is the intention with all of these new methods? Are they eventually going to be called somewhere?

pragnya17: Yup they will be called here - https://github.com/microsoft/sbom-tool/pull/1011

pragnya17: I just wanted to separate the changes since it is a larger change

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Utils/SPDXToSbomFormatConverterExtensions.cs
```diff
@@ -15,36 +18,186 @@ namespace Microsoft.Sbom.Utils;
...
+    public static SbomFile ToSbomFile(this File spdxFile, List<Element> spdx30Elements = null, List<Relationship> relationships = null)
     {
-        var checksums = spdxFile.VerifiedUsing?.Select(c => c.ToSbomChecksum());
-
-        // Not setting LicenseConcluded and LicenseInfoInFiles since the whole SBOM is required to set these values.
         return new SbomFile
         {
-            Checksum = checksums,
-            FileCopyrightText = spdxFile.CopyrightText,
+            Checksum = spdxFile.VerifiedUsing.ToSbomChecksum(),
```
sfoslund, line 25: If I'm reading this right, it looks like we used to allow spdxFile.VerifiedUsing to be null (see the ? in the old code) but now it's required to be non-null. Is that an intentional change? Also why is the .Select no longer necessary?

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Utils/SPDXToSbomFormatConverterExtensions.cs
```diff
@@ -15,36 +18,186 @@ namespace Microsoft.Sbom.Utils;
...
+            FileCopyrightText = spdxFile.CopyrightText == "NOASSERTION" ? null : spdxFile.CopyrightText,
```
sfoslund, line 26: Why do we want null instead of no assertion?

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Utils/SPDXToSbomFormatConverterExtensions.cs
```diff
@@ -15,36 +18,186 @@ namespace Microsoft.Sbom.Utils;
...
+    public static SbomReference ToSbomReference(this ExternalMap externalDocumentReference)
+    {
+        if (externalDocumentReference is null)
+        {
+            return null;
+        }
+
+        return new SbomReference
+        {
+            Checksum = externalDocumentReference.VerifiedUsing?.ToSbomChecksum().FirstOrDefault(),
```
sfoslund, line 87: Do we ever expect there to be more than one here? Why FirstOrDefault?

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Utils/SPDXToSbomFormatConverterExtensions.cs
pragnya17: I removed the ? in favor of adding a null check in the helper method instead. Select is no longer necessary because I am returning a List<SbomChecksum> rather than a singular SbomChecksum.

pragnya17: Correction on the first part: It looks like in SPDX 2.2 extensions we use ? rather than adding a null check in the helper method, so I will add back the ? to maintain consistency. Also in the helper method, I returned null if VerifiedUsing.Count == 0, but this is not consistent with SPDX 2.2 extension. Rather than returning null it should return an empty checksum list.

I will also add a unit test to validate this behavior for checksums.

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Utils/SPDXToSbomFormatConverterExtensions.cs
pragnya17: NOASSERTION is the public facing value we use in SPDX documents. Internally, this is equivalent to a null value. 

In our SBOM generation code, when we go from SBOMFile to SPDXFile we convert null values to NOASSERTION. Similarly, in these extension methods I reversed the logic so that the NOASSERTION values are converted to null values.

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Utils/SPDXToSbomFormatConverterExtensions.cs
pragnya17: There will never be more than one. In the public facing SPDX documentation, VerifiedUsing is always represented as a list. However, internally for SbomReference's there's only one allowed checksum value.

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Utils/SPDXToSbomFormatConverterExtensions.cs
```diff
@@ -15,36 +18,186 @@ namespace Microsoft.Sbom.Utils;
...
         {
-            Checksum = checksums,
-            FileCopyrightText = spdxFile.CopyrightText,
+            Checksum = spdxFile.VerifiedUsing?.ToSbomChecksum(),
+            FileCopyrightText = spdxFile.CopyrightText == "NOASSERTION" ? null : spdxFile.CopyrightText,
+            Path = spdxFile.Name,
             Id = spdxFile.SpdxId,
-            Path = spdxFile.Name
+            LicenseConcluded = spdx30Elements != null && relationships != null
+            ? spdxFile.GetSingleLicense(RelationshipType.HAS_CONCLUDED_LICENSE, spdx30Elements, relationships)
```
sfoslund, line 30: nit: feels like we should be checking if spdx30Elements != null and relationships != null in spdxFile.GetSingleLicense and just returning null in that case. That way, this logic could just be simplified to a call to spdxFile.GetSingleLicense. (Same below for spdxFile.GetMultipleLicenses)

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Utils/SPDXToSbomFormatConverterExtensions.cs
```diff
@@ -15,36 +18,186 @@ namespace Microsoft.Sbom.Utils;
...
+        };
+    }
+
+    public static SbomPackage ToSbomPackage(this Package spdxPackage, List<Element> spdx30Elements, List<Relationship> relationships)
+    {
+        var sbomPackage = new SbomPackage
+        {
+            PackageName = spdxPackage.Name,
+            PackageVersion = spdxPackage.PackageVersion,
+            PackageSource = spdxPackage.DownloadLocation == "NOASSERTION" ? null : spdxPackage.DownloadLocation,
```
sfoslund, line 44: nit: we use the string "NOASSERTION" a lot throughout this class, can we keep it as a const instead of duplicating the string every time?

sfoslund (commented):
Looks good, just a few more nits

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Utils/SPDXToSbomFormatConverterExtensions.cs
pragnya17: Slight correction: While the above is still true, it looks like SPDX 2.2 extensions does not follow this format and instead just uses whatever is provided in the spdx element. So for example if NOASSERTION is provided then the value remains as NOASSERTION.

Since the point of adding these extensions is to convert all SPDX data to an internal standard, I will conform with what is already in place for SPDX 2.2 and remove the NOASSERTION to null conversions.

PR #1013, opened 2025-04-10
Update SPDX 2.2 extension

pragnya17:
This pull request updates the SPDX 2.2 extension to include new functionality for converting an External Document Reference to an internal SBOM component that is standardized.


pragnya17:
These changes are redundant. Rather than using a different method we can reuse ToSbomReference().

PR #1015, opened 2025-04-18
Fix SPDX 3.0 relationship generation

pragnya17:
1. SPDX id's for relationships were not being generated correctly. The SPDX id's were used to filter out duplicate relationships, therefore relationships were being incorrectly identified as duplicates and were missing from the final manifest SBOM file. 
2. For certain relationship types we were not using the correct directionality as stated by the SPDX specification.

This PR aims to fix how the SPDX id for relationships are generated, so that the ID is truly a unique representation of the content of the relationship. It also fixes the directionality for certain relationship types.

### DIFF
--- a/src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Generator.cs
+++ b/src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Generator.cs
@@ -295,14 +295,13 @@ public class Generator : IManifestGenerator
         };
 
         spdxExternalMap.AddExternalSpdxId(externalDocumentReferenceInfo.ExternalDocumentName, externalDocumentReferenceInfo.Checksum);
-        var externalDocumentReferenceId = spdxExternalMap.ExternalSpdxId;
 
         return new GenerationResult
         {
             Document = JsonDocument.Parse(JsonSerializer.Serialize(spdxExternalMap, this.serializerOptions)),
             ResultMetadata = new ResultMetadata
             {
-                EntityId = externalDocumentReferenceId
+                EntityId = spdxExternalMap.SpdxId
             }
         };
     }
--- a/src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Generator.cs
+++ b/src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Generator.cs
@@ -125,8 +125,8 @@ public class Generator : IManifestGenerator
             CopyrightText = packageInfo.CopyrightText ?? Constants.NoAssertionValue,
             SuppliedBy = spdxSupplier.SpdxId,
         };
-        var packageId = SPDXExtensions.GetSpdxElementId(packageInfo);
-        spdxPackage.AddSpdxId(packageId);
+
+        spdxPackage.AddSpdxId(packageInfo);
 
         var spdxRelationshipAndLicensesFromSbomPackage = GetSpdxRelationshipsAndLicensesFromSbomPackage(packageInfo, spdxPackage);
 
@@ -160,7 +160,7 @@ public class Generator : IManifestGenerator
         var dependOnId = packageInfo.DependOn;
         if (dependOnId is not null && dependOnId != Constants.RootPackageIdValue)
         {
-            dependOnId = SPDXExtensions.GenerateSpdxId(spdxPackage, packageInfo.DependOn);
+            dependOnId = spdxPackage.AddSpdxId(packageInfo);
         }
 
         return new GenerationResult
@@ -230,7 +230,6 @@ public class Generator : IManifestGenerator
         };
 
         spdxSupplier.AddSpdxId();
-        spdxPackage.AddSpdxId();
         spdxRelationshipLicenseDeclaredElement.AddSpdxId();
         spdxRelationshipLicenseConcludedElement.AddSpdxId();
 
@@ -296,7 +295,6 @@ public class Generator : IManifestGenerator
         };
 
         spdxExternalMap.AddExternalSpdxId(externalDocumentReferenceInfo.ExternalDocumentName, externalDocumentReferenceInfo.Checksum);
-        spdxExternalMap.AddSpdxId();
         var externalDocumentReferenceId = spdxExternalMap.ExternalSpdxId;
 
         return new GenerationResult
@@ -496,8 +494,8 @@ public class Generator : IManifestGenerator
             Name = GeneratorUtils.EnsureRelativePathStartsWithDot(fileInfo.Path),
             CopyrightText = fileInfo.FileCopyrightText ?? Constants.NoAssertionValue,
         };
-        var fileId = SPDXExtensions.GetSpdxFileId(fileInfo.Path, fileInfo.Checksum);
-        spdxFileElement.AddSpdxId(fileId);
+
+        spdxFileElement.AddSpdxId(fileInfo);
 
         // Generate SPDX spdxRelationship elements
         var spdxRelationshipsFromSbomFile = GetSpdxRelationshipsFromSbomFile(spdxFileElement, fileInfo);
--- a/src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Generator.cs
+++ b/src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Generator.cs
@@ -602,7 +602,7 @@ public class Generator : IManifestGenerator
     {
         var spdxRelationshipType = this.GetSPDXRelationshipType(relationshipType);
 
-        // Switch source and target IDs for these specific relationship types
+        // Switch source and target IDs for these specific relationship types to invert directionality.
         if (relationshipType == SbomEntities.RelationshipType.PREREQUISITE_FOR ||
             relationshipType == SbomEntities.RelationshipType.DESCRIBED_BY ||
             relationshipType == SbomEntities.RelationshipType.PATCH_FOR)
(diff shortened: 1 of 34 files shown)


src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Utils/SPDXExtensions.cs
```diff
@@ -106,7 +106,13 @@ public static void AddSpdxId(this CreationInfo creationInfo)
 
     public static void AddSpdxId(this Common.Spdx30Entities.Relationship relationship)
     {
-        relationship.SpdxId = GenerateSpdxIdBasedOnElement(relationship, relationship.To + relationship.RelationshipType.ToString());
+        var relationsipToString = string.Empty;
+        if (relationship?.To != null && relationship.To.Any())
```
DaveTryon, line 110:
```suggestion
        if (relationship?.To is not null && relationship.To.Any())
```

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Utils/SPDXExtensions.cs
```diff
@@ -106,7 +106,13 @@ public static void AddSpdxId(this CreationInfo creationInfo)
 
     public static void AddSpdxId(this Common.Spdx30Entities.Relationship relationship)
     {
-        relationship.SpdxId = GenerateSpdxIdBasedOnElement(relationship, relationship.To + relationship.RelationshipType.ToString());
+        var relationsipToString = string.Empty;
```
DaveTryon, line 109: Typo? Will need to be fixed in multiple lines
```suggestion
        var relationshipToString = string.Empty;
```

test/Microsoft.Sbom.Parsers.Spdx30SbomParser.Tests/Parser/GeneratorTests.cs
```diff
@@ -181,6 +181,69 @@ public void GenerateJsonDocumentTest_Relationship()
...
+        };
+
+        var generatorResult = generator.GenerateJsonDocument(relationshipInfo);
+        var generatedJsonString = generatorResult.Document.RootElement.GetRawText();
+        generatedJsonString = NormalizeString(generatedJsonString);
+
+        var expectedJsonContentAsString = SbomRelationshipJsonStrings.RelationshipPrereqForJsonString;
+        expectedJsonContentAsString = NormalizeString(expectedJsonContentAsString);
+
+        Assert.IsFalse(generatedJsonString.Contains("null"));
```
DaveTryon, line 201: It feels like this line is unneeded because the `Assert.AreEqual` on the next line will catch the same case
```suggestion
```

test/Microsoft.Sbom.Parsers.Spdx30SbomParser.Tests/Parser/GeneratorTests.cs
```diff
@@ -181,6 +181,69 @@ public void GenerateJsonDocumentTest_Relationship()
...
+        };
+
+        var generatorResult = generator.GenerateJsonDocument(relationshipInfo);
+        var generatedJsonString = generatorResult.Document.RootElement.GetRawText();
+        generatedJsonString = NormalizeString(generatedJsonString);
+
+        var expectedJsonContentAsString = SbomRelationshipJsonStrings.RelationshipDescribedByJsonString;
+        expectedJsonContentAsString = NormalizeString(expectedJsonContentAsString);
+
+        Assert.IsFalse(generatedJsonString.Contains("null"));
```
DaveTryon, line 222: It feels like this line is unneeded because the `Assert.AreEqual` on the next line will catch the same case
```suggestion
```

test/Microsoft.Sbom.Parsers.Spdx30SbomParser.Tests/Parser/GeneratorTests.cs
```diff
@@ -181,6 +181,69 @@ public void GenerateJsonDocumentTest_Relationship()
...
+        };
+
+        var generatorResult = generator.GenerateJsonDocument(relationshipInfo);
+        var generatedJsonString = generatorResult.Document.RootElement.GetRawText();
+        generatedJsonString = NormalizeString(generatedJsonString);
+
+        var expectedJsonContentAsString = SbomRelationshipJsonStrings.RelationshipPatchForJsonString;
+        expectedJsonContentAsString = NormalizeString(expectedJsonContentAsString);
+
+        Assert.IsFalse(generatedJsonString.Contains("null"));
```
DaveTryon, line 243: It feels like this line is unneeded because the `Assert.AreEqual` on the next line will catch the same case
```suggestion
```

test/Microsoft.Sbom.Parsers.Spdx30SbomParser.Tests/Parser/GeneratorTests.cs
```diff
@@ -181,6 +181,69 @@ public void GenerateJsonDocumentTest_Relationship()
         Assert.AreEqual(expectedJsonContentAsString, generatedJsonString);
     }
 
+    [TestMethod]
+    public void GenerateJsonDocumentTest_PrereqFor_Relationship()
```
DaveTryon, line 185: Feels like we should test the direction both ways, since we have code to enforce it. Something like this:
[TestMethod]
[DataRow("source-id", "target-id")]
[DataRow("target-id", "source-id")]
public TestName(string sourceElementId, string targetElementId)
{
  var relatinshipInfo = new Relationship
  {
    SourceElementId = sourceElementId,
    TargetElementId = targetElementId,
...
```

File: test/Microsoft.Sbom.Parsers.Spdx30SbomParser.Tests/Parser/GeneratorTests.cs [hunk 7]
```diff
@@ -181,6 +181,69 @@ public void GenerateJsonDocumentTest_Relationship()
...
+        Assert.IsFalse(generatedJsonString.Contains("null"));
+        Assert.AreEqual(expectedJsonContentAsString, generatedJsonString);
+    }
+
+    [TestMethod]
+    public void GenerateJsonDocumentTest_DescribedBy_Relationship()
+    {
+        var relationshipInfo = new Relationship
+        {
+            SourceElementId = "source-id",
```
DaveTryon, line 210: Rather than use "source-id" and "target-id" as specific strings, it might make sense to make them `const` values in the class. That's doubly true if you use the `[DataRow]` mentioned above

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Generator.cs
```diff
@@ -601,6 +595,28 @@ private Element GenerateLicenseElement(string licenseInfo)
         return licenseElement;
     }
 
+    private SpdxEntities.Relationship GetSpdxRelationship(string sourceElement, string targetElement, SbomEntities.RelationshipType relationshipType)
+    {
+        var spdxRelationshipType = this.GetSPDXRelationshipType(relationshipType);
+
+        // Switch source and target IDs for these specific relationship types
+        if (relationshipType == SbomEntities.RelationshipType.PREREQUISITE_FOR ||
```
DaveTryon, line 603: It feels odd to do this for SPDX 3 but not SPDX 2. Did SPDX redefine the direction of the relationship between the two versions?

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Generator.cs
```diff
@@ -601,6 +595,28 @@ private Element GenerateLicenseElement(string licenseInfo)
...
+            relationshipType == SbomEntities.RelationshipType.DESCRIBED_BY ||
+            relationshipType == SbomEntities.RelationshipType.PATCH_FOR)
+        {
+            (sourceElement, targetElement) = (targetElement, sourceElement);
```
DaveTryon, line 607: Do/should we have unit tests for the `RelationshipType` values that don't change the relationship direction?

pragnya17: Yup there's multiple other relationship types that do not change direction. There's already 1 UT that's representative of all of them here - GenerateJsonDocumentTest_Relationship

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Generator.cs
pragnya17: In SPDX 2.2 there were some relationships that the provided inverses for as well. For example, DESCRIBES and DESCRIBED_BY were inverses of each other. 

In 3.0, these inverses are no longer present. So in the above example, only DESCRIBES is valid. Therefore if we get a DESCRIBED_BY relationship we must convert it to DESCRIBES by changing the directionality for it to be a valid SPDX 3.0 relationship type.

DaveTryon: Ah. Looking at `GetSPDXRelationshipType`, I now see where we map `PREREQUISITE_FOR`, `DESCRIBED_BY`, and `PATCH_FOR` to `HAS_PREREQUISITE`, `DESCRIBES`, and `PATCHED_BY`, respectively. I completely missed that subtlety until now. That feels like it deserves a comment--maybe something like this inside the switch block on line 631?

```
// These 3 relationships intentionally change the relationship direction to conform with the SPDX 3.0 spec
case SbomEntities.RelationshipType.PREREQUISITE_FOR: return RelationshipType.HAS_PREREQUISITE;
case SbomEntities.RelationshipType.DESCRIBED_BY: return RelationshipType.DESCRIBES;
case SbomEntities.RelationshipType.PATCH_FOR: return RelationshipType.PATCHED_BY;
```

test/Microsoft.Sbom.Parsers.Spdx30SbomParser.Tests/Parser/GeneratorTests.cs
```diff
@@ -32,6 +32,9 @@ public class GeneratorTests
     private readonly Mock<IFileSystemUtils> fileSystemMock = new Mock<IFileSystemUtils>(MockBehavior.Strict);
     private readonly Mock<IManifestConfigHandler> mockConfigHandler = new Mock<IManifestConfigHandler>(MockBehavior.Strict);
 
+    private const string SourceElementIdValue = "source-id";
+    private const string TargetElementIdValue = "source-id";
```
DaveTryon, line 36: Copy/paste error ?
```suggestion
    private const string TargetElementIdValue = "target-id";
```

DaveTryon (approved):

DaveTryon (approved):

PR #1016, opened 2025-04-18
NoAssertion bug for SBOM file and package generation

pragnya17:
For files, if there is no `LicenseInfoInFiles` we should create a `NOASSERTION` element to and a `HAS_DECLARED_LICENSE` relationship between the file and the `NOASSERTION` element. However, instead of generating a NOASSERTION element to represent the license we were just generating no element. This PR fixes this bug by generating the correct NOASSERTION element when there is no license info.

Additionally, we make the NOASSERTION value for the element capitalized, so that it is consistent with SPDX 2.2. This applies to both package and files.


src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Generator.cs
```diff
@@ -524,18 +524,22 @@ private List<Element> GetSpdxRelationshipsFromSbomFile(Element spdxFileElement,
...
 
-        // If they exist, convert licenseDeclared to SPDX license elements and add Relationship elements for them
-        if (fileInfo.LicenseInfoInFiles == null || !fileInfo.LicenseInfoInFiles.Any())
-        {
-            return spdxRelationshipAndLicenseElementsToAddToSBOM;
-        }
-
         var toRelationships = new List<string>();
-        foreach (var licenseInfoInOneFile in fileInfo.LicenseInfoInFiles)
+        if (fileInfo.LicenseInfoInFiles == null)
```
DaveTryon, line 528:
```suggestion
        if (fileInfo.LicenseInfoInFiles is null)
```

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Generator.cs
```diff
@@ -524,18 +524,22 @@ private List<Element> GetSpdxRelationshipsFromSbomFile(Element spdxFileElement,
...
         {
-            var licenseDeclaredElement = GenerateLicenseElement(licenseInfoInOneFile);
+            var licenseDeclaredElement = GenerateLicenseElement(null);
```
DaveTryon, line 530: This always generates a new `NoAssertionElement`. Would it be useful to create a single immutable `NoAssertionElement` that we just use over and over? The "immutable" part may make this more trouble that it's worth

DaveTryon: That looks OK, as long as we're confident that we never change its properties. The underlying `Element` object is not immutable, so it creates a potential for an unintended change to apply to multiple places

DaveTryon (approved):

DaveTryon (approved):

PR #1017, opened 2025-04-18
Package DependOnId bug fix

pragnya17:
The dependOnId field for packages was being generated incorrectly. This did not affect package generation, but did impact the relationship generation since the package's dependOnId is used to generate DEPENDS_On relationships.


test/Microsoft.Sbom.Parsers.Spdx30SbomParser.Tests/Parser/GeneratorTests.cs
```diff
@@ -85,6 +86,49 @@ public void GenerateJsonDocumentTest_Package()
...
+    {
+        var packageInfo = new SbomPackage
+        {
+            PackageName = "TestPackage",
+            DependOn = null
+        };
+
+        var result = generator.GenerateJsonDocument(packageInfo);
+
+        Assert.IsNull(result.ResultMetadata.DependOn, "DependOnId should be null when DependOn is null.");
```
DaveTryon, line 100: You probably don't need the description in lines 100, 114, and 129. These tests are single case and will already report the expected and actual values.

DaveTryon (approved):

PR #1019, opened 2025-04-18
Bug fix for supplier and suppliedBy for root package in SPDX 3.0

pragnya17:
The Organization and Tool name for the root package in SPDX 3.0 generation should have a prefix to match SPDX 2.2. This will ensure that the supplier info for organization and tool names are consistently formatted between SPDX 2.2 and 3.0.

### DIFF
--- a/src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Utils/SPDXToSbomFormatConverterExtensions.cs
+++ b/src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Utils/SPDXToSbomFormatConverterExtensions.cs
@@ -206,6 +206,6 @@ public static class SPDXToSbomFormatConverterExtensions
 
         var externalIdentifier = (ExternalIdentifier)spdx30Elements
             .FirstOrDefault(element => element is ExternalIdentifier && element.SpdxId == externalIdentifierSpdxId);
-        return externalIdentifier.Identifier;
+        return externalIdentifier?.Identifier;
     }
 }
--- a/src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Utils/SPDXToSbomFormatConverterExtensions.cs
+++ b/src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Utils/SPDXToSbomFormatConverterExtensions.cs
@@ -185,30 +185,15 @@ public static class SPDXToSbomFormatConverterExtensions
 
     private static string GetSupplier(this Package spdxPackage, List<Element> spdx30Elements)
     {
-        var organizationSpdxId = spdxPackage.GetOrganizationSpdxId(spdx30Elements);
+        var organizationSpdxId = spdxPackage.SuppliedBy;
         if (organizationSpdxId is null)
         {
             return null;
         }
 
         var organizationElement = spdx30Elements
-            .FirstOrDefault(element => element.SpdxId == organizationSpdxId);
+            .FirstOrDefault(element => element is Organization && element.SpdxId == organizationSpdxId);
         return organizationElement?.Name;
-    }
-
-    private static string GetOrganizationSpdxId(this Package spdxPackage, List<Element> spdx30Elements)
-    {
-        // Handle special case for the root package.
-        if (spdxPackage.SpdxId == Parsers.Spdx30SbomParser.Constants.RootPackageIdValue)
-        {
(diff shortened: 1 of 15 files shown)


src/Microsoft.Sbom.Parsers.Spdx30SbomParser/Utils/SPDXToSbomFormatConverterExtensions.cs
```diff
@@ -185,17 +185,32 @@ private static string GetSingleLicense(this Element element, RelationshipType re
...
     }
 
+    private static string GetOrganizationSpdxId(this Package spdxPackage, List<Element> spdx30Elements)
+    {
+        // Handle special case for the root package.
+        if (spdxPackage.SpdxId == Parsers.Spdx30SbomParser.Constants.RootPackageIdValue)
+        {
+            var creationInfo = spdx30Elements
+            .First(element => element is CreationInfo) as CreationInfo;
+            return creationInfo.CreatedBy.First();
```
DaveTryon, line 206: My first thought here was "Can `creationInfo` ever be null?" Searching for callers of this, I find only `SPDXToSbomFormatConverterExtensions.ToSbomPackage`, which only seems to be called from test code.

Are there customer scenarios where this would get called? It's ok if the answer is no, but I'm a bit curious about the intent of the change

pragnya17: The intent of this change is so that SPDX 3.0 elements can be converted to a standardized internal format, much like the SPDX 2.2 extensions that are a mirror of this. These extensions are then used in the SBOM diff checker - https://github.com/microsoft/sbom-tool/pull/1011 so that a given SPDX 2.2 file and SPDX 3.0 file can be compared based on content. 

Currently this would never be called by the customer.

I am rethinking this change being part of this PR though. Would it be helpful if I removed the extensions change as part of this PR?

pragnya17: I've separated the extensions changes in a separate PR - https://github.com/microsoft/sbom-tool/pull/1022. This can be reviewed later as well since it is not a release blocker.

test/Microsoft.Sbom.Api.Tests/Workflows/ManifestGenerationWorkflowTests.cs
```diff
@@ -401,7 +401,9 @@ await externalDocumentReferenceChannel.Writer.WriteAsync(new ExternalDocumentRef
             Assert.AreEqual(1, noAssertionElements.Count);
 
             var organizationElements = elements.Where(element => element["type"].ToString() == "Organization").ToList();
-            Assert.AreEqual(3, organizationElements.Count);
+            Assert.AreEqual(2, organizationElements.Count);
+            var organizationElementWithSupplierName = organizationElements.Where(element => element["name"].ToString() == "Organization: testSupplier").ToList();
+            var organizationElementWithNoAssertion = organizationElements.Where(element => element["name"].ToString() == "NOASSERTION").ToList();
```
DaveTryon, line 406: These variables don't seem to be used... not sure if you meant to remove them or possibly Assert their contents in some way?

pragnya17: They should be asserted, that was an oversight on my part.

test/Microsoft.Sbom.Api.Tests/Workflows/ManifestGenerationWorkflowTests.cs
```diff
@@ -402,9 +402,13 @@ await externalDocumentReferenceChannel.Writer.WriteAsync(new ExternalDocumentRef
 
             var organizationElements = elements.Where(element => element["type"].ToString() == "Organization").ToList();
             Assert.AreEqual(2, organizationElements.Count);
+
```
DaveTryon, line 405: (nit): Ordering here feels off. I'd expect us to consistently follow one of these patterns
1. All the vars get set, then all of the Asserts
2. Set 1 var, then Assert, (and probably leave an empty newline)

Asserts 1 and 4 follow pattern 1, while Asserts 2 and 3 follow pattern 2

DaveTryon (approved):

PR #1022, opened 2025-04-21
SPDX to internal SBOM conversion of supplier info for root packages (special case)

pragnya17:
In SPDX 3.0 the supplier information for root packages is stored in the CreationInfo element which is different from SPDX 2.2. This PR handles this special case and gets the supplier info correctly for SPDX 3.0 so that this information is present when converting to the internal SBOM format.


pragnya17:
These changes are already in PR #1011

PR #1030, opened 2025-04-29
Rename "Compliance" to "Conformance"

DaveTryon:
We've been asked to use the term "Conformance" instead of "Compliance", "ComplianceStandard", or "Compliance Standard" in our code and docs. Since this would be a breaking change, we want to squeeze it into the 4.x release.

After this change, the terms "compliance" has 0 case-insensitive matches in the code base.


docs/sbom-tool-api-reference.md
```diff
@@ -217,7 +217,7 @@ var file = new SbomFile
 };
 ```
 
-The API looks for a relative path starting with a period `.`.  All path separators should include forward slashes `/` in compliance with the SPDX version 2.2 specification.
+The API looks for a relative path starting with a period `.`.  All path separators should include forward slashes `/` in conformance with the SPDX version 2.2 specification.
```
pragnya17, line 220: This change can be reverted since it is unrelated to the SPDX 3.0 addition of compliance standard validation. This is something that already existed before my changes.

src/Microsoft.Sbom.Common/Conformance/ConformanceeExtensions.cs
```diff
@@ -1,9 +1,9 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-namespace Microsoft.Sbom.Common.ComplianceStandard;
+namespace Microsoft.Sbom.Common.Conformance;
 
-internal static class ComplianceExtensions
+internal static class ConformanceeExtensions
```
pragnya17, line 6:
```suggestion
internal static class ConformanceExtensions
```

DaveTryon: Thanks for catching! Fixed.

pragnya17 (approved):

PR #1033, opened 2025-05-02
Teach ManifestValidator about extensions

DaveTryon:
The `ManifestValidator` class works against a static list of manifest types. This constant list doesn't allow us to specify a ManifestInfo for a single extension, which is problematic. The `ManifestGeneratorProvider` class knows about all registered manifest types, so is it the authoritative list of valid types that should be accepted.

Validated with our internal client that has additional manifest types.

Note: I considered completely removing `Constants.SupportedSpdxManifests` completely, but it is used in both `ComponentDetectionBaseWalker` and `BaseManifestConfigHandler`. I wanted to keep this PR tightly focused, but we should consider removing this constant in the future if it's not too much effort.

### DIFF
--- a/src/Microsoft.Sbom.Api/Config/Validators/ManifestInfoValidator.cs
+++ b/src/Microsoft.Sbom.Api/Config/Validators/ManifestInfoValidator.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using Microsoft.Sbom.Api.Manifest;
 using Microsoft.Sbom.Api.Utils;
 using Microsoft.Sbom.Common.Config.Attributes;
 using Microsoft.Sbom.Common.Config.Validators;
@@ -17,17 +18,38 @@ namespace Microsoft.Sbom.Api.Config.Validators;
 /// </summary>
 public class ManifestInfoValidator : ConfigValidator
 {
-    public ManifestInfoValidator(IAssemblyConfig assemblyConfig)
-        : base(typeof(ValidManifestInfoAttribute), assemblyConfig)
+    private readonly HashSet<ManifestInfo> supportedManifestInfos;
+
+    /// <summary>
+    /// Production constructor (satisfied by dependency injection)
+    /// </summary>
+    public ManifestInfoValidator(IAssemblyConfig assemblyConfig, ManifestGeneratorProvider manifestGeneratorProvider)
+        : this(assemblyConfig, GetAvailableManifestInfos(manifestGeneratorProvider))
     {
+    }
+
+    /// <summary>
+    /// Test constructor
+    /// </summary>
+    public ManifestInfoValidator(IAssemblyConfig assemblyConfig, HashSet<ManifestInfo> supportedManifestInfos)
+    : base(typeof(ValidManifestInfoAttribute), assemblyConfig)
+    {
+        this.supportedManifestInfos = supportedManifestInfos ?? throw new ArgumentNullException(nameof(supportedManifestInfos));
     }
 
     public override void ValidateInternal(string paramName, object paramValue, Attribute attribute)
     {
-        if (paramValue is not null && paramValue is List<ManifestInfo> listOfManifestInfos && !Constants.SupportedSpdxManifests.Any(listOfManifestInfos.Contains))
+        if (paramValue is not null && paramValue is List<ManifestInfo> listOfManifestInfos && !supportedManifestInfos.Any(listOfManifestInfos.Contains))
         {
-            var supportedManifests = string.Join(", ", Constants.SupportedSpdxManifests.Select(m => m.ToString()));
-            throw new ValidationArgException($"Please provide a valid value for the ManifestInfo (-mi) parameter. Supported values include: {supportedManifests}. The values are case-insensitive.");
+            var providedValues = string.Join(", ", listOfManifestInfos);
+            var validManifestInfoa = string.Join(", ", supportedManifestInfos.Select(m => m.ToString()));
+            throw new ValidationArgException($"The value '{providedValues}' contains no values supported by the ManifestInfo (-mi) parameter. Please provide supported values. Supported values include: {validManifestInfoa}. The values are case-insensitive.");
         }
+    }
+
+    private static HashSet<ManifestInfo> GetAvailableManifestInfos(ManifestGeneratorProvider manifestGeneratorProvider)
+    {
+        ArgumentNullException.ThrowIfNull(manifestGeneratorProvider, nameof(manifestGeneratorProvider));
+        return [.. manifestGeneratorProvider.GetSupportedManifestInfos()];
     }
 }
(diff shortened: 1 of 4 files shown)


src/Microsoft.Sbom.Api/Config/Validators/ManifestInfoValidator.cs
```diff
@@ -17,17 +18,38 @@ namespace Microsoft.Sbom.Api.Config.Validators;
...
 
     public override void ValidateInternal(string paramName, object paramValue, Attribute attribute)
     {
-        if (paramValue is not null && paramValue is List<ManifestInfo> listOfManifestInfos && !Constants.SupportedSpdxManifests.Any(listOfManifestInfos.Contains))
+        if (paramValue is not null && paramValue is List<ManifestInfo> listOfManifestInfos && !supportedManifestInfos.Any(listOfManifestInfos.Contains))
         {
-            var supportedManifests = string.Join(", ", Constants.SupportedSpdxManifests.Select(m => m.ToString()));
-            throw new ValidationArgException($"Please provide a valid value for the ManifestInfo (-mi) parameter. Supported values include: {supportedManifests}. The values are case-insensitive.");
+            var providedValues = string.Join(", ", listOfManifestInfos);
+            var validManifestInfoa = string.Join(", ", supportedManifestInfos.Select(m => m.ToString()));
```
pragnya17, line 45:
```suggestion
            var validManifestInfos = string.Join(", ", supportedManifestInfos.Select(m => m.ToString()));
```

src/Microsoft.Sbom.Api/Config/Validators/ManifestInfoValidator.cs
```diff
@@ -17,17 +18,38 @@ namespace Microsoft.Sbom.Api.Config.Validators;
...
+            throw new ValidationArgException($"The value '{providedValues}' contains no values supported by the ManifestInfo (-mi) parameter. Please provide supported values. Supported values include: {validManifestInfoa}. The values are case-insensitive.");
```
pragnya17, line 46:
```suggestion
            throw new ValidationArgException($"The value '{providedValues}' contains no values supported by the ManifestInfo (-mi) parameter. Please provide supported values. Supported values include: {validManifestInfos}. The values are case-insensitive.");
```

src/Microsoft.Sbom.Api/Config/Validators/ManifestInfoValidator.cs
```diff
@@ -17,17 +18,38 @@ namespace Microsoft.Sbom.Api.Config.Validators;
...
         }
     }
+
+    private static HashSet<ManifestInfo> GetAvailableManifestInfos(ManifestGeneratorProvider manifestGeneratorProvider)
+    {
+        ArgumentNullException.ThrowIfNull(manifestGeneratorProvider, nameof(manifestGeneratorProvider));
+        return [.. manifestGeneratorProvider.GetSupportedManifestInfos()];
```
pragnya17, line 53: Would GetSupportedManifestInfos() also get supported manifest info values through dependency injection? For example, would this call return WdacCatalogManifestInfo that's injected in drop validator?

test/Microsoft.Sbom.Api.Tests/Config/Validators/ManifestInfoValidatorTests.cs
```diff
@@ -15,6 +17,11 @@ namespace Microsoft.Sbom.Api.Tests.Config.Validators;
 public class ManifestInfoValidatorTests
 {
     private readonly Mock<IAssemblyConfig> mockAssemblyConfig = new Mock<IAssemblyConfig>();
+    private readonly HashSet<ManifestInfo> supportedManifestInfosForTesting = new HashSet<ManifestInfo>
```
pragnya17, line 20: There's also a TestManifestInfo provided in Constants.cs, should that be added to this list for consistency?

test/Microsoft.Sbom.Tool.Tests/IntegrationTests.cs
```diff
@@ -202,7 +202,7 @@ public void E2E_Generate_WithBadManifestInfo_ReturnsNonZeroExitCode()
         var testFolderPath = CreateTestFolder();
         var arguments = $"generate -ps IntegrationTests -pn IntegrationTests -pv 1.2.3 -m \"{testFolderPath}\" -b \"{testDropDirectory}\" -bc \"{GetSolutionFolderPath()}\" -mi randomName:randomVersion";
         var (stdout, stderr, exitCode) = LaunchAndCaptureOutput(arguments);
-        Assert.AreEqual("Please provide a valid value for the ManifestInfo (-mi) parameter. Supported values include: SPDX:2.2, SPDX:3.0. The values are case-insensitive.\r\n", stderr);
+        Assert.AreEqual("The value 'randomName:randomVersion' contains no values supported by the ManifestInfo (-mi) parameter. Please provide supported values. Supported values include: SPDX:2.2, SPDX:3.0. The values are case-insensitive.\r\n", stderr);
```
pragnya17, line 205: We should probably remove this part since it doesn't include supported values that might come from dependency injection

test/Microsoft.Sbom.Tool.Tests/IntegrationTests.cs
```diff
@@ -221,7 +221,7 @@ public void E2E_Validate_WithBadManifestInfo_ReturnsNonZeroExitCode()
         var (arguments, outputFile) = GetValidateManifestArguments(testFolderPath, "randomVersion");
 
         var (stdout, stderr, exitCode) = LaunchAndCaptureOutput(arguments);
```
pragnya17, line 223: same comment as above

test/Microsoft.Sbom.Tool.Tests/IntegrationTests.cs
```diff
@@ -221,7 +221,7 @@ public void E2E_Validate_WithBadManifestInfo_ReturnsNonZeroExitCode()
...
-        Assert.AreEqual("Please provide a valid value for the ManifestInfo (-mi) parameter. Supported values include: SPDX:2.2, SPDX:3.0. The values are case-insensitive.\r\n", stderr);
+        Assert.AreEqual("The value 'SPDX:randomVersion' contains no values supported by the ManifestInfo (-mi) parameter. Please provide supported values. Supported values include: SPDX:2.2, SPDX:3.0. The values are case-insensitive.\r\n", stderr);
```
pragnya17, line 224: Same comment as above

pragnya17 (commented):
Left a few comments/minor fixes

test/Microsoft.Sbom.Tool.Tests/IntegrationTests.cs
DaveTryon: These tests only support the types that we define in sbom-tool. We'll update it if/when we change the built-in types

pragnya17 (approved):

PR #1034, opened 2025-05-03
Avoid Exception if an unsupported format is requested

DaveTryon:
If a config file is used, it is possible to pass in a request for a manifest type that we don't support. When this happens, the code throws an exception in `SbomGenerationWorkflow.GenerateAsync()`. This PR uses a `TryGet` pattern to more gracefully detect this situation and display a warning so the caller can decide to fix it if they want. It also adds code to return false if none of manifest types in the config file are supported. Unit tests cover both supported and unsupported manifest types.

### DIFF
--- a/test/Microsoft.Sbom.Api.Tests/Workflows/ManifestGenerationWorkflowTests.cs
+++ b/test/Microsoft.Sbom.Api.Tests/Workflows/ManifestGenerationWorkflowTests.cs
@@ -499,6 +499,7 @@ public class ManifestGenerationWorkflowTests
 
         var result = await workflow.RunAsync();
 
+        Assert.IsFalse(result);
         fileArrayGeneratorMock.VerifyAll();
         packageArrayGeneratorMock.VerifyAll();
         relationshipsArrayGeneratorMock.VerifyAll();
--- a/test/Microsoft.Sbom.Api.Tests/Workflows/ManifestGenerationWorkflowTests.cs
+++ b/test/Microsoft.Sbom.Api.Tests/Workflows/ManifestGenerationWorkflowTests.cs
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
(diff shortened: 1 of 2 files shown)


test/Microsoft.Sbom.Api.Tests/Workflows/ManifestGenerationWorkflowTests.cs
```diff
@@ -453,7 +454,7 @@ public async Task ManifestGenerationWorkflowTests_SBOMDir_NotDefault_NotDeleted(
         using var manifestStream = new MemoryStream();
 
         fileSystemMock.Setup(f => f.DirectoryExists(It.IsAny<string>())).Returns(true);
-        var sbomConfig = new SbomConfig(fileSystemMock.Object)
+        ISbomConfig sbomConfig = new SbomConfig(fileSystemMock.Object)
```
DaveTryon, line 457: Changed 'var' to 'ISbomConfig` because `Mock.Setup` with an `out` parameter is very finicky about exactly matching the type

test/Microsoft.Sbom.Api.Tests/Workflows/ManifestGenerationWorkflowTests.cs
```diff
@@ -482,7 +483,51 @@ public async Task ManifestGenerationWorkflowTests_SBOMDir_NotDefault_NotDeleted(
...
+            mockOSUtils.Object,
+            recorderMock.Object);
+
+        var result = await workflow.RunAsync();
+
+        fileArrayGeneratorMock.VerifyAll();
+        packageArrayGeneratorMock.VerifyAll();
+        relationshipsArrayGeneratorMock.VerifyAll();
+        externalDocumentReferenceGeneratorMock.VerifyAll();
+        fileSystemMock.Verify(f => f.DeleteDir(It.IsAny<string>(), true), Times.Once);
```
pragnya17, line 506: Should we remove `Assert.IsFalse(result);` ? It is present in the previous code, not sure of the rationale behind removing that assert statement

DaveTryon: Added 😃

pragnya17 (approved):

PR #1043, opened 2025-05-09
Generate only supported manifests, get target configs, and use SourcesProviders as the source of truth

pragnya17:
Only iterate through sourcesProviders once and use that as a source of truth to generate all requested manifests. In previous code we were iterating through sourcesProviders once per manifest. This creates performance and data reliability issues.
For example, when we were trying to generate SPDX 2.2 and WdacCatalog manifest files through the drop validator task, which both use the SPDX 2.2 format, the WdacCatalog file was missing data. This data inconsistency was because we were asynchronously iterating through the sourcesProviders for each manifest. This is also a performance issue since we only need to iterate sourcesProviders and use that as a source of truth to generate all requested manifests.

Generate only manifests that are provided in the config rather than all supported configs. If the config is not supported log a warning instead of throwing an exception. This builds on the work that was already done in https://github.com/microsoft/sbom-tool/commit/6fdf28debb12e489b6a1fd995deb8dab326a8a33

Generation is updated to use target configs instead of target manifests. Rather than passing manifests and getting configs based on that, we get the target configs once during SbomGenerationWorkflow and then use that for all the GenerateAsync() method. 

Use generics like IEnumerable and ISet where applicable

### DIFF
--- a/src/Microsoft.Sbom.Api/Manifest/Configuration/SbomConfigProvider.cs
+++ b/src/Microsoft.Sbom.Api/Manifest/Configuration/SbomConfigProvider.cs
@@ -127,8 +127,7 @@ public class SbomConfigProvider : ISbomConfigProvider
     {
         foreach (var manifestInfo in manifestInfosFromConfig)
         {
-            var supportedManifestInfo = this.TryGet(manifestInfo, out var config);
-            if (supportedManifestInfo)
+            if (this.TryGet(manifestInfo, out var config))
             {
                 config.StartJsonSerialization();
             }
--- a/src/Microsoft.Sbom.Api/Manifest/Configuration/SbomConfigProvider.cs
+++ b/src/Microsoft.Sbom.Api/Manifest/Configuration/SbomConfigProvider.cs
@@ -118,15 +118,22 @@ public class SbomConfigProvider : ISbomConfigProvider
         return this;
     }
 
-    public IAsyncDisposable StartJsonSerializationAsync()
+    /// <summary>
+    /// Starts asynchronous JSON serialization of supported ISbomConfig objects from IConfiguration.
+    /// </summary>
+    /// <param name="manifestInfosFromConfig"></param>
+    /// <returns></returns>
+    public IAsyncDisposable StartJsonSerializationAsync(IList<ManifestInfo> manifestInfosFromConfig)
     {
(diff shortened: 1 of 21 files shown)


src/Microsoft.Sbom.Api/Manifest/Configuration/SbomConfigProvider.cs
```diff
@@ -118,15 +118,18 @@ public IDisposable StartJsonSerialization()
         return this;
     }
 
-    public IAsyncDisposable StartJsonSerializationAsync()
+    /// <summary>
+    /// Starts asynchronous JSON serialization of supported ISbomConfig objects from IConfiguration.
+    /// </summary>
+    /// <param name="manifestInfosFromConfig"></param>
```
DaveTryon, line 124:
```suggestion
    /// <param name="targetConfigs"></param>
```

src/Microsoft.Sbom.Api/Workflows/SbomGenerationWorkflow.cs
```diff
@@ -96,45 +98,56 @@ public virtual async Task<bool> RunAsync()
                     log.Information("Manifest directory path was explicitly defined. Will not attempt to delete any existing _manifest directory.");
                 }
 
-                var manifestInfos = configuration.ManifestInfo.Value;
+                // Write manifests based on manifestInfo values in the configuration.
+                var manifestInfosFromConfiguration = configuration.ManifestInfo.Value;
```
DaveTryon, line 102: Let's avoid making this a variable:

```
var targetConfigs = GetTargetConfigs(configuration.ManifestInfo.Value);
```

src/Microsoft.Sbom.Api/Workflows/SbomGenerationWorkflow.cs
```diff
@@ -180,6 +193,37 @@ public virtual async Task<bool> RunAsync()
         }
     }
 
+    public IEnumerable<ISbomConfig> GetTargetConfigs(IEnumerable<ManifestInfo> manifestInfosFromConfiguration)
```
DaveTryon, line 196: Should be private
```suggestion
    private IEnumerable<ISbomConfig> GetTargetConfigs(IEnumerable<ManifestInfo> manifestInfosFromConfiguration)
```

DaveTryon (approved):

PR #1047, opened 2025-05-13
remove redundant dictionary lookups

SimonCropp:


src/Microsoft.Sbom.Api/Utils/ComponentDetectorCachedExecutor.cs
```diff
@@ -42,13 +42,13 @@ public virtual async Task<ScanResult> ScanAsync(ScanSettings args)
 
         var scanSettingsHash = args.ToString().GetHashCode();
```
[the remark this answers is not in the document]
SimonCropp: perhaps raise an issue to discuss

src/Microsoft.Sbom.Common/Extensions/DictionaryExtensions.cs
```diff
@@ -38,9 +38,9 @@ public static void AddIfKeyNotPresentAndValueNotNull<TKey, TValue>(
 
         // Check if the value is not null, and if the dictionary already
         // contains a value for the key.
-        if (value != null && !dictionary.ContainsKey(key))
+        if (value != null)
         {
-            dictionary.Add(key, value);
```
DaveTryon, line 43: This doesn't remove a lookup--it simply hides it inside an extension method that does exactly the same thing. I've included the code for `TryAdd` below.

The other changes legitimately reduce redundancy. This one does not. Can we please back it out?

```
        public static bool TryAdd<TKey, TValue>(this IDictionary<TKey, TValue> dictionary, TKey key, TValue value)
        {
            if (dictionary is null)
            {
                ThrowHelper.ThrowArgumentNullException(ExceptionArgument.dictionary);
            }

            if (!dictionary.ContainsKey(key))
            {
                dictionary.Add(key, value);
                return true;
            }

            return false;
        }
```

SimonCropp: that is correct. but is is less code, and the other changes do reduce lookups.

would you like me to remove this change, even though it is less code and functionally equivalent?

DaveTryon: Yes, please

SimonCropp: reverted.

although note on newer runtimes (net core 2 and up) thas TryAdd is actually a first class method on dicitonary and does have better performance

https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2.tryadd?view=net-9.0
the resulting method called is this:
https://github.com/dotnet/runtime/blob/1d1bf92fcf43aa6981804dc53c5174445069c9e4/src/libraries/System.Private.CoreLib/src/System/Collections/Generic/Dictionary.cs#L1061C13-L1061C58

DaveTryon: That may be true, but the type is `IDictionary`, not `Dictionary`, so we don't know about the first class method

PR #1053, opened 2025-05-15
Update MSTest and use Combinatorial.MSTest

Youssef1313:
This PR updates MSTest to current latest stable, simplifes TrxReport and CC references which are now transitive dependency of MSTest metapackage, and also uses https://github.com/Youssef1313/Combinatorial.MSTest (**NOTE:** *The package is **not** maintained by Microsoft*) to simplify `ManifestGenerationWorkflowTests_Succeeds`

### DIFF
--- a/test/Directory.Build.props
+++ b/test/Directory.Build.props
@@ -12,9 +12,8 @@
   </PropertyGroup>
   
   <ItemGroup Label="Package References">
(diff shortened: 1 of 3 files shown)


DaveTryon:
We'll pass on this one. Thanks, @Youssef1313!

PR #1063, opened 2025-05-19
explicitly reference Serilog

SimonCropp:
given downstream extensions to Serilog do not deploy when a new version of Serilog is deployed, it is better to explicitly reference Serilog so we get the current stable version


DaveTryon (approved):
This wouldn't be necessary if Serilog referenced the appropriate version in their `.nuspec` files, but since we don't control that...

PR #1064, opened 2025-05-19
use collection expressions

SimonCropp:


DaveTryon (approved):
Thanks, @SimonCropp! The title on this intrigued me, since I've always considered this sort of change as syntactic sugar without perf impact. I used https://sharplab.io/ to compare the lowered code of the various initialization options:

Original code
```
public class C {
    public static void PerfTest()
    {
        var test1 = new string[]{"abc", "def", "ghi"};
        var test2 = new []{"abc", "def", "ghi"};
        string[] test3 = ["abc", "def", "ghi"];
    }
}
```

Lowered code
```
public class C
{
    public static void PerfTest()
    {
        string[] array = new string[3];
        array[0] = "abc";
        array[1] = "def";
        array[2] = "ghi";
        string[] array2 = new string[3];
        array2[0] = "abc";
        array2[1] = "def";
        array2[2] = "ghi";
        string[] array3 = new string[3];
        array3[0] = "abc";
        array3[1] = "def";
        array3[2] = "ghi";
    }
}
```

Lowered code is identical, so perf impact is 0. I've updated the title accordingly.

SimonCropp:
@DaveTryon it depends on the target type. so for example if you do

```
IEnumerable<int> value = [1, 2, 3, 4];
```

then u get 
```
IEnumerable<int> value = new global::<>z__ReadOnlyArray<int>(new int[4] { 1, 2, 3, 4 });
```
where `z__ReadOnlyArray` is an optimized implementation of  IEnumerable for a small number of items.

if u do

```
IEnumerable<int> value = [1];
```
then u get

```
IEnumerable<int> value = new global::<>z__ReadOnlySingleElementList<int>(1);
```

where `z__ReadOnlySingleElementList` is an optimized implementation of  IEnumerable for 1 item


so it can absolutely have perf impacts

SimonCropp:
in this pr we are asigning to an `IEnumerable<Checksum>` so there will be perf improvements

PR #1067, opened 2025-05-19
fix some incorrect as usages in GetDocumentNamespaceUri

SimonCropp:
the assumption seems to be that these values must be non nullable strings. since if they are null or not strings, it would result in a null ref inside EscapeDataString.

so it is better to do a cast. and if the assumption is ever incorrect, the exception will occur earlier and be more accurate


DaveTryon:
Thanks, @SimonCropp, but I think we'll pass on this one. The code that populates the Dictionary omits the key if the value is null, so the Dictionary would still throw before this change takes effect. Yes, the Dictionary is unfortunately public and changeable by other code, but fixing that impacts the API, which is a bigger change than this potential issue warrants.

PR #1068, opened 2025-05-19
remove string allocations in OSUtils usage of GetEnvironmentVariables

SimonCropp:


src/Microsoft.Sbom.Common/OSUtils.cs
```diff
@@ -35,7 +35,7 @@ public OSUtils(ILogger logger, IEnvironmentWrapper environment)
 
         foreach (DictionaryEntry de in this.environment.GetEnvironmentVariables())
         {
-            environmentVariables.Add(de.Key.ToString(), de.Value.ToString());
```
DaveTryon, line 38: Thanks, @SimonCropp, but there is no allocation here. If, as we expect, the keys and values are strings, then `System.String.ToString` will be called, which is implemented as follows:

```
public override string ToString()
{
    return this;
}
```

If, however, the key or value are unexpected some other type, this code normalizes them correctly. Simply casting them is not a safe operation and will throw an exception.

Bottom line, the code is doing exactly what we want it to do.

PR #1073, opened 2025-05-21
use String.Concat over string.Join(string.Empty,

SimonCropp:


DaveTryon (approved):
TIL I learned that there is a `string.Concat` that accepts an `IEnumerable`

PR #1075, opened 2025-05-21
Add IsPackable to target condition

### LINKED ISSUE #1074: [Microsoft.Sbom.Targets] GenerateSbomTarget should not run when IsPackable is not true
Currently the condition for the `GenerateSbomTarget` target is `'$(GenerateSBOM)' ==  'true'`.

This condition is not complete enough to avoid problems. Even when the `GenerateSBOM` property is true, if a project has `IsPackable` set to `false`, `GenerateSbomTarget` should not run.

 For example, NuGet's own `GenerateNuspec` has a condition that evaluates `IsPackable`: [link](https://github.com/NuGet/NuGet.Client/blob/0ccca7268b42aafa776420f04aad0612fc16fb7d/src/NuGet.Core/NuGet.Build.Tasks.Pack/NuGet.Build.Tasks.Pack.targets#L212)

Without checking `IsPackable`, it is possible to have the `GenerateSbomTarget` run in a project that has not produced a NuGet package, and then the target fails because it's trying to unzip the contents of a package that doesn't exist.

I encountered this problem when I added the Microsoft.Sbom.Targets package to my packaging tooling. This results in the package being referenced transitively from the tooling package. This tooling package is often references globally in a solution. The tooling then sets the `GenerateSBOM` to true so that by default, a package will have an SBOM.

However, it's often the case where some of the projects in a solution don't need a package, so they have `IsPackable` set to `false` in the project files. These projects fail to build properly now that I've added SBOM generation.

While I've been able to work around this problem for now by also setting `GenerateSBOM` to false in the affected projects, it's not reasonable to have this packaging "implementation detail" leak into the project file when I've already told it not to create a NuGet package.

bording:
This PR adds `'$(IsPackable)' == 'true'` to the condition that controls if `GenerateSbomTarget` will run.

Fixes #1074

### DIFF
--- a/src/Microsoft.Sbom.Targets/Microsoft.Sbom.Targets.targets
+++ b/src/Microsoft.Sbom.Targets/Microsoft.Sbom.Targets.targets
@@ -35,7 +35,7 @@
   </PropertyGroup>
 
   <!-- After the Nuget Package is generated, we will unzip, scan, generate the SBOM and zip again. -->
-  <Target Name="GenerateSbomTarget" AfterTargets="Pack" Condition=" '$(GenerateSBOM)' ==  'true'" >
+  <Target Name="GenerateSbomTarget" AfterTargets="Pack" Condition="'$(IsPackable)' == 'true' AND '$(GenerateSBOM)' ==  'true'" >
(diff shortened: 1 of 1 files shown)


baronfel:
This makes total sense, good call-out and good fix.

PR #1077, opened 2025-05-23
Remove lazy reflection

SimonCropp:
so the several uses of this pattern surprised me
```
private static readonly Lazy<string> VersionValue = new Lazy<string>(() => typeof(SbomToolCmdRunner).Assembly.GetCustomAttribute<AssemblyInformationalVersionAttribute>()?.InformationalVersion ?? string.Empty);
```

given the reflection perf, i did not think Lazy would speed anything up. given the slight overhead of Lazy compared to the cost of reflection. so i did some simplistic timings
# Raw

## Init

```
var watch = Stopwatch.StartNew();
var value = typeof(SbomToolCmdRunner).Assembly.GetCustomAttribute<AssemblyInformationalVersionAttribute>()?.InformationalVersion ?? string.Empty;
Console.WriteLine(watch.Elapsed);
```

00:00:00.0000279 (.0279 ms)

## Subsequent

this is putting a static pointer the stack. so single digit nanoseconds

# Lazy

## Init

```
var watch = Stopwatch.StartNew();
var lazy = new Lazy<string>(() => typeof(SbomToolCmdRunner).Assembly.GetCustomAttribute<AssemblyInformationalVersionAttribute>()?.InformationalVersion ?? string.Empty);
Console.WriteLine(watch.Elapsed);
```

00:00:00.0000036 (.0036 ms)

## Subsequent

```
var lazy = new Lazy<string>(() => typeof(SbomToolCmdRunner).Assembly.GetCustomAttribute<AssemblyInformationalVersionAttribute>()?.InformationalVersion ?? string.Empty);
var value1 = lazy.Value;
var watch = Stopwatch.StartNew();
var value2 = lazy.Value;
Console.WriteLine(watch.Elapsed);
```

00:00:00.0000024 (.0024 ms)

# Outcome

So it seems we are saving ~.0243ms at startup for an additional cost of ~.002 ms per subsequent call.

IMO this is not a good tradeoff.

But given the above numbers above are so small, i dont think this is really a performance conversation, IMO use of Lazy to defer a single static reflection call is a premature optimization

### DIFF
--- a/src/Microsoft.Sbom.Api/Config/ConfigSanitizer.cs
+++ b/src/Microsoft.Sbom.Api/Config/ConfigSanitizer.cs
@@ -28,9 +28,7 @@ public class ConfigSanitizer
     private readonly IFileSystemUtils fileSystemUtils;
     private readonly IAssemblyConfig assemblyConfig;
 
-    internal static string SbomToolVersion => VersionValue.Value;
-
-    private static readonly Lazy<string> VersionValue = new Lazy<string>(() => typeof(SbomToolCmdRunner).Assembly.GetCustomAttribute<AssemblyInformationalVersionAttribute>()?.InformationalVersion ?? string.Empty);
+    internal static string SbomToolVersion { get; } = typeof(SbomToolCmdRunner).Assembly.GetCustomAttribute<AssemblyInformationalVersionAttribute>()?.InformationalVersion ?? string.Empty;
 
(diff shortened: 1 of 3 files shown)


SimonCropp:
> Arguably true, but not enough of a gain to make me want to pick this one up.

thats surprises me, the existing code is unnecessary complex, and incorrectly implies that the performance characteristics of the reflection code is high. 

the resulting code is smaller, easier to understand and faster. seems an easy win

PR #1078, opened 2025-05-23
move to sdk 9

SimonCropp:
to avoid using an sdk with a cve
https://github.com/dotnet/announcements/issues/329


SimonCropp:
just to be clear. this isnt “about sticking to net8”. this is “about sticking to net8 sdk”. do u have reasons for sticking to net8 sdk. if so i would be really interested in hearing any of them

PR #1079, opened 2025-05-23
remove this prefix requirement

SimonCropp:
seems the majority of the code does not use the `this.` qualifier

image

using static analysis, 

 * 2218 without the use of `this.`
 * 2139 with the use of `this.`

I would like to settle on one and make it an error, or at least a warning. noting that this pr just makes it a suggestion.


DaveTryon:
FWIW, I agree with you--having `this.` littered around the code unnecessarily annoys me no end. I even have a local branch where I've made the changes to remove them. That said, other people on the team like to have the `this.`, and we aren't going to try to force either standard.

PR #1080, opened 2025-05-23
use using declarations

SimonCropp:
this might be a personal preference, but IMO the less indentation using declarations make the code easier to read


SimonCropp:
it also seems using declarations are used in many other places in the codebase

DaveTryon:
Usage is mixed, depending on who wrote the code. My sense is that more places use the braces than don't. We'll pass on trying to enforce this at this time

PR #1101, opened 2025-06-16
Fix for package dependency bug

pragnya17:
This PR aims to fix a bug that affects SBOM accuracy and determinism during generation. The bug's impact means that all the SBOMs we have been generating since this PR () was merged are not fully representative of the build drop path and are not deterministic. We are discarding a large proportion of our transitive dependency information, and the transitive dependencies that are selected for inclusion may vary between SBOM creations for the same drop. 

The root cause was that the package info created from the component detection results was incorrect. We were only getting the FirstOrDefault() AncestralReferrer for each package, which was not deterministic and did not consider the case of multiple referrers. 

This PR gets all the package dependencies from the component detection results and translates them correctly into package dependency relationships in the SBOM.

### DIFF
--- a/src/Microsoft.Sbom.Api/Executors/PackageInfoJsonWriter.cs
+++ b/src/Microsoft.Sbom.Api/Executors/PackageInfoJsonWriter.cs
@@ -66,7 +66,13 @@ public class PackageInfoJsonWriter
             {
                 var generationResult =
                     manifestGeneratorProvider.Get(sbomConfig.ManifestInfo).GenerateJsonDocument(packageInfo);
-                sbomConfig.Recorder.RecordPackageId(generationResult?.ResultMetadata?.EntityId, generationResult?.ResultMetadata?.DependOn);
+
+                // todo: move the inner for loop to here, that way you can pass in just a string instead of a list of strings for the second param
+                foreach (var dependency in generationResult?.ResultMetadata?.DependOn)
+                {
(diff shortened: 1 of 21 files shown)


src/Microsoft.Sbom.Api/Executors/PackageInfoJsonWriter.cs
```diff
@@ -66,7 +66,15 @@ private async Task GenerateJson(
...
                     manifestGeneratorProvider.Get(sbomConfig.ManifestInfo).GenerateJsonDocument(packageInfo);
-                sbomConfig.Recorder.RecordPackageId(generationResult?.ResultMetadata?.EntityId, generationResult?.ResultMetadata?.DependOn);
+
+                if (generationResult?.ResultMetadata?.DependOn != null)
+                {
+                    foreach (var dependency in generationResult?.ResultMetadata?.DependOn)
+                    {
+                        sbomConfig.Recorder.RecordPackageId(generationResult?.ResultMetadata?.EntityId, dependency);
+                    }
+                }
```
pragnya17, line 76: This is the only change to the API method. Since the method header itself has not changed, I think this is a non-breaking API change but would appreciate a double confirmation.

DaveTryon: It's technically still a breaking change, since it's changing a type in the data contract. That said, we don't really expect people to interact with this type directly, so I think we can treat it as non-breaking.

DaveTryon (approved):

PR #1105, opened 2025-06-24
Expand env vars included in input config files

sfoslund:
This PR adds logic to expand any env vars that are included in input config files provided via the `--ConfigFilePath` argument. The motivation behind this change is to allow users to include env vars in the consolidate verb's config file, which is not fully hooked up yet, but once we finish hooking it up we should get this behavior for all verbs for free

### DIFF
--- a/src/Microsoft.Sbom.Api/Config/ConfigFileParser.cs
+++ b/src/Microsoft.Sbom.Api/Config/ConfigFileParser.cs
@@ -29,6 +29,11 @@ public class ConfigFileParser
         }
 
         var content = await fileSystemUtils.ReadAllTextAsync(filePath);
+        if (string.IsNullOrEmpty(content))
+        {
+            return new ConfigFile();
+        }
+
         var expandedContent = ExpandEnvironmentVariablesInString(content);
         return JsonSerializer.Deserialize<ConfigFile>(expandedContent);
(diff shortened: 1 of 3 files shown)


test/Microsoft.Sbom.Api.Tests/Config/ConfigFileParserTests.cs
```diff
@@ -0,0 +1,52 @@
...
+    private ConfigFileParser testSubject;
+    private readonly string filePathStub = "test-path";
+    private readonly string contentStub = "{\"PackageSupplier\": \"TestSupplier\",\"BuildDropPath\": \"$(BuildDropPathEnvVar)\"}";
+    private readonly string envVarName = "BuildDropPathEnvVar";
+    private readonly string envVarValue = "TestPath";
+
+    [TestInitialize]
+    public void Initialize()
+    {
+        mockFileSystemUtils = new Mock<IFileSystemUtils>();
```
DaveTryon: I think this can be strict?
```suggestion
        mockFileSystemUtils = new Mock<IFileSystemUtils>(MockBehavior.Strict);
```

DaveTryon (approved):

DaveTryon (approved):

PR #1104, opened 2025-06-24
Add skeleton for consolidation action

DaveTryon:
This PR adds the skeleton for the consolidation action, so that we can have multiple people work on it.
In scope for this PR:
- Refactor command line arguments
- Add new action
- Plumbing to get action to run

Not in scope:
- Add consolidation-specific args to the config (and their unit tests)
- Implement the actual consolidation logic

PR #1106, opened 2025-06-24
Complete the stubbed plumbing for Consolidate action

DaveTryon:
PR #1104 compiled and passed all E2E tests, but was unable to get code into the `SbomConsolidationWorkflow` class. This adds more plumbing so that when we execute `sbomtool.exe consolidate`, we get into the correct workflow. I'll add the config mapping in a subsequent PR.

I discovered with this PR that we have separate implementations of *Service.cs in the `Microsoft.Sbom.Tool` and `DotNetTool` projects. Most of the existing duplicated classes are _mostly_ in sync, with some larger differences in ValidationService.cs. This feels like a significant violation of the DRY principle, but untangling it right is out of scope. The same `Program.cs` file gets used in both projects, so maybe that's a potential future pattern to sort things out.

### DIFF
--- a/src/Microsoft.Sbom.DotNetTool/ConsolidationService.cs
+++ b/src/Microsoft.Sbom.DotNetTool/ConsolidationService.cs
@@ -1,0 +1,59 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Hosting;
+using Microsoft.Sbom.Api;
+using Microsoft.Sbom.Api.Exceptions;
+using Microsoft.Sbom.Api.Output.Telemetry;
+using Microsoft.Sbom.Api.Workflows;
+
+namespace Microsoft.Sbom.Tool;
+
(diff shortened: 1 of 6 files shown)


src/Microsoft.Sbom.DotNetTool/ConsolidationService.cs
```diff
@@ -0,0 +1,59 @@
+// Copyright (c) Microsoft. All rights reserved.
```
sfoslund, line 1: Is there any difference here vs the other ConsiolidationService.cs added in the previous PR? If not, can we just [reference that one in this proj similarly to how we reference program.cs](https://github.com/microsoft/sbom-tool/blob/main/src/Microsoft.Sbom.DotNetTool/Microsoft.Sbom.DotNetTool.csproj#L12)?

DaveTryon: The ones in the `Microsoft.SbomTool` project all include `Environment.Exit` at the end of the task, while none of the ones in the `Microsoft.DotNetTool` project have that line. I think they were originally the same, then the exit statements were added in a later PR. I'm hesitant to change the existing pattern without fully understanding how it would impact the E2E

sfoslund (approved):

PR #1108, opened 2025-06-25
Add SPDXFormatDetector for SPDX version detection

sfoslund:
This PR adds a SPDXFormatDetector class which can be used to determine which version of SPDX a given SBOM uses. This class is not currently leveraged anywhere but it will be used in the new consolidate verb.

### DIFF
--- a/src/Microsoft.Sbom.Api/Utils/SPDXFormatDetector.cs
+++ b/src/Microsoft.Sbom.Api/Utils/SPDXFormatDetector.cs
@@ -1,0 +1,124 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using Microsoft.Sbom.Api.Manifest;
+using Microsoft.Sbom.Common;
+using Microsoft.Sbom.Extensions.Entities;
+using Microsoft.Sbom.JsonAsynchronousNodeKit;
+using Microsoft.Sbom.Parser;
+
+namespace Microsoft.Sbom.Api.Utils;
+
+public class SPDXFormatDetector : ISPDXFormatDetector
+{
+    private readonly IFileSystemUtils fileSystemUtils;
+    private readonly IManifestParserProvider manifestParserProvider;
+    private readonly IDictionary<ManifestInfo, Func<ISbomParser, Stream, bool>> supportedManifestInfos;
+
+    public SPDXFormatDetector(
+        IFileSystemUtils fileSystemUtils,
+        IManifestParserProvider manifestParserProvider)
+    {
+        this.fileSystemUtils = fileSystemUtils;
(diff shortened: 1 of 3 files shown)


src/Microsoft.Sbom.Api/Utils/SPDXFormatDetector.cs
```diff
@@ -0,0 +1,113 @@
...
+            do
+            {
+                result = parser.Next();
+
+                if (result is not null && result.Result is not null)
+                {
+                    switch (result.FieldName)
+                    {
+                        case Constants.SPDXContextHeaderName:
+                            var contextReslt = (result as ContextsResult)?.Contexts.FirstOrDefault();
```
DaveTryon: Typo: `contextReslt` should be `contextResult` (with a 'u')

src/Microsoft.Sbom.Api/Utils/SPDXFormatDetector.cs
```diff
@@ -0,0 +1,113 @@
...
+    {
+        try
+        {
+            var parser = new SPDX30Parser(stream);
+            ParserStateResult? result = null;
+            do
+            {
+                result = parser.Next();
+
+                if (result is not null && result.Result is not null)
```
DaveTryon: Null propagation?
```suggestion
                if (result?.Result is not null)
```

test/Microsoft.Sbom.Api.Tests/Utils/SPDXFormatDetectorTests.cs
```diff
@@ -0,0 +1,101 @@
...
+        mock22ManifestInterface.Setup(m => m.CreateParser(It.IsAny<Stream>())).Returns((Stream stream) => new SPDXParser(stream));
+        mock30ManifestInterface.Setup(m => m.CreateParser(It.IsAny<Stream>())).Returns((Stream stream) => new SPDX30Parser(stream));
+
+        testSubject = new SPDXFormatDetector(mockFileSystemUtils.Object, mockManifestParserProvider.Object);
+    }
+
+    [TestCleanup]
+    public void Verify()
+    {
+        mockFileSystemUtils.VerifyAll();
```
DaveTryon, line 51: Is there merit in callindg `VerifyAll` on the other mocks? That means calling some of the `.Setup` methods in the unit tests, but that would make the tests more focused. Your call...

sfoslund: I considered it but IMO it's not worth making the tests uglier/ harder to read, but if you have strong feelings happy to change this

DaveTryon (approved):

test/Microsoft.Sbom.Api.Tests/Utils/SPDXFormatDetectorTests.cs
DaveTryon: All good...

PR #1110, opened 2025-06-25
Add config file for Consolidate action

DaveTryon:
This Adds the config file for the Consolidate verb. The key new element is the ArtifactInfoMap, which `ConfigSanitizer` will now reject for the `consolidate` action if it's malformed.

This touches `IConfiguration` so it's technically an API breaking change, but I'm not sure we need to worry about that one. It'll be a minor version bump either way when we release the `consolidate` action (and we'll also need to add it to the API, but we have a separate story for that and won't merge it until after we have a better idea of the shape of the data.


DaveTryon:
Trying to jump-start CLA check

PR #1111, opened 2025-06-26
remove broken param docs

SimonCropp:


src/Microsoft.Sbom.Api/Config/ApiConfigurationBuilder.cs
```diff
@@ -30,7 +30,6 @@ public static class ApiConfigurationBuilder
     /// <param name="metadata"></param>
     /// <param name="specifications"></param>
     /// <param name="runtimeConfiguration"></param>
-    /// <param name="verbosity"></param>
```
pragnya17, line 33: Could you also please add the last two missing params - externalDocumentReferenceListFile and componentPath?

SimonCropp: whats the point of having empty param docs?

PR #1115, opened 2025-06-26
Adding validate plumbing to consolidate verb

sfoslund:
This PR adds some additional plumbing to the consolidate verb to parse the user provided ArtifactInfoMap and find all valid SBOMs that need to be validated/ consolidated. This PR has logic to account for a variety of flavors of SBOMs- spdx 2.2/3.0 and ADO/ CloudBuild

### DIFF
--- a/src/Microsoft.Sbom.Api/Manifest/Configuration/SbomConfigFactory.cs
+++ b/src/Microsoft.Sbom.Api/Manifest/Configuration/SbomConfigFactory.cs
@@ -2,9 +2,11 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using Microsoft.Sbom.Api.Recorder;
 using Microsoft.Sbom.Common;
 using Microsoft.Sbom.Extensions;
 using Microsoft.Sbom.Extensions.Entities;
+using Constants = Microsoft.Sbom.Api.Utils.Constants;
 
 namespace Microsoft.Sbom.Api.Manifest.Configuration;
 
@@ -39,4 +41,29 @@ public class SbomConfigFactory : ISbomConfigFactory
             Recorder = recorder
         };
     }
+
+    public ISbomConfig Get(
+        ManifestInfo manifestInfo,
+        string manifestPath,
+        IMetadataBuilderFactory metadataBuilderFactory)
+    {
+        var sbomDirPath = GetSbomDirPath(manifestPath, manifestInfo);
+        var sbomFilePath = GetSbomFilePath(manifestPath, manifestInfo);
+        return Get(manifestInfo,
+            manifestPath,
+            sbomFilePath,
+            $"{sbomFilePath}.sha256",
+            fileSystemUtils.JoinPaths(sbomDirPath, Constants.CatalogFileName),
+            fileSystemUtils.JoinPaths(sbomDirPath, Constants.BsiFileName),
+            new SbomPackageDetailsRecorder(),
+            metadataBuilderFactory.Get(manifestInfo));
+    }
+
+    public string GetSbomDirPath(string manifestDirPath, ManifestInfo manifestInfo) => fileSystemUtils.JoinPaths(
+        manifestDirPath,
+        $"{manifestInfo.Name.ToLower()}_{manifestInfo.Version.ToLower()}");
+
+    public string GetSbomFilePath(string manifestDirPath, ManifestInfo manifestInfo) => fileSystemUtils.JoinPaths(
+        GetSbomDirPath(manifestDirPath, manifestInfo),
+        $"manifest.{manifestInfo.Name.ToLower()}.json");
 }
--- a/src/Microsoft.Sbom.Api/Manifest/Configuration/SbomConfigFactory.cs
+++ b/src/Microsoft.Sbom.Api/Manifest/Configuration/SbomConfigFactory.cs
@@ -47,23 +47,34 @@ public class SbomConfigFactory : ISbomConfigFactory
         string manifestPath,
         IMetadataBuilderFactory metadataBuilderFactory)
     {
-        var sbomDirPath = GetSbomDirPath(manifestPath, manifestInfo);
+        var sbomDirPath = GetSpdxDirPath(manifestPath, manifestInfo);
         var sbomFilePath = GetSbomFilePath(manifestPath, manifestInfo);
+        var shaFilePath = $"{sbomFilePath}.sha256";
+        var catFilePath = fileSystemUtils.JoinPaths(sbomDirPath, Constants.CatalogFileName);
+        var bsiFilePath = fileSystemUtils.JoinPaths(sbomDirPath, Constants.BsiFileName);
+        if (!fileSystemUtils.FileExists(shaFilePath) && !fileSystemUtils.FileExists(catFilePath) && !fileSystemUtils.FileExists(bsiFilePath))
+        {
+            // This is likely a CloudBuild SBOM, adjust paths accordingly
+            shaFilePath = null;
+            catFilePath = fileSystemUtils.JoinPaths(manifestPath, Constants.CatalogFileName);
+            bsiFilePath = fileSystemUtils.JoinPaths(manifestPath, Constants.BsiFileName);
+        }
+
         return Get(manifestInfo,
             manifestPath,
             sbomFilePath,
-            $"{sbomFilePath}.sha256",
-            fileSystemUtils.JoinPaths(sbomDirPath, Constants.CatalogFileName),
-            fileSystemUtils.JoinPaths(sbomDirPath, Constants.BsiFileName),
+            shaFilePath,
+            catFilePath,
+            bsiFilePath,
             new SbomPackageDetailsRecorder(),
             metadataBuilderFactory.Get(manifestInfo));
     }
 
-    public string GetSbomDirPath(string manifestDirPath, ManifestInfo manifestInfo) => fileSystemUtils.JoinPaths(
+    public string GetSpdxDirPath(string manifestDirPath, ManifestInfo manifestInfo) => fileSystemUtils.JoinPaths(
         manifestDirPath,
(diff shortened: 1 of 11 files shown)


src/Microsoft.Sbom.Api/Workflows/SbomConsolidationWorkflow.cs
```diff
@@ -3,41 +3,81 @@
...
+    }
+
+    private IEnumerable<(ISbomConfig config, ArtifactInfo info)> GetSbomsToValidate(string artifactPath, ArtifactInfo info)
+    {
+        var manifestDirPath = info?.ExternalManifestDir ?? fileSystemUtils.JoinPaths(artifactPath, Constants.ManifestFolder);
+        var isValidSpdxFormat = sPDXFormatDetector.TryGetSbomsWithVersion(manifestDirPath, out var detectedSboms);
+        if (!isValidSpdxFormat)
+        {
+            logger.Information($"No SBOMs located in {manifestDirPath} of a recognized SPDX format.");
+            return null;
```
DaveTryon, line 68: Given the scenarios in mind for consolidation, we might want to treat the "no SBOM was found" scenario the same as the "an invalid SBOM was found". The logging would be different, but the workflow would fail in both cases.

src/Microsoft.Sbom.Api/Workflows/SbomConsolidationWorkflow.cs
```diff
@@ -3,41 +3,81 @@
...
+        return await ValidateSourceSbomsAsync() && GeneratedConsolidatedSbom();
     }
-#pragma warning restore CS1998 // Placeholder, will use async in the future.
 
-    private bool ValidateSourceSboms()
+    private async Task<bool> ValidateSourceSbomsAsync()
     {
-        // TODO : Implement the source SBOMs.
-        return true;
+        var sbomsToValidate = ArtifactInfoMap.Select(artifact => GetSbomsToValidate(artifact.Key, artifact.Value))
```
DaveTryon: The generation step will need to use the data that is stored in `sbomsToValidate`. Ultimately, that code will need the path to the SBOM file, which we've already computed once. Is it worth capturing the data when we have it, rather than computing it multiple times?

src/Microsoft.Sbom.Api/Utils/SPDXFormatDetector.cs
```diff
@@ -17,17 +17,38 @@ public class SPDXFormatDetector : ISPDXFormatDetector
...
+
+    public bool TryGetSbomsWithVersion(string manifestDirPath, out IDictionary<string, ManifestInfo> detectedSboms)
+    {
+        detectedSboms = new Dictionary<string, ManifestInfo>();
+        foreach (var mi in supportedManifestInfos.Keys) {
+            var filePath = sbomConfigFactory.GetSbomFilePath(manifestDirPath, mi);
+            if (fileSystemUtils.FileExists(filePath))
+            {
+                var result = TryDetectFormat(filePath, out var detectedManifestInfo);
+                if (result && mi.Equals(detectedManifestInfo))
```
sfoslund, line 44: @DaveTryon do you have any opinion on the behavior in the (very edge) case where a SBOM with spdx2.2 in the path is actually a 3.0 SBOM (or vice versa)? This now this logic would just ignore it. Do you even think it's worthwhile to attempt to parse a SBOM with 2.2 in the path as 3.0 and vice versa?

src/Microsoft.Sbom.Api/Manifest/Configuration/SbomConfigFactory.cs
```diff
@@ -39,4 +41,29 @@ public ISbomConfig Get(
...
+        ManifestInfo manifestInfo,
+        string manifestPath,
+        IMetadataBuilderFactory metadataBuilderFactory)
+    {
+        var sbomDirPath = GetSbomDirPath(manifestPath, manifestInfo);
+        var sbomFilePath = GetSbomFilePath(manifestPath, manifestInfo);
+        return Get(manifestInfo,
+            manifestPath,
+            sbomFilePath,
+            $"{sbomFilePath}.sha256",
```
DaveTryon: We will need to be able to ingest manifests produced by CloudBuild, which use a slightly different file structure (and don't have a `.sha256` file). It would be ideal to include test cases that mimic the CloudBuild structure to be sure that we're covering both cases

src/Microsoft.Sbom.Api/Workflows/SbomConsolidationWorkflow.cs
sfoslund: Definitely, will refactor this so that the generation part can share the paths

sfoslund: Updated in latest push

src/Microsoft.Sbom.Api/Workflows/SbomConsolidationWorkflow.cs
sfoslund: With the current logic we log "No valid SBOMs detected." if there are no SBOMs found and return false in the validation step, causing the overall workflow to return false, so I think this case is already covered unless you prefer different behavior here.

src/Microsoft.Sbom.Api/Manifest/Configuration/SbomConfigFactory.cs
sfoslund: Added in latest push

src/Microsoft.Sbom.Api/Workflows/SbomConsolidationWorkflow.cs
```diff
@@ -3,41 +3,76 @@
...
+            return null;
+        }
+
+        return detectedSboms.Select((sbom) => (sbomConfigFactory.Get(sbom.manifestInfo, manifestDirPath, metadataBuilderFactory), info));
+    }
+
+    private async Task<bool> ValidateSourceSbomsAsync(IEnumerable<(ISbomConfig config, ArtifactInfo info)> sbomsToValidate)
+    {
+        var validationWorkflows = sbomsToValidate
+            .Select(async sbom => await Task.FromResult(true)); // TODO: Run validation workflow
```
DaveTryon, line 77: I love the idea of running the validations concurrently. Two thoughts (probably not for this PR, though):
1. Would it make sense to move the concurrency on level up, so that we're also running the detectors (and trying to parse the files) in parallel?
2. Logging from parallel activities can be difficult to decipher, and some of the telemetry that we capture may need to be broken out by SPDX format. That might impact the design for concurrency.

Ultimately, I wonder if V1 might be better without concurrency, then we can gather data to determine if the performance win of making it concurrent justifies the extra complexith

sfoslund: Great points, I'll consider this in my next PRs where I actually implement validation. I suspect I'll end up going with your suggestion to do the simple thing, and not run concurrently at least for v1

test/Microsoft.Sbom.Api.Tests/Utils/SPDXFormatDetectorTests.cs
```diff
@@ -36,19 +39,121 @@ public void SetUp()
...
+    [DataRow(Spdx22ContentStub, Spdx22VersionStub)]
+    [DataRow(Spdx30ContentStub, Spdx30VersionStub)]
+    public void TryGetSbomsWithVersion_SingleResult_Success(string testContent, string expectedVersion)
+    {
+        var spdx22FilePathStub = FilePathStub + Spdx22VersionStub;
+        var spdx30FilePathStub = FilePathStub + Spdx30VersionStub;
+        mockSbomConfigFactory
+            .Setup(m => m.GetSbomFilePath(DirPathStub, Api.Utils.Constants.SPDX22ManifestInfo))
+            .Returns(spdx22FilePathStub)
+            .Verifiable();
```
DaveTryon: Is `.Verifiable()` adding value here? I've typically seen it used with something like `.Verifiable(Times.Once())` or similar, but only in cases where we actually care about it not being called more than once. Since the test calls the `.VerifyAll` method, it's already getting checked.

test/Microsoft.Sbom.Api.Tests/Utils/SPDXFormatDetectorTests.cs
```diff
@@ -36,19 +39,121 @@ public void SetUp()
...
+        mockFileSystemUtils
+            .Setup(m => m.OpenRead(FilePathStub + Spdx30VersionStub))
+            .Returns(TestUtils.GenerateStreamFromString(Spdx30ContentStub))
+            .Verifiable();
+
+        var result = testSubject.TryGetSbomsWithVersion(DirPathStub, out var detectedSboms);
+        Assert.IsTrue(result);
+        Assert.IsNotNull(detectedSboms);
+        Assert.AreEqual(2, detectedSboms.Count());
+        Assert.IsTrue(detectedSboms.Any(value => value.manifestInfo.ToString().Equals(Spdx22VersionStub) && value.sbomFilePath.Equals(FilePathStub + Spdx22VersionStub)));
```
DaveTryon: If this fails, the unit test will just say something like "expected true, got false", which isn't great for debugging. I try to lean toward 'AreEqual`, or using the comment field to clarify the test failure. Something like
```
Assert.IsTrue(detectedSboms.Any(...), "SPDX 2.2. files should have been detected");
```

test/Microsoft.Sbom.Api.Tests/Utils/SPDXFormatDetectorTests.cs
sfoslund: Nope, removed!

src/Microsoft.Sbom.Api/Utils/SPDXFormatDetector.cs
sfoslund: Discussed offline and we will mention this to John Wade to get his input on this behavior

DaveTryon (approved):

DaveTryon (approved):

PR #1117, opened 2025-06-27
Add plumbing to collect packages from SPDX 2.2 files

DaveTryon:
This PR is all about adding the `IMergeableContentProvider interface, but includes some housekeeping. Specifically, it restructures the `./samples` folder to include both SPDX 2.2 and SPDX 3.0 files. These files were built against the win64 4.0.3 release, using the win 64 tool from the same release. These samples are available for tests

It also adds a functional SPDX 2.2 implementation of `IMergeableContentProvider` (enabled via dependency injection) and a placeholder SPDX 3.0 implementation of `IMergeableContentProvider` (not yet in dependency injection).

The `SbomConsolidationWorkflow.SourceSbomsTemp` property is a test stub that will be removed in the near future.

This code does not yet include plumbing to pass the SBOMs discovered during validation, or plumbing to capture the `MergeableContent` in the generation workflow.


src/Microsoft.Sbom.Extensions/IMergeableContentProviderInternal.cs
```diff
@@ -0,0 +1,21 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.IO;
+
+namespace Microsoft.Sbom.Extensions;
+
+/// <summary>
+/// Implementations of this interface provide mergeable SBOM content based on a given input stream.
+/// </summary>
+public interface IMergeableContentProviderInternal : IMergeableContentProvider
```
sfoslund: What's the motivation for making this a separate interface instead of having both of these methods in the IMergableContentProvider interface?

src/Microsoft.Sbom.Parsers.Spdx22SbomParser/MergeableContentProvider.cs
```diff
@@ -0,0 +1,125 @@
...
+{
+    /// <summary>
+    /// This provider supports only SPDX 2.2 files.
+    /// </summary>
+    public ManifestInfo ManifestInfo => Constants.Spdx22ManifestInfo;
+
+    /// <summary>
+    /// Implements <see cref="IMergeableContentProvider.TryGetContent(string, out MergeableContent)"/>.
+    /// </summary>
+    public bool TryGetContent(string filePath, out MergeableContent mergeableContent)
```
sfoslund, line 38: nit: this boilerplate appears to be the same in both the 22 and 30 versions of this class. Would it be worthwhile to create a base class they can both inherit from so as to share this code?

src/Microsoft.Sbom.Parsers.Spdx22SbomParser/MergeableContentProvider.cs
```diff
@@ -0,0 +1,125 @@
...
+
+    /// <summary>
+    /// Core engine for converting the stream to the content.
+    /// </summary>
+    private bool GetMergeableContent(Stream stream, out MergeableContent mergeableContent)
+    {
+        // Skip sections that are expensive to parse and would require additional logic to properly ignore.
+        var parser = new SPDXParser(stream, ["files", "externalDocumentRefs", "relationships"]);
+        var packages = Enumerable.Empty<SbomPackage>();
+        var relationships = Enumerable.Empty<SbomRelationship>();
```
sfoslund, line 78: I assume the relationships logic will come in a separate PR?

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/MergeableContentProvider.cs
```diff
@@ -0,0 +1,122 @@
...
+
+        ParserStateResult result = null;
+        do
+        {
+            result = parser.Next();
+            if (result is not null)
+            {
+                switch (result)
+                {
+                    //case PackagesResult packagesResult:
```
sfoslund: Accidentally commented? Can we delete or uncomment this?

src/Microsoft.Sbom.Parsers.Spdx22SbomParser/MergeableContentProvider.cs
```diff
@@ -0,0 +1,125 @@
...
+    /// Implements <see cref="IMergeableContentProvider.TryGetContent(string, out MergeableContent)"/>.
+    /// </summary>
+    public bool TryGetContent(string filePath, out MergeableContent mergeableContent)
+    {
+        if (filePath is null)
+        {
+            throw new ArgumentNullException(nameof(filePath));
+        }
+
+        if (!File.Exists(filePath))
```
sfoslund: Use file system utils to make this more testable?

test/Microsoft.Sbom.Parsers.Spdx22SbomParser.Tests/MergeableContentProviderTests.cs
```diff
@@ -0,0 +1,69 @@
...
+        Assert.IsFalse(result);
+        Assert.IsNull(mergeableContent);
+    }
+
+    [TestMethod]
+    public void TryGetContent_FilePath_FileIsValid_ReturnsExpectedContent()
+    {
+        var filePath = Path.GetFullPath(Path.Combine(
+            Assembly.GetExecutingAssembly().Location, "..", "..", "..", "..", "..", "..", "samples", "spdx_2.2", "manifest.spdx.json"));
+        var result = provider.TryGetContent(filePath, out var mergeableContent);
```
sfoslund: Would love to be able to avoid touching the file system in unit tests, could we use file system utils to mock and reuse the test spdx content in the Parser/jsonstrings dir? Same goes for the 30 tests

src/Microsoft.Sbom.Parsers.Spdx22SbomParser/MergeableContentProvider.cs
DaveTryon: Yes. Relationships are in a separate feature

test/Microsoft.Sbom.Parsers.Spdx22SbomParser.Tests/MergeableContentProviderTests.cs
DaveTryon: There are opposing forces here. While unit tests are ideally independent of the file system, it's also ideal to test with meaningful data. The mergeable content is complex--and will become even more complex when relationships are involved-- so I didn't want to put it into a string. Perhaps a reasonable way to use real data but still avoid the file system would be to embed the on-disk file into the test assembly as a resource, then have the mock return its contents.

src/Microsoft.Sbom.Parsers.Spdx22SbomParser/MergeableContentProvider.cs
DaveTryon: Sadly, they're in separate assemblies. It doesn't seem to fit any of the existing assemblies, and it seems like overkill to add a new assembly just for this, so I think this one falls into the category of "it would be nice..."

src/Microsoft.Sbom.Extensions/IMergeableContentProviderInternal.cs
DaveTryon: There were no scenarios for production code to operate on the stream, so it was for testability. Injecting the `IFileSystemUtils` into the object removed the need, so the `Internal` class has been removed

src/Microsoft.Sbom.Parsers.Spdx22SbomParser/MergeableContentProvider.cs
DaveTryon: Done

test/Microsoft.Sbom.Parsers.Spdx22SbomParser.Tests/MergeableContentProviderTests.cs
DaveTryon: I was able to move this into a file resource, which (when combined with the injected `IFileSystemUtils`), made the final code a lot cleaner

src/Microsoft.Sbom.Parsers.Spdx30SbomParser/MergeableContentProvider.cs
DaveTryon: Removed the file for this PR

src/Microsoft.Sbom.Api/Workflows/SbomConsolidationWorkflow.cs
```diff
@@ -23,12 +25,21 @@ public class SbomConsolidationWorkflow : IWorkflow<SbomConsolidationWorkflow>
...
     private readonly IMetadataBuilderFactory metadataBuilderFactory;
     private readonly IWorkflow<SbomGenerationWorkflow> sbomGenerationWorkflow;
+    private readonly IReadOnlyDictionary<ManifestInfo, IMergeableContentProvider> contentProviders;
 
-#pragma warning disable IDE0051 // We'll use this soon.
     private IReadOnlyDictionary<string, ArtifactInfo> ArtifactInfoMap => configuration.ArtifactInfoMap.Value;
-#pragma warning restore IDE0051 // We'll use this soon.
 
-    public SbomConsolidationWorkflow(ILogger logger, IConfiguration configuration, IWorkflow<SbomGenerationWorkflow> sbomGenerationWorkflow, ISbomConfigFactory sbomConfigFactory, ISPDXFormatDetector sPDXFormatDetector, IFileSystemUtils fileSystemUtils, IMetadataBuilderFactory metadataBuilderFactory)
+    internal IEnumerable<(ManifestInfo, string)> SourceSbomsTemp { get; set; } = Enumerable.Empty<(ManifestInfo, string)>(); // Stub property for testing, will remove soon
```
sfoslund, line 32: Can reuse sbomsToConsolidate now instead? Also fine if you want to wait and make this change in a future PR though

sfoslund (approved):

src/Microsoft.Sbom.Api/Workflows/SbomConsolidationWorkflow.cs
DaveTryon: Yep, that'll be the next PR

PR #1118, opened 2025-06-27
Running validation workflow in consolidate

sfoslund:
This PR adds logic to run the validation workflow for each SBOM provided in the consolidate command. Some things to note about this implementation:
- It updates telemetry to differentiate between validation failures for different SBOMs using an identifier. For example: 
```
...  Errors=ErrorContainer`1 {Count=8, Errors=[FileValidationResult {Path="./extra/file", ErrorType=AdditionalFile, Identifier="334415413"}, FileValidationResult {Path="./extra/file", ErrorType=AdditionalFile, Identifier="548817155"}]}, ....
```
- It writes all validation results both to output files in the temp dir and to stdout


src/Microsoft.Sbom.Api/Output/Telemetry/TelemetryRecorder.cs
```diff
@@ -145,7 +145,7 @@ public TimingRecorder TraceEvent(string eventName)
     /// <exception cref="ArgumentNullException">If the errors object is null.</exception>
     public void RecordTotalErrors(IList<FileValidationResult> errors)
     {
-        this.errors = errors ?? throw new ArgumentNullException(nameof(errors));
+        this.errors = this.errors.Concat(errors ?? throw new ArgumentNullException(nameof(errors))).ToList();
```
DaveTryon, line 148: This is changing the behavior, but I can't find any callers that call it more than once, so it should be ok

src/Microsoft.Sbom.Api/Workflows/SbomConsolidationWorkflow.cs
```diff
@@ -92,15 +95,38 @@ private IEnumerable<ConsolidationSource> GetSbomsToConsolidate(string artifactPa
...
-        return results.All(b => b);
+        var result = true;
+        foreach (var source in consolidationSources)
+        {
+            var identifier = Math.Abs(string.GetHashCode(source.SbomConfig.ManifestJsonFilePath)).ToString();
+            var defaultOutputPath = configuration.OutputPath.Value;
+            var workflowResult = false;
+            try
+            {
+                configuration.ValidateSignature.Value = !source.ArtifactInfo.SkipSigningCheck ?? true;
```
DaveTryon: Not urgent, but the `SkipSigningCheck` should probably be treated as a break-glass setting and emit a warning message that tells us which artifact we're not checking

DaveTryon: What does the `!` operator do if the input is `null` ?

src/Microsoft.Sbom.Api/Workflows/SbomConsolidationWorkflow.cs
```diff
@@ -92,15 +95,38 @@ private IEnumerable<ConsolidationSource> GetSbomsToConsolidate(string artifactPa
...
+                configuration.IgnoreMissing.Value = source.ArtifactInfo.IgnoreMissingFiles ?? false;
+                configuration.BuildDropPath.Value = source.BuildDropPath;
+                configuration.OutputPath.Value = fileSystemUtils.GetTempFile($"validation-results-{identifier}.json");
```
DaveTryon: For debugging, would be useful to create a temp folder, then put the files into that folder? That seems like it might be simpler to look through a few files than to try to filter from many. It also might make it a bit easier if we wanted to clean up the files after execution completes

src/Microsoft.Sbom.Api/Workflows/SbomConsolidationWorkflow.cs
```diff
@@ -92,15 +95,38 @@ private IEnumerable<ConsolidationSource> GetSbomsToConsolidate(string artifactPa
...
+
+                Console.WriteLine($"Running validation for {source.SbomConfig.ManifestJsonFilePath} with identifier {identifier}. Writing output results to {configuration.OutputPath}.");
```
DaveTryon: Will this eventually go through the logger?

src/Microsoft.Sbom.Api/Workflows/SbomParserBasedValidationWorkflow.cs
```diff
@@ -177,7 +216,7 @@ public async Task<bool> RunAsync()
 
                 await outputWriter.WriteAsync(JsonSerializer.Serialize(validationResultOutput, options));
 
-                validFailures = fileValidationFailures.Where(f => !Constants.SkipFailureReportingForErrors.Contains(f.ErrorType));
+                validFailures = fileValidationFailures.Where(f => !Constants.SkipFailureReportingForErrors.Contains(f.ErrorType)).ForEach(f => f.Identifier = eventIdentifier);
```
DaveTryon, line 219: Just confirming--rather than have multiple sets of errors, we're putting all of the errors into a single list then filtering to just the ones that are associated with this iteration?

src/Microsoft.Sbom.Api/Workflows/ISbomValidationWorkflowFactory.cs
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Sbom.Api.Workflows;
```
DaveTryon: Existing code puts the namespace after the using blocks

src/Microsoft.Sbom.Api/Workflows/SbomConsolidationWorkflow.cs
DaveTryon: The generation code will need to add a wrapper of [Directory.CreateTempSubDirectory](https://learn.microsoft.com/en-us/dotnet/api/system.io.directory.createtempsubdirectory?view=net-8.0). Maybe there's some synergy here, maybe not

src/Microsoft.Sbom.Api/Workflows/SbomParserBasedValidationWorkflow.cs
sfoslund: The existing code in this line filters out certain error types that we don't report to the user, and the new addition here adds the 'identifier` property on all of the valid failure events so that when we report these events in telemetry we know which validation run they're associated with

src/Microsoft.Sbom.Api/Workflows/SbomConsolidationWorkflow.cs
sfoslund: I'm unsure- the validation workflow does [pretty much all of its printing via the console, not the logger](https://github.com/microsoft/sbom-tool/blob/main/src/Microsoft.Sbom.Api/Workflows/SbomParserBasedValidationWorkflow.cs#L234), which is obviously not ideal. That said, I'm following that pattern here because IMO by default this line should have the same verbosity as all of the validation results that get printed to the console, and I don't think we can achieve that with the logger since by default only warnings and errors written to the logger are displayed. Feels like we could do a separate feature to clean up all of the logging vs console stuff throughout this repo, but I don't want to do a overhaul of that right now.

src/Microsoft.Sbom.Api/Workflows/SbomConsolidationWorkflow.cs
sfoslund: I've validated that this resolves to true when the SkipSigningCheck input is empty/null

sfoslund: Added in this PR 👍

src/Microsoft.Sbom.Api/Workflows/Helpers/ConsolidationSource.cs
```diff
@@ -16,10 +16,13 @@ internal class ConsolidationSource
...
     public ISbomConfig SbomConfig { get; }
 
-    public ConsolidationSource(ArtifactInfo artifactInfo, ISbomConfig sbomConfig)
+    public string BuildDropPath { get; }
+
+    public ConsolidationSource(ArtifactInfo artifactInfo, ISbomConfig sbomConfig, string buildDropPath)
     {
         ArtifactInfo = artifactInfo ?? throw new ArgumentNullException(nameof(artifactInfo));
         SbomConfig = sbomConfig ?? throw new ArgumentNullException(nameof(sbomConfig));
+        BuildDropPath = buildDropPath ?? throw new ArgumentNullException(nameof(buildDropPath));
```
DaveTryon: Should we have a check against this being an empty string?

src/Microsoft.Sbom.Api/Workflows/SbomConsolidationWorkflow.cs
```diff
@@ -92,15 +95,44 @@ private IEnumerable<ConsolidationSource> GetSbomsToConsolidate(string artifactPa
...
+            var defaultOutputPath = configuration.OutputPath.Value;
+            var workflowResult = false;
+            try
+            {
+                configuration.ValidateSignature = new ConfigurationSetting<bool>(!source.ArtifactInfo.SkipSigningCheck ?? true);
+                configuration.IgnoreMissing = new ConfigurationSetting<bool>(source.ArtifactInfo.IgnoreMissingFiles ?? false);
+                configuration.BuildDropPath = new ConfigurationSetting<string>(source.BuildDropPath);
+                configuration.OutputPath = new ConfigurationSetting<string>(fileSystemUtils.JoinPaths(validationOutputDir, $"validation-results-{identifier}.json"));
+
+                Console.WriteLine($"Running validation for {source.SbomConfig.ManifestJsonFilePath} with identifier {identifier}. Writing output results to {configuration.OutputPath.Value}.");
```
DaveTryon, line 117: Should this be `logger.Information` instead of `Console.WriteLine` ?

src/Microsoft.Sbom.Api/Workflows/SbomConsolidationWorkflow.cs
```diff
@@ -92,15 +95,44 @@ private IEnumerable<ConsolidationSource> GetSbomsToConsolidate(string artifactPa
...
+                if (!configuration.ValidateSignature.Value)
+                {
+                    logger.Warning($"Signature validation disabled for SBOM with path {source.SbomConfig.ManifestJsonFilePath} and identifier {identifier}.");
+                }
+
+                var workflow = sbomValidationWorkflowFactory.Get(configuration, source.SbomConfig, identifier);
+                workflowResult = await workflow.RunAsync();
+            }
+            finally
```
DaveTryon, line 126: Should we have a catch block to set `workflowResult=false` if an exception gets thrown?

src/Microsoft.Sbom.Api/Workflows/SbomConsolidationWorkflow.cs
sfoslund: Discussed this issue here: https://github.com/microsoft/sbom-tool/pull/1118#discussion_r2177868698

src/Microsoft.Sbom.Api/Workflows/SbomConsolidationWorkflow.cs
sfoslund: I don't think that's necessary- if there's an error thrown anywhere in the try block then workflowResult will still be its default value (false) since we only set it to true at the very end of the try block if RunAsync succeeds

src/Microsoft.Sbom.Api/Workflows/SbomConsolidationWorkflow.cs
DaveTryon: It looks like this pattern was set in #274, although no reason or explanation is given in the PR. Our default log level is `Information`, so I would think that the most correct answer is to use `logger.Information`. Other workflows seem to reserve `Console.WriteLine` to specific error cases.

All that said, this feels like a separate conversation/PR about how we want to display the output holistically

src/Microsoft.Sbom.Api/Workflows/SbomConsolidationWorkflow.cs
DaveTryon: As discussed in another place in the PR, withdrawing this comment

src/Microsoft.Sbom.Api/Workflows/SbomConsolidationWorkflow.cs
DaveTryon: Ah--missed that!

PR #1119, opened 2025-06-28
Pass set of validated SBOMs to consolidation

DaveTryon:
#1117 included a temporary property that was just for testing purposes. This PR is about removing that temporary property.

1. Remove `SbomConsolidationWorkflow.SourceSbomsTemp`
2. The SBOM detection code had the sbomPath, but wasn't passing it back to where the consolidation code could access it. I could have added another field to the `Tuple`, but it felt cleaner to replace the `Tuple` with a new class--this is the `ConsolidationSource` class
3. Add a couple of mocks to replace the test that was setting the temporary property
4. Rename `sbomsToValidate` to `consolidationSources`, since the same data will flow to both validation and consolidation
5. Use some const strings in the unit tests
6. Casing consistency: We had a couple of variables whose names began with `sPDXFormatDetector`. Change them to begin with `spdxFormatDetector`

### DIFF
--- a/src/Microsoft.Sbom.Api/Workflows/Helpers/ConsolidationSource.cs
+++ b/src/Microsoft.Sbom.Api/Workflows/Helpers/ConsolidationSource.cs
@@ -24,4 +24,9 @@ internal class ConsolidationSource
         SbomConfig = sbomConfig ?? throw new ArgumentNullException(nameof(sbomConfig));
         SbomPath = sbomPath ?? throw new ArgumentNullException(nameof(sbomPath));
     }
+
+    public override string ToString()
+    {
+        return $"ConsolidationSource: {ArtifactInfo}, ManifestInfo: {SbomConfig.ManifestInfo}, SbomPath: {SbomPath}";
+    }
 }
--- a/src/Microsoft.Sbom.Api/Workflows/Helpers/ConsolidationSource.cs
+++ b/src/Microsoft.Sbom.Api/Workflows/Helpers/ConsolidationSource.cs
@@ -1,0 +1,27 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using Microsoft.Sbom.Common.Config;
+using Microsoft.Sbom.Extensions;
+
+namespace Microsoft.Sbom.Api.Workflows.Helpers;
+
+/// <summary>
+/// A class that lets us track information about the source artifact that feeds into a consolidated SBOM.
+/// </summary>
+internal class ConsolidationSource
+{
+    public ArtifactInfo ArtifactInfo { get; }
+
+    public ISbomConfig SbomConfig { get; }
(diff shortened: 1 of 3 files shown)


src/Microsoft.Sbom.Api/Workflows/SbomConsolidationWorkflow.cs
```diff
@@ -67,46 +66,46 @@ public SbomConsolidationWorkflow(
...
-        var isValidSpdxFormat = sPDXFormatDetector.TryGetSbomsWithVersion(manifestDirPath, out var detectedSboms);
+        var isValidSpdxFormat = spdxFormatDetector.TryGetSbomsWithVersion(manifestDirPath, out var detectedSboms);
         if (!isValidSpdxFormat)
         {
             logger.Information($"No SBOMs located in {manifestDirPath} of a recognized SPDX format.");
             return null;
         }
 
-        return detectedSboms.Select((sbom) => (sbomConfigFactory.Get(sbom.manifestInfo, manifestDirPath, metadataBuilderFactory), info));
+        return detectedSboms.Select((sbom) => new ConsolidationSource(info, sbomConfigFactory.Get(sbom.manifestInfo, manifestDirPath, metadataBuilderFactory), sbom.sbomFilePath));
```
sfoslund: Why do we need to save `sbom.sbomFilePath` separately from the `SbomConfig`? `SbomConfig.ManifestJsonFilePath` should equal `sbom.sbomFilePath`

DaveTryon: Nice catch!

DaveTryon:
@sfoslund, we could build an argument for using the `ConsolidationSource` in `ISPDXFormatDetector.TryGetSbomsWithVersion`, but I've left that out of this PR. Happy to change it in a separate PR if you think that's useful

src/Microsoft.Sbom.Api/Workflows/Helpers/ConsolidationSource.cs
```diff
@@ -0,0 +1,29 @@
...
+
+    public ISbomConfig SbomConfig { get; }
+
+    public ConsolidationSource(ArtifactInfo artifactInfo, ISbomConfig sbomConfig)
+    {
+        ArtifactInfo = artifactInfo ?? throw new ArgumentNullException(nameof(artifactInfo));
+        SbomConfig = sbomConfig ?? throw new ArgumentNullException(nameof(sbomConfig));
+    }
+
+    public override string ToString()
```
sfoslund, line 25: nit: why is this needed?

DaveTryon: This is just a debugging aid, has no direct impact on the output

DaveTryon: I could make it debug-only if that's useful

PR #1124, opened 2025-07-01
Do not require outputPath in consolidate config file

sfoslund:
Fixes a bug in the consolidation code in which we expected the `OutputPath` param to always be specified in the consolidate config file.

### DIFF
--- a/src/Microsoft.Sbom.Api/Workflows/SbomConsolidationWorkflow.cs
+++ b/src/Microsoft.Sbom.Api/Workflows/SbomConsolidationWorkflow.cs
@@ -105,7 +105,6 @@ public class SbomConsolidationWorkflow : IWorkflow<SbomConsolidationWorkflow>
         foreach (var source in consolidationSources)
         {
             var identifier = Math.Abs(string.GetHashCode(source.SbomConfig.ManifestJsonFilePath)).ToString();
-            var defaultOutputPath = configuration.OutputPath.Value;
             var workflowResult = false;
             try
             {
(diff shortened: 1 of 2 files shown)


src/Microsoft.Sbom.Api/Workflows/SbomConsolidationWorkflow.cs
```diff
@@ -126,7 +125,7 @@ private async Task<bool> ValidateSourceSbomsAsync(IEnumerable<ConsolidationSourc
             finally
             {
                 configuration.BuildDropPath.Value = null;
-                configuration.OutputPath.Value = defaultOutputPath;
+                configuration.OutputPath.Value = null;
```
DaveTryon, line 136: If we never set it, do we need to clear it?

sfoslund: We have to set it per-validation workflow run a few lines up from here, in order to make sure the validation outputs are being written to the right temp dir

test/Microsoft.Sbom.Api.Tests/Workflows/SbomConsolidationWorkflowTests.cs
```diff
@@ -148,7 +148,7 @@ public async Task RunAsync_MixOfValidAndInvalidInputSboms()
     {
         SetUpSbomsToValidate();
 
-        configurationMock.Setup(m => m.OutputPath).Returns(new ConfigurationSetting<string>("test-out-path"));
+        configurationMock.Setup(m => m.OutputPath).Returns(new ConfigurationSetting<string>());
```
DaveTryon, line 189: Is the setup needed if we no longer access it?

sfoslund: It's still needed because we're setting outputPath to a temp dir for each validation run and then resetting it to null after

PR #1126, opened 2025-07-03
Generated a consolidated SBOM

DaveTryon:
This PR creates the full end-to-end experience with consolidation, using real SBOM as the basis for the local testing. It needs more unit tests and maybe some E2E tests, but this connects the pieces. Here are the key issues that it addresses:

1. The first issue was that we needed to get the correct set of packages, instead of the ones from the component-detection code. To do this, we set the `IConfiguration.PackagesList` property. This causes [SbomPackagesProvider](https://github.com/microsoft/sbom-tool/blob/main/src/Microsoft.Sbom.Api/Providers/PackagesProviders/SbomPackagesProvider.cs#L34-L46) to replace [CGScannedPackagesProvider](https://github.com/microsoft/sbom-tool/blob/main/src/Microsoft.Sbom.Api/Providers/PackagesProviders/CGScannedPackagesProvider.cs#L43-L56) as the source of the `packages` section of the generated SBOM.
2. The second issue was that the validation of packages with the _manifest folder contained in the drop were failing during consolidation. This was because the validation code uses the `ManifestDirPath` property to automatically ignore the contents of this folder, meaning that it needs to be set for each SBOM that we validate. We now set, then restore the property around each validation workflow. The code that sets the property is duplicated, but there was no obvious to avoid this, short of creating a new class, which seemed like overkill for the amount of code involved. It's easy to refactor if needed.
3. The third issue was that the consolidate SBOM was being written to the wrong location. Instead of going to the location specified by the `ManifestDirPath`, it was overwriting one of the input SBOMs. This took a while to figure out, but ultimately it boils down to the fact that the `SbomConfigProvider` was written with the assumption that `ManifestDirPath` was constant throughout the app's lifetime. The value set during validation was saved, then reused during generation. There are multiple ways we could address this, but I elected to add a `ISbomConfigProvider` that clears the cache between validation and generation. I also considered changing the `ISbomConfigProvider from a Singleton to a Transient, but I wanted to minimize the potential side effects. The new method only gets called by the consolidation workflow, so the side effects are constrained.

The rest of the changes are mainly about unit tests (including some refactoring of setting up the mocks) and adding an extension class for managing the `MergeableContent`. I also added code to clean up our temp directory after exiting. We can tweak that as needed.

### DIFF
--- a/src/Microsoft.Sbom.Api/Workflows/SbomConsolidationWorkflow.cs
+++ b/src/Microsoft.Sbom.Api/Workflows/SbomConsolidationWorkflow.cs
@@ -21,6 +21,8 @@ namespace Microsoft.Sbom.Api.Workflows;
 
 public class SbomConsolidationWorkflow : IWorkflow<SbomConsolidationWorkflow>
 {
+    public const string WorkingDirPrefix = "sbom-consolidation-";
+
     private readonly ILogger logger;
     private readonly IConfiguration configuration;
     private readonly ISbomConfigFactory sbomConfigFactory;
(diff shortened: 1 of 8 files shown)


src/Microsoft.Sbom.Api/Workflows/SbomConsolidationWorkflow.cs
```diff
@@ -90,7 +96,24 @@ public virtual async Task<bool> RunAsync()
...
+        catch (Exception)
+        {
+            // This is here to help debug issues during active development. It will be removed before release.
+            System.Diagnostics.Debugger.Break();
+            throw;
+        }
+#endif
+        finally
+        {
+            fileSystemUtils.DeleteDir(workingDir, true);
```
sfoslund: Are we sure we want to delete all of the validation output files? I'm of two minds about this- I can imagine scenarios where users may want these files around to dive into validation errors, but we obviously don't want to leave a mess behind. At a minimum, if we do decide to delete this dir, we should probably [remove the log messages that print the path](https://github.com/microsoft/sbom-tool/pull/1126/files#diff-ec5e66cabef4dcfd4b4e4183e932d33dcc18a4daef09b6ada571cf0c12912202R148), since that could get confusing to users.

sfoslund (approved):
One comment, otherwise LGTM!

src/Microsoft.Sbom.Api/Workflows/SbomConsolidationWorkflow.cs
DaveTryon: Removed the delete, but added a prefix to the working directory so it's easy to clean up manually

sfoslund (approved):

PR #1128, opened 2025-07-04
Add aggregation telemetry

DaveTryon:
This PR adds the package count and relationship count for each `AggregationSource` to the telemetry stream. Steps to this point:

1. Adds `Events.SbomAggregationWorkflow` so we can record this as a discrete event
2. Adds `AggregationSource.Identifier` and uses it for validation logs. We'll also use it for reporting package information in telemetry
3. Add a top-level exception handler to `SbomAggregationWorkflow.RunAsync` so that we can catch and log the exceptions
4. Adds a small refactor to the code that calls the various `GenerateAsync` methods, so that aggregation calls only the generators that need to be called. Aggregating relationships will occur in a later feature.
5. Add `IRecorder.RecordAggregationSource` and its implementation to capture the data
6. Call `IRecorder.RecordAggregationSource` from the workflow

When all the pieces land, we get the following added to the telemetry stream:
```
AggregationSourceTelemetry = {
  ["1832146932"]=AggregationSourceTelemetry {PackageCount=266, RelationshipCount=0},
  ["1810187031"]=AggregationSourceTelemetry {PackageCount=3, RelationshipCount=0}
}
```

I'd love to remove the "AggregationSourceTelemetry" prefix on each item, but so far I haven't found a way to make that happen.

### DIFF
--- a/src/Microsoft.Sbom.Api/Output/Telemetry/Entities/AggregationSourceTelemetry.cs
+++ b/src/Microsoft.Sbom.Api/Output/Telemetry/Entities/AggregationSourceTelemetry.cs
@@ -1,0 +1,14 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Sbom.Api.Output.Telemetry.Entities;
+
+/// <summary>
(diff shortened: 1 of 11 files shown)


DaveTryon:
Closing/reopening to jumpstart CLA check

sfoslund (approved):

PR #1135, opened 2025-07-11
Convert info message about invalid aggregation input to warn

sfoslund:
This can be an indication that a user's config file is not properly formatted, so we don't want it to get lost in the logs.


DaveTryon (approved):
Part of me wants to make this an error, but let's start with warning and see how that works out

PR #1136, opened 2025-07-14
Include empty files and relationships arrays in aggregated SBOMs

sfoslund:
The SBOM parser requires the files and relationships properties to be present. This PR adds those properties to aggregated SBOMs

### DIFF
--- a/src/Microsoft.Sbom.Api/Providers/FilesProviders/CGScannedExternalDocumentReferenceFileProvider.cs
+++ b/src/Microsoft.Sbom.Api/Providers/FilesProviders/CGScannedExternalDocumentReferenceFileProvider.cs
@@ -53,8 +53,11 @@ public class CGScannedExternalDocumentReferenceFileProvider : PathBasedFileToJso
     {
         if (providerType == ProviderType.Files)
         {
-            Log.Debug($"Using the {nameof(CGScannedExternalDocumentReferenceFileProvider)} provider for the files workflow.");
-            return true;
+            if (Configuration.FilesList?.Value == null && string.IsNullOrWhiteSpace(Configuration.BuildListFile?.Value))
+            {
+                Log.Debug($"Using the {nameof(CGScannedExternalDocumentReferenceFileProvider)} provider for the files workflow.");
+                return true;
+            }
         }
 
         return false;
--- a/src/Microsoft.Sbom.Api/Providers/FilesProviders/CGScannedExternalDocumentReferenceFileProvider.cs
+++ b/src/Microsoft.Sbom.Api/Providers/FilesProviders/CGScannedExternalDocumentReferenceFileProvider.cs
@@ -51,13 +51,10 @@ public class CGScannedExternalDocumentReferenceFileProvider : PathBasedFileToJso
 
     public override bool IsSupported(ProviderType providerType)
     {
-        if (providerType == ProviderType.Files)
+        if (providerType == ProviderType.Files && Configuration.ManifestToolAction != ManifestToolActions.Aggregate)
         {
-            if (Configuration.FilesList?.Value == null && string.IsNullOrWhiteSpace(Configuration.BuildListFile?.Value))
-            {
-                Log.Debug($"Using the {nameof(CGScannedExternalDocumentReferenceFileProvider)} provider for the files workflow.");
-                return true;
-            }
+            Log.Debug($"Using the {nameof(CGScannedExternalDocumentReferenceFileProvider)} provider for the files workflow.");
+            return true;
         }
 
         return false;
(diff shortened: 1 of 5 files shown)


src/Microsoft.Sbom.Api/Providers/FilesProviders/CGScannedExternalDocumentReferenceFileProvider.cs
```diff
@@ -53,8 +53,11 @@ public override bool IsSupported(ProviderType providerType)
     {
         if (providerType == ProviderType.Files)
         {
-            Log.Debug($"Using the {nameof(CGScannedExternalDocumentReferenceFileProvider)} provider for the files workflow.");
-            return true;
+            if (Configuration.FilesList?.Value == null && string.IsNullOrWhiteSpace(Configuration.BuildListFile?.Value))
```
DaveTryon: Would it make sense to change line 54 to something like this, and ship the check for the `FileList`?
```
if (providerType == ProviderType.File || Configuration.ManifestToolAction.Value == ManifestToolActions.Consolidate)
```

If we keep the check on line 56, the null propagation in the call to `IsNullOrWhiteSpace` is superfluous because the first guard will fail if `Configuration.BuildListFile` is null

src/Microsoft.Sbom.Api/Workflows/SbomGenerationWorkflow.cs
```diff
@@ -201,11 +198,8 @@ private async Task<IEnumerable<FileValidationResult>> CallGeneratorsAync(IEnumer
             validErrors.AddRange(externalDocumentReferenceGeneratorResult.Errors);
         }
 
-        if (configuration.ManifestToolAction == ManifestToolActions.Generate)
-        {
-            var relationshipGeneratorResult = await relationshipsArrayGenerator.GenerateAsync(targetConfigs, elementsSpdxIdList);
-            validErrors.AddRange(relationshipGeneratorResult.Errors);
-        }
+        var relationshipGeneratorResult = await relationshipsArrayGenerator.GenerateAsync(targetConfigs, elementsSpdxIdList);
```
DaveTryon, line 201: What's the impact of enabling the relationship generator in this PR? I'll be enabling this in a separate PR and just want to know what I need to watch for when both pieces come together.

src/Microsoft.Sbom.Api/Providers/FilesProviders/CGScannedExternalDocumentReferenceFileProvider.cs
sfoslund: Discussed offline- a comment in IConfiguration implies that the original intent was not to use any file providers if the FilesList config value is set, but if that is the case then this is a bug in the CG file provider that has been around for a long time. To avoid breaking users who now depend on the current behavior, we will leave it as is and update the comment.

src/Microsoft.Sbom.Api/Workflows/SbomGenerationWorkflow.cs
sfoslund: Discussed offline, this change just adds a relationships section with only one instance, the root package relationship. This change just ensures that aggregated SBOMs have all of the required SBOM properties to be parse-able by our parser.

src/Microsoft.Sbom.Api/Providers/FilesProviders/CGScannedExternalDocumentReferenceFileProvider.cs
```diff
@@ -51,7 +51,7 @@ public CGScannedExternalDocumentReferenceFileProvider(
 
     public override bool IsSupported(ProviderType providerType)
     {
-        if (providerType == ProviderType.Files)
+        if (providerType == ProviderType.Files && Configuration.ManifestToolAction != ManifestToolActions.Aggregate)
```
DaveTryon: Do we need this for validation or redaction? If not, maybe make it a positive?
```suggestion
        if (providerType == ProviderType.Files && Configuration.ManifestToolAction == ManifestToolActions.Generate)
```

src/Microsoft.Sbom.Api/Workflows/SbomAggregationWorkflow.cs
```diff
@@ -215,6 +215,7 @@ private void SetConfigurationForAggregation(IEnumerable<MergeableContent> mergea
         configuration.BuildDropPath = new ConfigurationSetting<string>(buildDropPath);
         configuration.BuildComponentPath = new ConfigurationSetting<string>(buildDropPath);
         configuration.PackagesList = new ConfigurationSetting<IEnumerable<SbomPackage>>(mergeableContents.ToMergedPackages());
+        configuration.FilesList = new ConfigurationSetting<IEnumerable<SbomFile>>(Enumerable.Empty<SbomFile>());
```
DaveTryon: Do we still need this? It will eventually interfere with DirectoryTraversingFileToJsonProvider

DaveTryon (approved):

PR #1142, opened 2025-07-17
Refactor constructor for Generator class

DaveTryon:
#1139 added an `IConfiguration` object to the constructor of the `Generator` class. We have an internal consumer of this class, and that consumer does not have a readily available `IConfiguration` object to pass to the constructor. Rather than disrupt that consumer, it is preferable to allow a default constructor for the `Generator` class. Ultimately, `Generator` only needs to know how to handle existing SPDX ID's, so that's the only state that it stores. That state can now be set in any of three ways:
1. The default constructor (will retain the behavior that existed before #1139) 
2. A parameterized constructor that accepts a `bool`
3. A parameterized constructor that accepts an `IConfiguration`

Unit tests cover all scenarios with all applicable constructors. Because each test needs to use a specific constructor, the unit tests no longer use a fixture for the test subject.

### DIFF
--- a/src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Generator.cs
+++ b/src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Generator.cs
@@ -41,12 +41,13 @@ public class Generator : IManifestGenerator
     public string ExternalDocumentRefArrayHeaderName => Constants.ExternalDocumentRefArrayHeaderName;
 
     // This constructor gets called by an internal consumer that does not use IConfiguration.
+    // In this repo, we use it in tests to pin the existing behavior via this constructor.
     public Generator()
     {
         actionIsAggregate = false;
     }
 
-    // This constructor gets used within this repo.
+    // This constructor gets used by the production code and by at least one test.
     public Generator(IConfiguration configuration)
     {
         ArgumentNullException.ThrowIfNull(configuration, nameof(configuration));
--- a/src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Generator.cs
+++ b/src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Generator.cs
@@ -26,7 +26,7 @@ namespace Microsoft.Sbom.Parsers.Spdx22SbomParser;
 /// </summary>
 public class Generator : IManifestGenerator
 {
-    private readonly bool actionIsAggregate;
+    private readonly bool keepAllExistingIds;
 
     public AlgorithmName[] RequiredHashAlgorithms => new[] { AlgorithmName.SHA256, AlgorithmName.SHA1 };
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Generator.cs
```diff
@@ -185,15 +195,13 @@ public GenerationResult GenerateJsonDocument(SbomPackage packageInfo)
 
     private bool ShouldWeKeepTheExistingId(string spdxId)
     {
-        switch (configuration.ManifestToolAction)
```
DaveTryon, line 188: Quick note: Prior to #1139, this code always checked for equality with `Constants.RootPackageIdValue`. In that sense, the code in this PR gets closer to the code before #1139.

src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Generator.cs
```diff
@@ -185,15 +195,8 @@ public GenerationResult GenerateJsonDocument(SbomPackage packageInfo)
...
-        {
-            case ManifestToolActions.Aggregate:
-                return true;
-            case ManifestToolActions.Generate:
-                return spdxId.Equals(Constants.RootPackageIdValue, StringComparison.OrdinalIgnoreCase);
-        }
-
-        return false;
+        return keepAllExistingIds ||
```
DaveTryon, line 204: Quick note: Prior to #1139, this code this method replaced _always_ checked for equality with `Constants.RootPackageIdValue`. After this PR, the pre- #1139 behavior is restored if `keepAllExistingIds` is false

src/Microsoft.Sbom.Parsers.Spdx22SbomParser/Generator.cs
```diff
@@ -40,9 +40,25 @@ public class Generator : IManifestGenerator
...
+    }
+
+    // This constructor gets used by the default constructor and by tests.
+    public Generator(bool keepAllExistingIds)
+    {
+        this.keepAllExistingIds = keepAllExistingIds;
+    }
+
+    // This constructor gets used by the production code and by at least one test.
     public Generator(IConfiguration configuration)
```
ZhengHong-Tan, line 57: do we use this constructor for both the generate and aggregate flows? if we do, i'm not quite understanding why we would need the other 2 constructors, as we could mock the configuration to be an aggregate action for testing.

DaveTryon: The 3rd constructor (accepting the IConfiguration) is called by Dependency Injection within this repo and is used by both generation and aggregation _within this repo_. The internal repo calls a constructor with no parameters, and the expected behavior (for now) is that the parameter will always be false. With that in mind, it would be possible to omit lines 46-51 and have just the 2 constructors (one with the IConfiguration, the other with no parameters).

Then I wondered if the internal repo would ever need to set it to a different value, and I couldn't anticipate that answer, so I added a constructor that took a bool with a default value of false. There are various schools of thought on default parameters, so I expanded it to 3 constructors--one with no parameters (and the flag is always false), one with a single bool parameter, and one with the IConfguration parameter.

Bottom line, we need 2 constructors (one with IConfiguration and one without), but we don't necessarily need 3 constructors. I'm adding it for completeness, rather than based on necessity.

PR #1146, opened 2025-07-23
Exclude samples folder from externaldocreferences

DaveTryon:
When we build the manifests for release, they are treating the SPDX files in the `./samples` folder as external document references, which is causing some downstream pain. To prevent this, we just delete the `./samples` folder after building and running our tests. I've run this in a [test build](https://dev.azure.com/mseng/AzureDevOps/_build/results?buildId=30266343&view=results) and confirmed that the manifests we produce are no longer including the samples up as external document references.


DaveTryon:
Closing since this works incompletely.

DaveTryon:
Restoring this since it's a step in the right direction, even if it might need a little more work to get completely right

pragnya17 (approved):

DaveTryon:
Closing and reopening to jumpstart CLA

PR #1161, opened 2025-08-16
Exit with appropriate exit code when providing version

GDWR:
This Pull Request updates the handling of the `--version` & `version` subcommand to exit immediately with a `0` (successful) exit code, in scenarios that the version is correctly output. `1` (failure) will still be emitted when the version cannot be retrieved correctly.

I've noticed this while outputting debug information in pipelines, as it causes the task to fail. I can always ignore the unsuccessful exit `|| true` etc, but I think outputting the correct exit code would be ideal.

---

I did have a look at utilising the "Args" return to carry the exit code back to `Program.cs` for handling the exiting of the application in unified manner, but it feels like a misuse of "Args" as they're intended to be client driven. If this approach is desired, from the structure of the application I would suggest moving the handling of version into a `VersionService` to unify how the handling of commands and subsequent exit codes are across the application. As the version logic is already different to the rest in where it is handled, I did not think this refactor would be desired. I am happy to refactor into a service if that feels more appropriate.

### DIFF
--- a/test/Microsoft.Sbom.Tool.Tests/IntegrationTests.cs
+++ b/test/Microsoft.Sbom.Tool.Tests/IntegrationTests.cs
@@ -870,4 +870,13 @@ public class IntegrationTests
             Assert.Fail($"The following packages are not linked to the root package: {unlinkedPackages}");
         }
     }
+
+    [DataRow("version")]
+    [DataRow("--version")]
+    [TestMethod]
+    public void E2E_Version_SuccessfulExitCode(string arguments)
+    {
+        var (_, _, exitCode) = LaunchAndCaptureOutput(arguments);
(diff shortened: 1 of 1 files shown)


GDWR:

GDWR:
> LGTM. It will probably need to be to the latest commit before it can be merged

I was going to rebase the PR, but it seems like its already based on the latest commit in `main`. 
Is that what you wanted me to do?

DaveTryon:
@GDWR Yep, my mistake. We had a commit on Friday and I mistakenly thought that it had merged after your base commit. Merging now. Thanks for the contribution!

Version 4.1.1 should release later this week. Your change will probably be in release 4.1.2 (no ETA at the moment)

PR #1237, opened 2025-09-26
Include only directly referenced packages

pragnya17:
Compliance requirements have led us to simplify the package dependency tree that is reported in SBOMs. There were issues in the original implementation of how transitive package dependencies were being represented in SBOMs, therefore we are making the default behavior to just include directly referenced packages. This means that the packages and relationships sections of SBOMs will be impacted, and will now only represent explicitly referenced packages.

In summary the goals are:
1. Only include explicitly referenced packages in the packages array (this is achieved by ProcessDependencyGraphs). 
2. Only include direct dependencies in the relationships list (this is achieved by the changes to the DependsOn property).

### DIFF
--- a/test/Microsoft.Sbom.Tool.Tests/IntegrationTests.cs
+++ b/test/Microsoft.Sbom.Tool.Tests/IntegrationTests.cs
@@ -195,7 +195,7 @@ public class IntegrationTests
         var jsonElement = ReadJsonFile(originalManifestFilePath);
         var packages = jsonElement.GetProperty("packages");
         var relationships = jsonElement.GetProperty("relationships");
-        Assert.AreEqual(packages.GetArrayLength(), relationships.GetArrayLength(), "The number of relationships should equal the number of packages + 1, indicating only direct package dependencies are recorded. 1 is added because there is a default DESCRIBES relationship.");
+        Assert.AreEqual(packages.GetArrayLength(), relationships.GetArrayLength(), "The number of relationships should equal the number of packages.");
     }
 
     [TestMethod]


src/Microsoft.Sbom.Api/Executors/ComponentDetectionBaseWalker.cs
```diff
@@ -113,7 +113,7 @@ async Task Scan(string path)
 
             var scanSettings = cliArgumentBuilder.BuildScanSettingsFromParsedArgs(cmdLineParams);
 
-            var scanResult = await componentDetector.ScanAsync(scanSettings);
+            var scanResult = await componentDetector.ScanAsync(scanSettings) as DefaultGraphScanResult;
 
             if (scanResult.ResultCode != ProcessingResultCode.Success)
```
KalleOlaviNiemitalo: Because this does bot tolerate `scanResult` being null, you should just cast rather than use `as`; so that if the type does not match, you get an InvalidCastException that shows which type it is, rather than NullReferenceException.

src/Microsoft.Sbom.Api/Executors/PackageInfoJsonWriter.cs
```diff
@@ -67,21 +67,22 @@ internal async Task GenerateJson(
                 var generationResult =
                     manifestGeneratorProvider.Get(sbomConfig.ManifestInfo).GenerateJsonDocument(packageInfo);
 
-                var recordedAnyDependencies = false;
+                //var recordedAnyDependencies = false;
```
sfoslund: I'm assuming we're commenting this out instead of deleting it because we imagine we might want to reenable some version of this behavior some day? If that's the case, can we add a comment above this section with a quick statement about that?

alisonlomaka: Could we rely on Git to keep track of the previous behavior for us instead of leaving the commented out code?

src/Microsoft.Sbom.Api/Executors/ComponentDetectionBaseWalker.cs
```diff
@@ -123,6 +123,22 @@ async Task Scan(string path)
...
+                var explicitComponentIds = new HashSet<string>();
+
+                foreach (var dependencyGraphPair in defaultGraphScanResult.DependencyGraphs)
+                {
+                    var dependencyGraph = dependencyGraphPair.Value;
+                    explicitComponentIds.UnionWith(dependencyGraph.ExplicitlyReferencedComponentIds);
+                }
+
+                uniqueComponents = uniqueComponents
+                .Where(component => explicitComponentIds.Contains(component.Component.Id));
```
alisonlomaka: Not sure how this will behave if the uniqueComponents are SbomComponents. Perhaps this logic should move inside PackagesWalker.FilterScannedComponents?

src/Microsoft.Sbom.Api/Executors/ComponentDetectionBaseWalker.cs
pragnya17: I've added a null check, so this should no longer be an issue

src/Microsoft.Sbom.Api/Executors/ComponentDetectionBaseWalker.cs
pragnya17: Since FilterScannedComponents is an abstract method, I'd have to duplicate the logic amongst PackagesWalker and SbomComponentsWalker, so I chose to add the code here instead, to avoid repetition. I did refactor into a separate method which is called after the FilterScannedComponents though, so now we are getting explicit components after filtering for SpdxComponents.

PR #1324, opened 2025-11-17
Fix release pipeline internal feed logic

sfoslund:
The packageParentPath needs to be a parent dir to where the nuget packages are stored- see [docs ](https://eng.ms/docs/coreai/devdiv/one-engineering-system-1es/1es-docs/1es-pipeline-templates/features/outputs/nuget-packages)


meoslund:
I approve. Please proceed.

PR #1493, opened 2026-04-24
Upgrade packageurl-dotnet to 2.0.0 and CD packages to 7.1.1

joshuamay-ms:
Upgrade packageurl-dotnet from 2.0.0-rc.3 to 2.0.0 to unblock nupkg creation


src/Microsoft.Sbom.Common/JsonAsynchronousNodeKit/LargeJsonParser.cs
```diff
@@ -149,7 +149,7 @@ static int GetStartPosition(Stream stream)
 
     private ParserStateResult HandleExplicitProperty(ref Utf8JsonReader reader, string propertyName)
     {
-        var handler = this.handlers![propertyName];
+        var handler = this.handlers[propertyName];
```
joshuamay-ms, line 152: This is fixing an explicit compiler break

samikay101 (approved):

Issue #944, opened 2025-02-21
License information may not be fully recorded


schuhbsi:
sbom-tool version: 3.1.0

I execute the following command in an Azure Devops pipeline:

sbom-tool generate -b $(Build.ArtifactStagingDirectory) -bc ${{ parameters.workingDirectory }} -pn $(Build.DefinitionName) -pv 1.0.0 -ps sbom -nsb https://sbom.com -li true -V Verbose -D true -lto 90

The following result is returned

```
Sbom-tool is already installed: 3.1.0

##[debug]Retrieving license information for 408 components took 85,7697623 seconds
##[warning]Encountered error while attempting to parse response. License information may not be fully recorded.
##[information]Found license information for 0 out of 408 unique components.

```

If I leave out the -lto parameter I get the timeout error

sbom-tool generate -b $(Build.ArtifactStagingDirectory) -bc ${{ parameters.workingDirectory }} -pn $(Build.DefinitionName) -pv 1.0.0 -ps sbom -nsb https://sbom.com -li true -V Verbose -D true

```
##[debug]Retrieving license information for 408 components...
##[warning]Error encountered while fetching license information from API, resulting SBOM may have incomplete license information: The request was canceled due to the configured HttpClient.Timeout of 30 seconds elapsing.
##[debug]Retrieving license information for 408 components took 30,020855 seconds

```

But strangely enough, sometimes the licence generation works even if the timeout is not specified

sfoslund:
@schuhbsi I'm not sure I'm understanding the issues here- it sounds like the timeout is working as expected and terminating license discovery after the default of 30 seconds. We fetch license data from clearly defined which can sometimes be a long running operation or run into throttling, so it is expected that there will occasionally be cases in which license data gathering can take longer than the default of 30 seconds. Can you please clarify what this issue is/ what kinds of changes you would like to be made here?

schuhbsi:
@sfoslund Sorry that I did not express myself clearly. I would simply like to have the necessary licenses, i.e. license types, for my components that are integrated in the projects. But I have the following problem that I sometimes get a timeout or another error like here ( License information may not be fully recorded), although I would have increased the timeout with the parameter from the default value 30 seconds as here in the example to 90 seconds. However, I have to say that my project also includes a large number of components, sometimes over 1000, where a license should be found. Would the solution then be to set the timeout even higher or would there be other parameters besides -lto and -li that would be important for the sbom generation with licenses?

sfoslund:
[...]

Issue #733, opened 2024-10-02
Option to fail when license information cannot be fetched


rjpowers10:
I'm using a pipeline in Azure DevOps Server. I have a custom PowerShell task to install and run sbom-tool.

Occasionally I will see warnings like these:
```
 ##[warning]Error encountered while fetching license information from API, resulting SBOM may have incomplete license information. Request returned status code: TooManyRequests
```
```
##[warning]Error encountered while fetching license information from API, resulting SBOM may have incomplete license information: The request was canceled due to the configured HttpClient.Timeout of 30 seconds elapsing.
```

The underlying problem may be related to #584 but I have a more general issue. I want this task to end in failure as missing the license information is a no-go for me. In other words I'd rather this warning be elevated to an error. Maybe there is a way to do this already but I haven't found one. For example, I tried changing the PowerShell task to set `$WarningPreference = 'Stop'` but that had no effect.

https://github.com/microsoft/sbom-tool/blob/59a68d158c7d3e9860fa8c248911ef8f038d5d9a/src/Microsoft.Sbom.Api/Executors/LicenseInformationService.cs#L61

rjpowers10:
As a workaround I'll probably use PowerShell to inspect the resulting JSON file to see if any of the license properties are NOASSERTION and fail accordingly.

I think that's good enough for me, but I raised the issue just to see if anyone else had thoughts or encountered the same issue.

DaveTryon:
@rjpowers10, thanks for the suggestion! Our team is currently focusing on other areas and won't have the bandwidth to pick this up. That said, if you wanted to open a PR to implement it, we'd be happy to review and eventually merge your change. Some thoughts for the PR:

- There are a few places where you might want to make the change, depending on how much you want to define as "in scope":
  - [LicenseInformationService line 61](https://github.com/microsoft/sbom-tool/blob/59a68d158c7d3e9860fa8c248911ef8f038d5d9a/src/Microsoft.Sbom.Api/Executors/LicenseInformationService.cs#L61)
  - [LicenseInformationService line 78](https://github.com/microsoft/sbom-tool/blob/59a68d158c7d3e9860fa8c248911ef8f038d5d9a/src/Microsoft.Sbom.Api/Executors/LicenseInformationService.cs#L78)
[...]

PR #1532, opened 2026-08-26
Make the ClearlyDefined batch size configurable via -lbs

ChristophHornung:
# Make the ClearlyDefined batch size configurable via `-lbs`

## Problem

When `-li true` is set, `LicenseInformationService.FetchLicenseInformationFromAPI` sends components to the ClearlyDefined batch API in slices of a hardcoded `batchSize = 500`. For a project with a few hundred components that is a single POST to `https://api.clearlydefined.io/definitions?expand=-files` asking the service to resolve every coordinate at once.

That request regularly exceeds the origin timeout in front of `api.clearlydefined.io`, which responds **HTTP 524**. Because there is no retry and no per-batch fallback, one failed response means the SBOM ends up with no license information at all:

```
##[debug]Retrieving license information for 292 components...
##[debug]Retrieving license information for 292 components took 125.1564491 seconds
##[warning]Error encountered while fetching license information from API, resulting SBOM may have
          incomplete license information. Request returned status code: 524
##[information]Found license information for 0 out of 296 unique components.
```

Raising `-lto` does not help — the client timeout was already honored here (the request ran a full 125 s); the failure is on the server side, so a longer client timeout just waits longer for a response that never arrives. Splitting the work into smaller requests does help, but the batch size is not reachable from the outside.

## What this changes

Per the second maintainer takeaway in #944 (expose the batch size as a CLI argument):

- New `LicenseInformationBatchSize` (`-lbs`) generation argument, plumbed through `ConfigurationMapper`, `IConfiguration` / `Configuration` / `InputConfiguration`, and sanitized in `ConfigSanitizer` exactly like `-lto`: null → default, non-positive → default with a warning, above max → truncated with a warning, and a warning when `-lbs` is supplied without `-li`.
- `LicenseInformationService` takes the batch size as a parameter instead of hardcoding it.
- `Constants.DefaultLicenseFetchBatchSize = 500`, `Constants.MaxLicenseFetchBatchSize = 1000`.

**The default of 500 is deliberately the previous hardcoded value, so behaviour is unchanged unless `-lbs` is passed.** The #944 discussion also floated lowering the default to 100; that is a one-line change on top of this if you would prefer users to get smaller batches without opting in. I kept it opt-in so this PR carries no behavioural change of its own — happy to switch if you would rather have the new default.

## Notes for reviewers

**This touches version-pinned interfaces.** `ILicenseInformationService.FetchLicenseInformationFromAPI` and `ILicenseInformationFetcher.FetchLicenseInformationAsync` each gain an `int batchSize` parameter, and `IConfiguration` gains a property; `VersionSpecificPins/Version_4_0/InterfaceConcretionTests.cs` is updated accordingly. The precedent I followed is #1110, which added a property to the pinned `IConfiguration` and updated the same pin class, shipping in v4.1.0.

If you would rather not change that surface at all, the alternative is to inject `IConfiguration` into `LicenseInformationService` and read the batch size there, leaving both executor interfaces untouched. Happy to switch — just say which you prefer.

**Not included here:** the third takeaway in #944 (a CLI argument to fail the run when license fetching fails). That is independent of this change and also covered by #733; glad to open a separate PR for it.

**Two adjacent gaps noticed while doing this** — both pre-existing and left alone to keep the diff focused:

1. `src/Microsoft.Sbom.Targets/README.md` documents `<SbomGenerationLicenseInformationTimeoutInSeconds>`, but no such property exists in `GenerateSbom.cs`, `Microsoft.Sbom.Targets.targets`, or `SbomCLIToolTask.cs` — the MSBuild task cannot actually set `-lto`. `-lbs` is therefore not added to that README either, so as not to document a second option that does nothing.
2. `ConfigFile.cs` has no `LicenseInformationTimeoutInSeconds`, so `-lto` cannot be set through a config file. `-lbs` mirrors that and is CLI-only.

Let me know if you want either folded into this PR.

## Testing

- Added `LicenseInformationBatchSize` sanitizer tests mirroring the existing `-lto` coverage: valid values pass through unchanged (1, 500, 1000), out-of-range values are corrected (negative, zero, max+1, `int.MaxValue`), and null falls back to the default with `SettingSource.Default`.
- Existing test suite unchanged apart from the interface pin updates.

Addresses #944.


ChristophHornung:

PR #1533, opened 2026-09-11
Add win-x64 executable NuGet package

JoseRenan:
## Summary

Adds a new `Microsoft.Sbom.Tool.Win-x64` NuGet package containing the self-contained Windows x64 executable.

The package is produced by the existing official release pipeline and uses the same version as the corresponding sbom-tool release.

## Changes

- Added a standalone nuspec for `Microsoft.Sbom.Tool.Win-x64`.
- Packages the executable at:
  `tools/win-x64/sbom-tool-win-x64.exe`
- Added a package README describing the payload and location.
- Updated the Windows release job to:
  - build and sign the self-contained executable;
  - derive the package version from the executable;
  - create the new NuGet package;
  - sign it through the existing ESRP package-signing step.
- Reuses the existing private-feed and nuget.org publication stages.

The existing `Microsoft.Sbom.Tool` and `Microsoft.Sbom.DotNetTool` packages are unchanged.

## Validation

- Packed the nuspec using the published `v4.1.5` Windows executable.
- Confirmed the resulting package contains:
  - `README.md`
  - `tools/win-x64/sbom-tool-win-x64.exe`
- Confirmed the packaged executable reports the expected version.
- Confirmed the pipeline YAML and nuspec XML parse successfully.


pipelines/build-test-tool-template.yaml
```diff
@@ -4,6 +4,15 @@ steps:
     inputs:
       useGlobalJson: true
 
+  - task: NuGetAuthenticate@1
+    displayName: 'Authenticate to Azure Artifacts'
+
+  - pwsh: |
+      $configPath = "$(Build.SourcesDirectory)/nuget.config"
+      dotnet nuget remove source nuget.org --configfile $configPath
+      dotnet nuget add source "https://pkgs.dev.azure.com/mseng/AzureDevOps/_packaging/ManifestTool/nuget/v3/index.json" --name ManifestTool --configfile $configPath
```
sfoslund: Why is this necessary?

sfoslund (approved):

pipelines/build-test-tool-template.yaml
JoseRenan: 1ESPT removed the network isolation bypass they had, so I just made the restore request from the internal feed instead of nuget.org, since the internal feed has nuget.org as upstream. That doesn't change the workflow outside the pipeline build, local builds and external contributors are still able to restore normally

sfoslund (approved):

